PR #3, opened 2017-09-11
Adding C to the tree traversal and verlet

Gathros:
Added C to tree traversal and verlet. I copied the c++ for the verlet and changed the cout to printf and the header.


leios:
These look great! Thanks a bunch!

PR #5, opened 2017-09-16
Adding javascript and c to euclidean algorithm

Gathros:
why did you remove euclid_rec


leios:
1. Thanks! That was quick!
2. I debated leaving the recursive method in, but I decided against it because it was just the mod-based method. I guess I should have left it in.

Gathros:
Oh ok and I made it in the live stream.

leios:
Haha, fair enough! Thanks so much for submitting / the support! You rock!

PR #6, opened 2017-09-16
Added Python Example for Euclidean

KingFredrickVI:


leios:
That was quick! Thanks for the submission!

PR #8, opened 2017-09-17
add rust, haskell euclidean

ubsan:
also remove unnecessary include of math.h for C and C++


leios:
Haha, you are totally right! We did not need those includes. Thanks for catching that!

Also, thanks for the submission! You rock!

PR #14, opened 2017-09-18
Fixing the euclid alg to allow for negative inputs

Gathros:


KingFredrickVI:
This seems like an unnecessary change. This isn't meant to be production code but example code for people to try and understand the concepts and by adding the abs() it lengthens the code / makes it slightly harder to read.

julianschacherpp:
I'am on your side there. I think we have to mark the lines of code for handling the negative values clearly or provide two versions of the functions.

leios:
Yeah, I think adding an abs() doesn't make things too confusing and serves a purpose. I do agree with KingFredrickVI in general, though. Code cleanliness is most important here.

I think I'll merge these two in this case.

Gustorn:
@KingFredrickVI Just to clarify something since we talked about this in length over on Discord: because of some C/C++ idiosyncrasies the behavior of the one without `abs` was actually implementation-defined:

```cpp
int main() {
  return -10 % -7;
}
```

The compiler is free to translate this to the following:

```cpp
int main() {
  return 0;
}
```

I think avoiding hard to debug and non-obvious mistakes is a nice goal in a public algorithm collection.

HappyLittleGit:
I agree with unnecessary change. Example code is best kept as clean and concise as possible. abs should do the trick.

ubsan:
@Gustorn I don't believe that's true of any modern versions of C or C++.

Gustorn:
@ubsan You are right, for some reason I thought the return value would be unspecified (and yes, I know the difference between undefined/unspecified/implementation-defined). Unfortunately it still has the ability to mess with the algorithm without the `abs` call.

ubsan:
@Gustorn yeah, but in a completely defined way 😄 

I'm assuming C99/C++11

PR #15, opened 2017-09-18
Negative number support for the Java and Python examples for Euclid's Algorithm

Butt4cak3:
This pull request does the same as #14 but with Java and Python. Both code examples now support negative numbers as input parameters.


KingFredrickVI:
This seems like an unnecessary change. This isn't meant to be production code but example code for people to try and understand the concepts and by adding the abs() it lengthens the code / makes it slightly harder to read.

julianschacherpp:
"I'am on your side there. I think we have to mark the lines of code for handling the negative values clearly or provide two versions of the functions." -me, PR 14

leios:
As mentioned in 14, I think this serves a purpose and is worth putting in.

Thanks for the updates!

PR #16, opened 2017-09-18
Add negative number support for EuclideanAlg C#

julianschacherpp:
If the changes are that little, I think negative number support isn't that messy as I thought.


julianschacherpp:
Should I keep the comments, which are explaining the Math.Abs()-lines?

leios:
I think it's actually better with the comments! Thanks!

PR #17, opened 2017-09-19
Rust/Haskell/OCaml/C++ negative nums

ubsan:
also standardize C++ style


julianschacherpp:
I'am not quite sure, if it's better to use a -= b instead of a = a - b

leios:
I think we are overthinking this. The `-=` is standard / I am not worried about that one. I just don't like the `auto` and `uint64_t`

All other changes are great and I really appreciate it! =)

ubsan:
@leios I'm not sure what issue you have with `auto` - that's... how C++ is written. You can write C++ like C with classes, but then it's not really C++.

leios:
Wow you are quick. Thanks! =)

PR #19, opened 2017-09-20
update C++ to newstyle

ubsan:
I didn't update the C++ in the body, because I think it should be changed to pseudocode


leios:
I agree. The next question is figuring out clean psuedocode. I'll write a section on that soon and update the in-text code ASAP (hopefully this week). There are a bunch of things that need to be done, so sorry for moving slowly on this. Again, we are just starting the project, so it has to move slow for now. 

Thanks for the update and the help! I think it's clean enough.

PR #26, opened 2017-10-17
Fixing an error

Gathros:
What you did James, didn't make sense or would lead people astray.


Gathros:
BTW, why are you using ≃?

leios:
yup, you are right

leios:
I am using ≃ to show it's approximately equal. Maybe it's the wrong notation, but it's what I've always used.

PR #28, opened 2017-12-20
Fix typo

julianschacherpp:
tiemstep -> timestep


julianschacherpp:
Shouldn't the sentence be "Now, let's say we actually need the velocity to calculate out **the** next timestep." instead of "Now, let's say we actually need the velocity to calculate out next timestep."?

leios:
Haha, thanks a lot!

PR #29, opened 2017-12-24
C++ FFT

Gathros:
I fixed it


Gathros:
just added it for python

chapters/computational_mathematics/FFT/code/fft.cpp
```diff
@@ -0,0 +1,84 @@
...
+	std::complex<double> w;
+        for(int k = 0; k < N/2; k++){
+            w = exp(std::complex<double>(0, -2.0*M_PI*k/N));
+            X[k + N/2] = X[k] - w*X[k + N/2];
+            X[k] -= (X[k + N/2]-X[k]);
+        }
+    }
+}
+
+void bitReverse(std::complex<double> *X, int N){
```
ubsan: you should name this `bit_reverse` in order to keep the style.

chapters/computational_mathematics/FFT/code/fft.cpp
```diff
@@ -0,0 +1,84 @@
+#include <complex>
+#include <cmath>
+#include <ctime>
+#include <cstdlib>
+#include <iostream>
+
+void cooley_tukey(std::complex<double> *X, int N){
+    if(N >= 2){
+        std::complex<double> temp[N/2];
```
ubsan: this is invalid C++.

chapters/computational_mathematics/FFT/code/fft.cpp
```diff
@@ -0,0 +1,84 @@
...
+				X[k + j + stride/2] = X[k + j] - v*X[k + j + stride/2];
+            			X[k + j] -= (X[k + j + stride/2]-X[k + j]);
+				v *= w;
+			}
+		}
+	}
+}
+
+int main(){
+	srand(time(NULL));
```
ubsan: no

ubsan:
You should let me rewrite this.

Gathros:
no problem no one else was writing it so i tried to take a swing at it.

ubsan:
I'll upload my version as soon as I'm done with it - it's basically a straight translation of yours.

ubsan:
Christmas - not a great time to write code :P

Gathros:
Take your time. BTW why is the second one not valid C++?

ubsan:
Arrays with runtime bound aren't valid (and would be a source of many stack overflow bugs if they were). One should use `std::vector`.

leios:
Man, I sleep in on Christmas morning, and miss a cool conversation! Thanks for being awesome folks!

Gathros:
Thanks ubsan. You didn't miss much James.

ubsan:
```
#include <iostream>
#include <iomanip>
#include <complex>
#include <vector>
#include <random>
#include <array>
#include <cstdint>

/* TODO: add some comments here before actually merging it */
/* TODO: in general, more commenting is good */
struct range {
  std::size_t begin;
  std::size_t end;
};

template <typename T>
class span {
  T* data_;
  std::size_t size_;
public:
  template <std::size_t N>
  span(std::array<T, N>& arr) : data_(arr.data()), size_(N) {}

  explicit span(T* ptr, std::size_t len) : data_(ptr), size_(len) {}

  T* data() { return data_; }

  std::size_t size() const { return size_; }

  T& operator[](std::size_t i) { return data_[i]; }

  span<T> operator[](range r) {
    return span<T>(data() + r.begin, r.end - r.begin);
  }
};

using c64 = std::complex<double>;
template <typename T>
constexpr T pi = 3.14159265358979323846264338327950288419716;

void cooley_tukey(span<c64> original, std::vector<c64>&& temp = {}) {
  auto size = original.size();
  if (size >= 2) {
    temp.reserve(size / 2);
    for (std::size_t i = 0; i < size / 2; ++i) {
      temp[i] = original[i * 2 + 1];
      original[i] = original[i * 2];
    }
    for (std::size_t i = 0; i < size / 2; ++i) {
      original[i + size / 2] = temp[i];
    }

    cooley_tukey(original[range{0, size / 2}], std::move(temp));
    cooley_tukey(original[range{size / 2, size}], std::move(temp));

    for (std::size_t k = 0; k < size / 2; ++k) {
      auto w = std::exp(c64(0, -2.0 * pi<double> * k / size));
      original[k + size / 2] = original[k] - w * original[k + size / 2];
      original[k] -= (original[k + size / 2] - original[k]);
    }
  }
}

void bit_reverse(span<c64> original) {
  auto size = original.size();

  for (std::uint32_t i = 0; i < size; ++i) {
    auto b = i;
    // Reverse bits
    b = (((b & 0xaaaaaaaa) >> 1) | ((b & 0x55555555) << 1));
    b = (((b & 0xcccccccc) >> 2) | ((b & 0x33333333) << 2));
    b = (((b & 0xf0f0f0f0) >> 4) | ((b & 0x0f0f0f0f) << 4));
    b = (((b & 0xff00ff00) >> 8) | ((b & 0x00ff00ff) << 8));
    b = ((b >> 16) | (b << 16)) >> (32 - std::uint32_t(log2(size)));
    if (b > i) {
      std::swap(original[b], original[i]);
    }
  }
}

void iterative_cooley_tukey(span<c64> original) {
  bit_reverse(original);

  auto size = original.size();
  for (std::size_t stride = 2; stride <= size; stride *= 2) {
    auto w = exp(c64(0, -2.0 * pi<double> / stride));
    for (std::size_t j = 0; j < size; j += stride) {
      auto v = c64(1.0);
      for (std::size_t k = 0; k < stride / 2; k++) {
        original[k + j + stride / 2] =
          original[k + j] - v * original[k + j + stride / 2];
        original[k + j] -= (original[k + j + stride / 2] - original[k + j]);
        v *= w;
      }
    }
  }
}

int main() {
  auto random_number_generator = std::mt19937_64();
  auto generate_random_double = [&]() {
    auto rn = random_number_generator();
    return double(rn) / double(UINT64_MAX);
  };

  std::array<c64, 64> initial;

  for (auto& el : initial) {
    el = generate_random_double();
  }

  auto recursive = initial;
  auto iterative = initial;

  cooley_tukey(recursive);
  iterative_cooley_tukey(iterative);

  std::cout
    << std::right
    << std::setw(16) << "idx"
    << std::setw(16) << "rec"
    << std::setw(16) << "it"
    << std::setw(16) << "subtracted"
    << '\n';
  for (int i = 0; i < initial.size(); ++i) {
    auto rec = recursive[i];
    auto it = iterative[i];
    std::cout
      << std::setw(16) << i
      << std::setw(16) << std::abs(rec)
      << std::setw(16) << std::abs(it)
      << std::setw(16) << (std::abs(rec) - std::abs(it))
      << '\n';
  }

  auto i = 0;
  std::cin >> i;
  return 0;
}
```

ubsan:
This is basically meant to be a straight-line change of the code to modern C++ style, with the exact same logic.

Gathros:
Thanks, I'll update it later.

ubsan:
that's much better for C code :) Although I think there might be something better to get randomness in C?

leios:
Sorry, I've been busy recently. I'll give a full update on these changes later today (I promise). The `rand()` function in this case is fine, I think. It's really just there to make sure that the fft works in both cases with arbitrary inputs.

julianschacherpp:
Didn’t looked at the code, but a comment, that this function has it flaws would be a solution.

Gathros:
https://stackoverflow.com/questions/822323/how-to-generate-a-random-number-in-c#822368
Since it's not for security, then rand is fine.

ubsan:
yeah, I'd add a comment about getting random numbers in C, and how rand is not secure or good for getting randomness.

Gathros:
I just need values, thats all.

Gathros:
I imagine anyone wanting to use the code knows not to use it for random numbers.

leios:
Yeah, I agree with @Gathros here. It's basically just a way to fill the array with values. No need to fuss about it.

julianschacherpp:
If it’s just for filling the array, it’s fine without a comment I guess (didn’t looked into code, sry). Randomness and what to use could be a topic for a chapter though (or even an algorithm?).

Gathros:
@julianshacherpp there is no algorithm for secure random number generator since for truly random numbers come from real world data

julianschacherpp:
hmm, not really into that, but I thought about algorithms for handling the data

leios:
So, the C code seems easy enough to follow (it uses tabs, though). I am not really that picky about that kind of stuff right now. If you could add a comment at the start with "submitted by Gathros / ubsan" that would be nice. I like the idea of highlighting who wrote what so people know it's a community effort.

We might need a few more comments in the C++ code (as indicated at the start of the code snippet provided). I can merge it now so @ubsan can play with it on her fork, or I can wait just a bit longer here. Let me know what you guys think.

Ultimately, I really just have 3 comments regarding the C++ code:

1. I'm having trouble understanding the point behind the `span` class. Though it is syntactically impressive, it is really hard to understand what all the contents are supposed to do. It seems to me that we could basically replace it with an `std::vector<std::complex<double>>`, which is easier to understand intuitively.
2. Is `std::vector<c64>&& temp = {}` common syntax? If so, keep it. I am not really sure how it works, if I'm being honest. When it is being called with `std::move(temp)`, it makes everything a little more challenging.
3. The `range` struct was really useful here, I think. I'm kinda a fan of it for this case.

Thanks again, you folks are great!

ubsan:
@leios Yeah, it wasn't intended to be final code.

1) `span` is a class from later C++ standards (likely to be in C++20). The reason you don't use `std::vector` is because sub-vectors are `O(N)`, while `span` is `O(1)` subslicing.

2) it's not - I need to think of a good way to do this. Basically, I'm doing this so that you don't have to reallocate for every single subcall, you only allocate once.

3) the `range` class is the thing I'm least keen on. It's quite Rust-y, and is unlikely to come into C++.

leios:
Thanks for the quick response @ubsan. I see that you are working on it. Would you prefer for me to merge the changes now or later (after you have reworked it a little)? Let me know what's easiest for you here.

Regarding the `span`: the way I see it, we should stick to what people know and can easily understand without too much thought. If they know the new features, they'll know immediately how to use them; however, I also see the other point that it's best to show the most efficient code because they can piece together simple code with the psuedocode and other code snippets provided. 

I should add this to the discussions list in the TODO section and bring it up to the community soon.

ubsan:
@leios oh, I have an idea.

julianschacherpp:
I think applying the best and most logical should be the way to go. In the end pseudo-code helps that everyone understands it and the implementations should provide the best possible and most reasonable implementation in the targeted language. Introducing one or two not well known features there shouldn't be a problem since everyone should be able to use [insert search engine of your choice here].
That way, people learn the new features as well, which is never a bad thing (as @leios kind of mentioned before).

In the future I can also imagine explanations of especially interesting/complex code snippets/examples...

Well, I guess it's really a topic for a discussion :D

ubsan:
https://gist.github.com/ubsan/1ef6659bac737c2720256ab1b231d108

@Gathros please comment, I don't understand the code :P

leios:
Ah, I see what you did. It seems much cleaner to me.

Gathros:
@ubsan Should I replace the c++ file with that one?

ubsan:
@Gathros yeah, although it'd be nice if you commented it, because I don't understand how this actually works :D it works tho :3

leios:
We might need a comment explaining the bit reversal in both the C and C++ versions.

Gathros:
OK I will comment on both.

leios:
Cool! You rock!

Gathros:
I commented a lot. @ubsan can you check the C++ header comments to see if it correct.

leios:
I think these comments are good. If there is not more discussion, I'll merge it in.

Thanks again!

ubsan:
@Gathros could you add the new version of the code, without the `range` and `span` types, linked here? https://gist.github.com/ubsan/1ef6659bac737c2720256ab1b231d108

This is more similar to what "good standard C++" should look like.

leios:
Oh, shoot. I just checked the comments, I kinda just browsed the code.

Gathros:
Done, I have updated it. There were two issues with the code; you (@ubsan) used the wrong array for the iterative method on line 98, and line 23 you also need to include arrays with two elements also.

ubsan:
Thanks @Gathros :)

Gathros:
No problem. We all make mistakes, some (me) more then other.

leios:
Alright, looks good! Thanks again!

PR #30, opened 2018-01-01
Fix and improve little things.

julianschacherpp:


leios:
Cool! Thanks for these updates!

PR #36, opened 2018-01-30
Enhance mult-lang.

julianschacherpp:
Update book.json and mult-lang implementation where needed.

### DIFF
--- a/book.json
+++ b/book.json
@@ -29,11 +29,11 @@
 				},
 				{
 					"lang": "py2",
-					"name": "python 2"
+					"name": "Python 2"
 				},
 				{
 					"lang": "py3",
-					"name": "python 3"
+					"name": "Python 3"
 				},
 				{
 					"lang": "js",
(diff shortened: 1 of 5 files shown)


leios:
Looks good, thanks!

julianschacherpp:
A quick note on how to implement mult-lang in the "new" way:

From now on you need to use the abbreviation of the language you want to implement (look at the first image for reference).
An example of what the new implementation looks like can be found in the second image.
If you need to implement a new language just follow the scheme showed in the first image.

Image 1:
screen shot 2018-01-30 at 02 54 03_book json_edit
Image 2:
screen shot 2018-01-30 at 03 03 10_enhance_mult-lang_old_new_edit

leios:
Yeah, I noticed that you changed the names. I think it's better with the abbreviations. It makes it a little easier for us to remember what's what.

julianschacherpp:
It adds to a uniform look of the AAA, since always the same title gets displayed with the same abbreviation. So typos are caught more easily.

PR #37, opened 2018-02-05
add clojure to euclidean algorithm in fundamental algorithms

earthfail:
not much but it's a start :)
sorry I didn't know how to preview the the book so expect things to break


leios:
Looks great! Thanks for the submission!

PR #39, opened 2018-02-08
Added Haskell code for convolutions 

jiegillet:
I did not add FFT code as it is messy in Haskell AFAIK.
Added my name to contributors :p

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -2,3 +2,4 @@ James Schloss
 Nicole Mazzuca
 Marius Becker
 Gathros
+Jeremie Gillet (- Jie -)
--- a/chapters/mathematical_background/convolutions/code/haskell/convolution.hs
+++ b/chapters/mathematical_background/convolutions/code/haskell/convolution.hs
@@ -1,0 +1,5 @@
+import Data.List (tails)
+
+convolution :: (Num a) => [a] -> [a] -> [a]
+convolution x = map (sum . zipWith (*) (reverse x)) . spread
+  where spread = init . tails . (replicate (length x - 1) 0 ++)
--- a/chapters/mathematical_background/convolutions/convolutions.md
+++ b/chapters/mathematical_background/convolutions/convolutions.md
@@ -1,7 +1,7 @@
 <script>
 MathJax.Hub.Queue(["Typeset",MathJax.Hub]);
 </script>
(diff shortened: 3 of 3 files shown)


julianschacherpp:
@leios you're right, it is
then @jiegillet implemented include haskell thing not quite right, going to change that then and explain what I did

leios:
Oh, @julianschacherpp I already made those changes and was going to push them in just a second. I realized the julia tags were also incorrect.

leios:
I needed to add a small text blurb around the fft-based julia implementation.

julianschacherpp:
@leios ok, perfect :)

leios:
Done. Thanks @jiegillet and @julianschacherpp !

You guys rock! =)

julianschacherpp:
@jiegillet you can take a look at my post at the bottom of that PR https://github.com/leios/algorithm-archive/pull/38 there I explain, how I implemented clojure in the book.json and in the book in general.

Following that logic, we need to use 
```
{% sample lang="hs" %}
[import:1-5, unindent:"true", lang:"hs"](code/haskell/convolution.hs) // hs instead of haskell
```

jiegillet:
Thanks guys. I tried to mimic what I saw on another page (FFT I think), maybe I got it wrong ^^

julianschacherpp:
@jiegillet the multiple languages with that book.json is quite new and we're missing documentation on that... so it's basically our fault :)

leios:
@jiegillet To be fair, the name change from `haskell` to `hs` doesn't change too much, it just standardizes everything and ensures that we have syntax highlighting enabled for the language. I had to change my `julia` tags to `jl` because I forgot about it too.

The only other change was just to make sure there was *some* code available in the fft box when people switch to haskell.

julianschacherpp:
ohh, I was a wrong... that you didn't wrote hs instead of haskell is regarding the include-codeblock plug-in and doesn't affect theme-api at all. So you did everything about the theme-api right. That you used haskell instead of hs isn't really important, because this include-codeblock uses this table: http://highlightjs.readthedocs.io/en/latest/css-classes-reference.html#language-names-and-aliases
So it's basically just standardization like @leios said.

I were wrong, sorry.

julianschacherpp:
@jiegillet me confused, see above, in case you don't saw my comment; sry

PR #53, opened 2018-02-16
Removing some printfs that aren't needed

Gathros:


leios:
You are right, those `printf`'s are not needed. Thanks!

Gathros:
I was using them for bug testing. I forgot to remove them.

leios:
Yeah, I figured. not a problem! =)

PR #55, opened 2018-02-17
Add Gale-Shapley algorithm in C#.

julianschacherpp:
@leios some things before it can get merged:
- [x] syntax highlighting doesn't work for some reason...
- [x] should I display the output of the program and should I include Program.cs?
- [x] are @Gustorn and @Butt4cak3 happy with the mention?
- [x] also give some time, if people want to look at it :)
- [x] Fix naming to follow naming conventions.
- [x] Change SortedList of choices to IList of choices.
- [x] Make algorithm static.
- [x] remove redundant `if` statement
- [x] simplify Gale-Shapley implementation

### DIFF
--- a/chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
+++ b/chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
@@ -1,0 +1,102 @@
+﻿// Submitted by Julian Schacher (jspp) with help and enhancements by Marius Becker and gustorn.
+using System.Collections.Generic;
+
+namespace ArcaneAlgorithmArchive.ComputationalMathematics.DecisionProblems.GaleShapley
+{
+    public class GaleShapleyAlgorithm<follow, lead>
+        where follow : Person<follow, lead>
+        where lead : Person<lead, follow>
+    {
+        public List<follow> Follows { get; set; }
+        public List<lead> Leads { get; set; }
+
+        private List<follow> _lonelyFollows;
+
+        public void RunGaleShapleyAlgorithm()
+        {
+            CheckRequirements();
+            
+            // All follows are lonely.
+            _lonelyFollows = new List<follow>(Follows);
+            // Carry on until there are no lonely follows anymore.
+            while (_lonelyFollows.Count > 0)
+            {
+                // Let the lead choose again and again.
+                foreach (var lead in Leads)
+                {
+                    ChoosePartner(lead);
+                }
+            }
+        }
+
+        private void ChoosePartner(lead lead)
+        {
+            // Get the follow who want the lead (bachelors).
+            var bachelors = GetBachelors(lead);
+
+            if (bachelors.Count > 0)
+            {
+                // Give the existing partner a chance to prove theirself.
+                if (lead.Partner != null)
+                {
+                    bachelors.Push(lead.Partner);
+                    _lonelyFollows.Add(lead.Partner);
+                    lead.Partner = null;
+                }
+                
+                // One partner for the beginning.
+                lead.Partner = bachelors.Pop();
+                // Let the lead choose.
+                foreach (var bachelor in bachelors)
+                {
+                    foreach (var choice in lead.Choices.Values)
+                    {
+                        if (choice == lead.Partner)
+                            break;
+                        if (choice == bachelor)
+                        {
+                            lead.Partner = bachelor;
+                            break;
+                        }
+                    }
+                }
+                _lonelyFollows.Remove(lead.Partner);
+            }
+        }
+        
+        private Stack<follow> GetBachelors(lead lead)
+        {
+            var bachelors = new Stack<follow>();
+
+            foreach (var lonelyFollow in _lonelyFollows)
+            {
+                // Look if the lonely follow's top choice matches.
+                if (lonelyFollow.IsCurrentTopChoice(lead))
+                {
+                    bachelors.Push(lonelyFollow);
+                    lonelyFollow.CurrentTopChoiceIndex++;
+                }
+            }
+            
+            return bachelors;
+        }
+        
+        // Check some requirements for the algorithm.
+        // Not actually a part of the algorithm.
+        public void CheckRequirements()
+        {
+            if (Follows.Count != Leads.Count)
+                throw new System.Exception("Lead and Follows are not of the same count.");
+
+            for (int i = 0; i < Follows.Count; i++)
+            {
+                if (Follows[i].Choices.Count != Leads.Count)
+                    throw new System.Exception(
+                        $"The count of choices by the follow at index {i} in Follows is not equal to the count of Leads.");
+                if (Leads[i].Choices.Count != Follows.Count)
+                    throw new System.Exception(
+                        $"The count of choices by the lead at index {i} in Leads is not equal to the count of Follows.");
+            }
+        }
+    }
+}
--- a/chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
+++ b/chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
@@ -1,38 +1,32 @@
-﻿// Submitted by Julian Schacher (jspp) with help and enhancements by Marius Becker and gustorn.
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 
 namespace ArcaneAlgorithmArchive.ComputationalMathematics.DecisionProblems.GaleShapley
 {
-    public class GaleShapleyAlgorithm<follow, lead>
-        where follow : Person<follow, lead>
-        where lead : Person<lead, follow>
+    public static class GaleShapleyAlgorithm<TFollow, TLead>
+        where TFollow : Person<TFollow, TLead>
+        where TLead : Person<TLead, TFollow>
     {
-        public List<follow> Follows { get; set; }
-        public List<lead> Leads { get; set; }
-
-        private List<follow> _lonelyFollows;
-
-        public void RunGaleShapleyAlgorithm()
+        public static void RunGaleShapleyAlgorithm(List<TFollow> follows, List<TLead> leads)
         {
-            CheckRequirements();
+            CheckRequirements(follows, leads);
             
             // All follows are lonely.
-            _lonelyFollows = new List<follow>(Follows);
+            var lonelyFollows = new List<TFollow>(follows);
             // Carry on until there are no lonely follows anymore.
-            while (_lonelyFollows.Count > 0)
+            while (lonelyFollows.Count > 0)
             {
                 // Let the lead choose again and again.
-                foreach (var lead in Leads)
+                foreach (var lead in leads)
                 {
-                    ChoosePartner(lead);
+                    ChoosePartner(lead, lonelyFollows);
                 }
             }
         }
 
-        private void ChoosePartner(lead lead)
+        private static void ChoosePartner(TLead lead, List<TFollow> lonelyFollows)
         {
-            // Get the follow who want the lead (bachelors).
-            var bachelors = GetBachelors(lead);
+            // Get the follows who want the lead (bachelors).
+            var bachelors = GetBachelors(lead, lonelyFollows);
 
             if (bachelors.Count > 0)
             {
@@ -40,7 +34,7 @@ namespace ArcaneAlgorithmArchive.ComputationalMathematics.DecisionProblems.GaleS
                 if (lead.Partner != null)
                 {
                     bachelors.Push(lead.Partner);
-                    _lonelyFollows.Add(lead.Partner);
+                    lonelyFollows.Add(lead.Partner);
                     lead.Partner = null;
                 }
                 
@@ -49,7 +43,7 @@ namespace ArcaneAlgorithmArchive.ComputationalMathematics.DecisionProblems.GaleS
                 // Let the lead choose.
                 foreach (var bachelor in bachelors)
                 {
-                    foreach (var choice in lead.Choices.Values)
+                    foreach (var choice in lead.Choices)
                     {
                         if (choice == lead.Partner)
                             break;
@@ -60,15 +54,15 @@ namespace ArcaneAlgorithmArchive.ComputationalMathematics.DecisionProblems.GaleS
                         }
                     }
                 }
-                _lonelyFollows.Remove(lead.Partner);
+                lonelyFollows.Remove(lead.Partner);
             }
         }
         
-        private Stack<follow> GetBachelors(lead lead)
+        private static Stack<TFollow> GetBachelors(TLead lead, List<TFollow> lonelyFollows)
         {
-            var bachelors = new Stack<follow>();
+            var bachelors = new Stack<TFollow>();
 
-            foreach (var lonelyFollow in _lonelyFollows)
+            foreach (var lonelyFollow in lonelyFollows)
             {
                 // Look if the lonely follow's top choice matches.
                 if (lonelyFollow.IsCurrentTopChoice(lead))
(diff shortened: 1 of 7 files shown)


leios:
I'll leave some time for discussion on this one (and future pull requests). I have the bad habit of merging before really reviewing the code. That said, this one was thoroughly discussed on the discord channel... So it's probably alright.

julianschacherpp:
Yeah, the last point in the list doesn't really apply here. Wanted to put it in anyway because we talked about it lately :)

ubsan:
I do prefer not gendering the algorithm - the C++ code's style of having "lead" and "follow" is something I much prefer. This code is oddly semi-gendered, and it seems reasonable to remove that gendering, imo.

julianschacherpp:
@ubsan The code of the actual algorithm (in GaleShapleyAlgorithm.cs and Person.cs) is gender neutral, since genders make no sense there. There "lead" and "follow" are used.
In Program.cs I implemented a group of man and woman however. This is because it maybe provides more vividness for someone, who's new to the algorithm. Also @leios used "men" and "women" in the chapter and explanation of the algorithm.
If we decide on using a gender neutral way of describing the algorithm, I have no problem changing the Program.cs as well. So I guess that needs to be discussed (@leios ).

leios:
As far as the gendering is concerned... I don't care which one we go with, so long as we are consistent. I realize that the notation for this problem is a bit sexist; however, it's a story that helps people understand why the algorithm works the way it does.

The only implementation that does not use "men" and "women" is the c++ implementation. I suppose this conversation should have been brought up then, but I failed to do so.

One note to @julianschacherpp , I cannot seem to get the c++ syntax highlighting working either. If the c# is also amiss, then there's something we are both doing that is stupid.

julianschacherpp:
Well, we in my implementation it's also gender neutral for the actual algorithm. Also you can change one thing and switch the roles :)
```
var galeShapley = new GaleShapleyAlgorithm<Man, Woman>
{
    Follows = men,
    Leads = women
};
```
to 
```
var galeShapley = new GaleShapleyAlgorithm<Man, Woman>
{
    Follows = women,
    Leads = men
};
```

@leios I used the same syntax to import C# as in other chapters, but it just doesn't work, which is quite weird...

leios:
@julianschacherpp Yeah, the same is true for c++. It seems like the syntax highlighting is working just fine for other chapters, but not for this one.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
```diff
@@ -0,0 +1,102 @@
+﻿// Submitted by Julian Schacher (jspp) with help and enhancements by Marius Becker and gustorn.
+using System.Collections.Generic;
+
+namespace ArcaneAlgorithmArchive.ComputationalMathematics.DecisionProblems.GaleShapley
+{
+    public class GaleShapleyAlgorithm<follow, lead>
```
Gustorn: To follow C# naming conventions these should be `TFollow` and `TLead`

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Output.txt
```diff
@@ -0,0 +1,6 @@
+GaleShapley
```
Gustorn: I'm not sure we really need the output but I don't really have strong feelings either way.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System.Collections.Generic;
+
+namespace ArcaneAlgorithmArchive.ComputationalMathematics.DecisionProblems.GaleShapley
+{
+    public class Person<TSelf, TPref>
+        where TSelf : Person<TSelf, TPref>
+        where TPref : Person<TPref, TSelf>
+    {
+        private TPref _partner;
+        public TPref Partner
```
Gustorn: I'm not the biggest fan of this as a whole. Properties should only do simple and expected things. Setting a value inside another property is a bit of a codea smell in my opinion. It also makes the setter logic needlessly complicated. Compare it with:

```cs
public TPref Partner { get; private set; }

public void EngageTo(TPref other)
{
    other.Partner?.Partner = null;
    Partner = other;
    other.Partner = this as TSelf;
}

// I don't think this one is actually necessary for the algorithm
public void Breakup()
{
    Partner?.Partner = null;
    Partner = null;
}
```

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
```diff
@@ -0,0 +1,66 @@
...
+                        if (_partner != null)
+                            _partner.Partner = null;
+                        
+                        _partner = value;
+                        _partner.Partner = this as TSelf;
+                    }
+                }
+            }
+        }
+        public SortedList<int, TPref> Choices { get; set; }
```
Gustorn: The whole `SortedList` thing seems a bit redundant, you actually never take advantage of it. Since you're already using a separate method to set the choices you should just use a regular list instead and call `Sort` on it as necessary (it's actually faster as well).

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
```diff
@@ -0,0 +1,66 @@
...
+        public int CurrentTopChoiceIndex { get; set; } = 0;
+
+        public string Name { get; set; }
+
+        public Person(string name) => Name = name;
+
+        // Returns the CurrentTopChoice based on CurrentTopChoiceIndex.
+        public TPref GetCurrentTopChoice() => Choices[Choices.IndexOfKey(CurrentTopChoiceIndex)];
+        // Returns whether or not the given woman is the CurrentTopChoice
+        public bool IsCurrentTopChoice(TPref woman)
```
Gustorn: Redundant `if` statement:

```cs
return woman == GetCurrentTopChoice();
```

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
julianschacherpp: You might be right that it is a bit unexpected.
On the other hand: If I get a new partner in a real situation, is the partner of my old partner still me? Probably not. I'm not quite sure about that.
I still think the functionality is needed since my implementation creates a GaleShapleyAlgorithm class which can be used multiple times while adding / removing people and changing the choices.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
julianschacherpp: So you're suggesting a normal List with `List<int, TPref>` ?

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            if (woman == GetCurrentTopChoice())
+                return true;
+            else
+                return false;
+        }
+
+        // Set the sorted list by providing a normal one. The first element of
+        // the list will get the lowest index in the sorted one.
+        public void SetChoicesWithList(List<TPref> unsortedChoices)
```
Gustorn: This has multiple issues.

1. It's generally a bad idea to include type names in property/method names. You're better off with `SetChoices`
2. `List<TPref>` is way too specific of a type. Use `IEnumerable<TPref>` if you really want to initialize from a collection but
3. The ergonomics of this method aren't the greatest. Let's see an example from the usage code and a way to improve it:

```cs
men[0].SetChoicesWithList(new List<Woman>() { women[3], women[2], women[4], women[1], women[0]});

// If you change the signature to
public void SetChoicesWithList(params TPref[] unsortedChoices)
{
    // ...
}

// Then you can do
men[0].SetChoicesWithList(women[3], women[2], women[4], women[1], women[0]);

File: chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs [hunk 7]
```diff
@@ -0,0 +1,66 @@
...
+        {
+            var choices = new SortedList<int, TPref>();
```
Gustorn: If you get rid of the `SortedList` then regardless if you take the `params` or the `IEnumerable` route this simplifies to:

```cs
Choices = unsortedChoices.ToList();
```

As mentioned in another comment, all `SortedList` does as it is right now is slow your program down.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            CheckRequirements();
+            
+            // All follows are lonely.
+            _lonelyFollows = new List<follow>(Follows);
+            // Carry on until there are no lonely follows anymore.
+            while (_lonelyFollows.Count > 0)
+            {
+                // Let the lead choose again and again.
+                foreach (var lead in Leads)
```
Gustorn: The algorithm implementation seems *really* convoluted. I also don't think it's a great idea to pass data in class variables, it just makes things harder to follow. I know our implementations aren't the same but here's the whole method for my C# version:

```cs
var bachelors = new Queue<Person>(guys);
while (bachelors.Any())
{
    var guy = bachelors.Dequeue();
    foreach (var girl in guy.Preferences)
    {
        if (!girl.HasFiance)
        {
            guy.EngageTo(girl);
            break;
        }
        else if (girl.Prefers(guy))
        {
            bachelors.Enqueue(girl.Fiance);
            guy.EngageTo(girl);
            break;
        }
    }
}
```

I don't think there's a reason for this algorithm to span 50 lines.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Output.txt
Gathros: We should get rid of the output, no ones done it before, since it doesn't really add anything.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
Gustorn: Even if it *is* needed, the `EngageTo/Breakup` thing actually paints a more clear picture. With the private setter noone's going to be able to set the partner outside of the class anyway

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
Gustorn: I'm suggesting `IList<TPref>`, already pre-sorted

julianschacherpp: ok

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Output.txt
julianschacherpp: @Gathros I shows what comes out of the program. Having that is interesting and can help people, if they write their own implementation.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
julianschacherpp: If I go for the `IList<TPref>` I don't really need that method at all. I can just do in the Program.cs: 
```cs
men[0].Choices = new List<Woman>() { women[3], women[2], women[4], women[1], women[0]};
```

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Output.txt
Gathros: No it donesn't, you just need to compile the program and run it, thats it. `Output.txt` helps no one.

julianschacherpp: Fair enough. It comes down to personal preference I guess. 
On the one hand you can argue that not everyone can run the code quickly, because he had to install .NET stuff or something first, and get the output.
On the other hand people may just run it and we keep it cleaner. Then I could also go for randomly generated input.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
julianschacherpp: I wasn't quite sure about using a class for that as well. It was kind of 50/50 for me, when I did. I'll look at the algorithm in a static one.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Output.txt
Gathros: For the first point, if someone is interested in C# then they definitely have .NET installed already. So there is no real reason to add a `Output.txt` file.

Butt4cak3:
> are @Gustorn and @Butt4cak3 happy with the mention?

I don't feel like I did enough to deserve a mention in this one, but go ahead if you want to. I'm fine with it.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
```diff
@@ -0,0 +1,56 @@
...
+        public string Name { get; set; }
+
+        public Person(string name) => Name = name;
+
+        // Returns the CurrentTopChoice based on CurrentTopChoiceIndex.
+        public TPref GetCurrentTopChoice() => Choices[CurrentTopChoiceIndex];
+        // Returns whether or not the given woman is the CurrentTopChoice
+        public bool IsCurrentTopChoice(TPref woman)
+        {
+            if (woman == GetCurrentTopChoice())
```
Gustorn: This `if` statement is still redundant

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/Person.cs
Gustorn: And I still think this is an abuse/misuse of the property syntax

julianschacherpp:
Ok, since @Butt4cak3 showed me this awesome To-Do-List feature, here's what I plan to do regarding the improvements by @Gustorn 
- [ ] remove redundant `if` statement
- [ ] simplify Gale-Shapley implementation

julianschacherpp:
A poll regarding the output-file thing: https://strawpoll.com/r7dc379k

Gustorn:
I'm fine with the mention although it's not really necessary :)

julianschacherpp:
I simplified and de-convoluted my implementation now. I still want to look into some things however:
- Whether or not I place the ProcessProposal method in the Person class.
- If I change the Partner logic like @Gustorn suggested or not.

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/GaleShapleyAlgorithm.cs
```diff
@@ -0,0 +1,46 @@
...
+            // All follows are lonely.
+            var lonelyFollows = new List<TFollow>(follows);
+            
+            // Carry on until there are no lonely follows anymore.
+            while (lonelyFollows.Count > 0)
+            {
+                // Let every lonely follow propose to their current top choice.
+                foreach (var lonelyFollow in lonelyFollows)
+                {
+                    ProcessProposal(lonelyFollow, lonelyFollow.GetCurrentTopChoice());
```
Gustorn: I would probably write these two lines as
```cs
ProcessProposal(lonelyFollow, lonelyFollow.GetNextChoice());

// And change the implementation of Person in the following manner:
public class Person
{
    private currentTopChoiceIndex = 0;
    public TPref GetNextChoice() => Choices[currentTopChoiceIndex++];
}
```

This way the API ensures that you can only propose to your next choice (it doesn't require you to increment the index manually) so it's more bulletproof.

julianschacherpp: Good idea. I thought about it as well, but I think I can implement it like that. My concerns were, that you still need to be able to set the Index to 0 since you may want to rerun the algorithm. So I would go for these methods while both stay public.

julianschacherpp:
I implemented random input with the latest change thanks to gustorn's shuffle method. The output file has been removed as well. These changes are due to the outcome of the straw poll.
screen shot 2018-02-22 at 03 09 54_random_input_output_file_straw_poll_result

chapters/computational_mathematics/decision_problems/stable_marriage/code/cs/IListExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace ArcaneAlgorithmArchive.Extensions
+{
+    public static class IListExtensions
```
Gustorn: This is not an interface, it shouldn't start with an 'I'.

julianschacherpp: Thanks for pointing it out, I changed it as we discussed on Discord.

julianschacherpp:
@Butt4cak3 discovered that `highlight.js`, which is used by the `import-codeblock` plugin, can't handle generics in C# code. So he is currently forking the repos and the problem may get resolved in the future. For now the PR is ready to get merged.

leios:
Alright, merging this one!

PR #56, opened 2018-02-18
Adding C code for forward euler method

Gathros:


chapters/computational_physics/euler/code/c/euler.c
```diff
@@ -0,0 +1,32 @@
...
+        for(size_t i = 2; i < n; ++i){
+            result[i] = result[i-1] - 3.0*result[i-1]*timestep;
+        }
+    }
+}
+
+int check_result(double *result, size_t n, double threshold, double timestep){
+    int is_approx = 1;
+    for(size_t i = 0; i < n; ++i){
+        double answr = exp(-3.0*(i - 1)* timestep);
```
Gustorn: I don't think saving a single character here is worth using a pretty unusual abbreviation

chapters/computational_physics/euler/code/c/euler.c
```diff
@@ -0,0 +1,32 @@
...
+            is_approx = 0;
+        }
+    }
+    return is_approx;
+}
+
+int main(){
+    double result[100];
+
+    solve_euler(0.01, result, 100);
```
Gustorn: Maybe having a `double timestamp = 0.01` variable would be a good idea since you should pass the same value to both `solve_euler` and `check_result`

chapters/computational_physics/euler/code/c/euler.c
```diff
@@ -0,0 +1,32 @@
+#include <stdio.h>
+#include <math.h>
+
+void solve_euler(double timestep, double *result, size_t n){
+    if(n != 0){
+        result[1] = 1;
```
leios: I think this should be `result[0] = 1`?

leios:
I merged because it looked like this was updated properly. Thanks! =)

Issue #60, opened 2018-02-20
Personal Issue List


ubsan:
- [x] tree_traversal: c++ - renamed to tree.cpp, fixed identifier style
- [x] tree_traveral: ocaml - added back
- [x] euclidean_algorithms: c++ - make sure it's good
- [x] euclidean_algorithms: all - rename the "euclidean_example" files to "euclidean"
- [x] convolutions: c++ - copy the fft from fft
- [x] convolutions: c++ - rewrite to make it compile
- [x] verlet: c++ - make sure it's good

- [x] sorting_searching: add C++

(gonna start fixing these once something similar to #61 is merged)

Gathros:
#86 are you done, now @ubsan?

julianschacherpp:
Maybe a better place for this To do list is a project?
Since we have the new organization now, people have the possibility to create projects. I did a project for C#, which isn't that loose however and targets a more strict tracking process.
But if you plan (@ubsan) this issue more as an ongoing issue list, a project may be a better place for that.

ubsan:
ah, yeah, I've finished now.

PR #61, opened 2018-02-20
switch over to clang-format style

ubsan:
I can add a .clang-format as well;

```
BasedOnStyle: LLVM

AlignAfterOpenBracket: AlwaysBreak
AlwaysBreakTemplateDeclarations: true
BinPackArguments: false
BinPackParameters: false
PointerAlignment: Left
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,7 @@
+BasedOnStyle: LLVM
+
+AlignAfterOpenBracket: AlwaysBreak
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: false
+BinPackParameters: false
+PointerAlignment: Left
--- a/chapters/computational_mathematics/FFT/code/c++/fft.cpp
+++ b/chapters/computational_mathematics/FFT/code/c++/fft.cpp
@@ -1,15 +1,15 @@
 // written by Gathros, modernized by Nicole Mazzuca.
 
-#include <complex>
-#include <vector>
-#include <array>
-#include <cstdint>
 #include <algorithm>
+#include <array>
+#include <complex>
+#include <cstdint>
+#include <vector>
 
 // These headers are for presentation not for the algorithm.
-#include <random>
-#include <iostream>
 #include <iomanip>
+#include <iostream>
+#include <random>
 
 using std::begin;
 using std::end;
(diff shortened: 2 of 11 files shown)


leios:
Yeah, I guess it makes sense to change the style to something that is easy to maintain. I assume that this is done simply with the `.editorconfig`?

ubsan:
It looks like .editorconfig has been entirely replaced by clang-format in the C++ community; similarly with rustfmt in the Rust community. Is it possible, for languages with formatting tools, to use those instead?

leios:
That seems reasonable and should probably be mentioned in PR #59

Butt4cak3:
A .clang-format file in the repo sounds like a good idea to me.

I briefly read up on clang-format and saw that it's not only used for C++. What I did *not* find, however, is a way to place multiple .clang-format files for multiple languages in the same repository. Is there a way to do this? Would it even make sense or should we just use it for C++ and leave it at that?

ubsan:
@Butt4cak3 I have zero opinion on non-C++ languages. That'd be up to the contributes for those languages.

leios:
Hey @ubsan I assume that this is related to the Issue #60 ? Should I wait for those issues to be resolved before merging?

ubsan:
Oh, it's actually quite trivial.

```
---
Language: Cpp
BasedOnStyle: LLVM

AlignAfterOpenBracket: AlwaysBreak
AlwaysBreakTemplateDeclarations: true
BinPackArguments: false
BinPackParameters: false
PointerAlignment: Left
---
Language: C
...
---
```

so basically, if people of other languages want to use .clang-format, they can.

@leios I don't think so; it's trivial to add support for other languages once they agree how they're going to format them, and this is blocking me on some other things I'd like to do.

leios:
@ubsan so you would prefer to merge this asap? I'll merge it soon, then. I just wanted to see if anyone else (@Gathros in particular) wanted to discuss the style.

ubsan:
Yep! although I'll add a clang-format to the base directory.

ubsan:
@leios please merge

leios:
Alright. Gathros said he didn't mind the format change earlier, so it should be fine.

PR #62, opened 2018-02-22
Added Elm code for Euler, with visualization

jiegillet:
I would appreciate everyone's feedback on this one, I am not confident this is the right thing to do. Also there are a bunch of unrelated commits, please ignore those (I don't know how to get rid of them, but I know Leios can do it later), focus on the changes themselves.

In increasing order of importance:

* Fixed some typos and added comments in Euler.md (and some newline characters got changed, ignore that)
* Added Elm to the choices of languages to book.json. When it comes to formatting Elm, I use elm-format, which formats the code following the official rules.
* Added Elm code for the Euler Method algorithm
* Added a paragraph "Interactive Visualization" to Euler.md, in which I link to the html version of the Elm code I've been working on. This is the bit I need your feedback on. Is it OK to add this kind of content to the AAA? I added the full Elm code at the end of Euler.md (in a separate box). The project itself is my first time playing with web stuff so it's very crude, and I'll appreciate feedback on that too, but let's talk about the general idea first.


Butt4cak3:
Unfortunately I don't know anything about Elm, but I can help you clean up your git repository and the pull request to get rid of all the merge conflicts and commits that shouldn't be there. Is there a more direct way that we could use to talk? Some instant messenger? You can usually find me on the community Discord.

Other than that, I looked at the code and the files you changed. I noticed that the file euler-elm.html is very big. Is that the output that you get when you "compile" the .elm file? If so, then I don't know whether we should include it in the repository or not. We usually only include the source code, but this one results in HTML so it could be viewed directly in the browser.

julianschacherpp:
Discord: https://discord.gg/qjGkBb

jiegillet:
Yes, I know the .html is quite big, because it's a web app (is this the right term?) meant to be used directly by AAA readers. If we end up using the app, I though it would make sense to provide the source code too. If we don't use it, we don't need the full elm code, just the two snippets on how the algorithm is actually implemented.

leios:
Alright, so to recap the discussions had so far on this topic on discord, there are basically 3 camps:

1. The visualization is good and we can grab whatever lines we want from the elm code
2. The visualization is good, but we need a separate, cleaner file for the elm code
3. Scrap the visualization and clean up the code.

On the one hand, I think the visualization is intuitive... on the other hand, it is also way different than any other language. I wouldn't mind having the visualization in the bulk text for all languages just to show how the method changes with different timesteps, and then having a clean elm implementation underneath. 

Obviously, if we are going to print the book, this will not work... but it is a neat visualization and I don't want to see it go to waste.

Thoughts?

PR #63, opened 2018-02-22
Add new .gitignore created with gitignore.io (arguments: Linux, macOS, Windows, GitBook).

julianschacherpp:
The updated `.gitignore` catches the files created by various operating systems.

Should we include common editors like `VSCode`, `Vim` and so on in the `.gitignore` as well?


leios:
Yup, this is just better. Thanks!

PR #70, opened 2018-03-02
modernize convolutions.cpp

ubsan:
I don't know if this is correct, but it's as correct as the old one is. It should do exactly what the lua version does.


chapters/mathematical_background/convolutions/code/c++/convolutions.cpp
```diff
@@ -1,155 +1,130 @@
...
-    Iter signal2start,
-    Iter_end signal2end,
-    Iter outstart,
-    Iter_end outend) {
-
-  auto size1 = (signal1end - signal1start);
-  auto size2 = (signal2end - signal2start);
-  c64 sum = 0;
-
-  for (std::size_t i = 0; i < (size1 < size2 ? size2 : size1); i++) {
```
ubsan: Note that this was originally incorrect.

chapters/mathematical_background/convolutions/code/c++/convolutions.cpp
```diff
@@ -1,155 +1,130 @@
...
-    } else {
-      signal1[i] = 0.0;
-      signal2[i] = 0.0;
-      signal3[i] = 0.0;
-      signal4[i] = 0.0;
-      out1[i] = 0.0;
-      out2[i] = 0.0;
-    }
-  }
+  auto signal1 = std::array<c64, 64>();
```
ubsan: I use `auto` style here in order to zero-initialize `signal1` - otherwise, it would be filled with nonsense.

Gustorn:
Wouldn't using `std::valarray` clean up this code quite a bit? You could avoid the whole odd-even partitioning with `std::slice` and it has element-wise operations already defined. I'd argue it would showcase the algorithm a lot better without the need of the iterator boilerplate (which is really nice for genericity but not that great for clarity).

chapters/mathematical_background/convolutions/code/c++/convolutions.cpp
```diff
@@ -1,155 +1,130 @@
...
 
-template <typename Iter, typename Iter_end>
-void ifft(Iter start, Iter_end end) {
-  auto size = end - start;
-  for (std::size_t i = 0; i < size; i++) {
-    start[i] = std::conj(start[i]);
-  }
+template <typename Iter>
+void inverse_fft(Iter const first, Iter const last) {
+  std::for_each(first, last, [](auto& it) { it = std::conj(it); });
```
Gustorn: I think `std::transform` would express the intent a little bit better here (and in the other `std::for_each` call in this function).

ubsan: Yeah, I agree.

ubsan:
@Gustorn honestly, I've never used `valarray`, or seen it used.

Gustorn:
I haven't seen it used very frequently either but since we basically need a (mathematical) vector-like type I think it's a good idea to use the one that the standard library already provides. Having said that, I can see the argument for not including less frequently used standard library features.

Just as a proof of concept I implemented the FFT parts of the convolutions with valarray: [godbolt](https://godbolt.org/g/r17S4E). I think it's pretty readable.

leios:
Is this PR ready to merge? I realize the discussion died down, but I am not sure if there is more discussion to be had.

PR #73, opened 2018-03-04
Cleaned up "Added Elm code for Euler, with visualization" (#62)

Butt4cak3:
This pull request is a cleand up version of #62. I got the branch up-to-date with master and removed the already merged commits/merge commits. I also changed a bit that we (I think) collectively agreed on, which is that the HTML file is code output and should, for that and other reasons, not be included.

I preserved the original commit author, so that @jiegillet still shows up as the contributor of the code.

### DIFF
--- a/book.json
+++ b/book.json
@@ -67,6 +67,10 @@
 				{
 					"lang": "clj",
 					"name": "Clojure"
+				},
+				{
+					"lang": "elm",
+					"name": "Elm"
 				}
 			],
 			"split": false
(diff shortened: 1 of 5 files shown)


jiegillet:
If we are not using the html, I'd rather not include the full code at all, just the algorithm is enough.

leios:
Alright, we still need more discussion on  this PR. Does it completely replace #62 ?

Butt4cak3:
@Leios Yes. It is the same as the other PR but rebased and with one change.

leios:
Ok, so my understanding is that it's ready to merge?

Butt4cak3:
There's no open discussion about its content, so we should be good. Go ahead, Schloss Boss.

leios:
Cool. Thanks!

PR #74, opened 2018-04-07
Update C# files fit the new CSharp repository. Fix import.

julianschacherpp:
## new C# AAA repo
Updates the C# files to fit the new [C# AAA repo](https://github.com/julianschacherpp/AlgorithmArchiveCSharp). This repo is a restructured version of the old C# AAA repo.

The most visible changes in the book are shorter and simpler namespaces.
Some comments are removed due to redundancy as well.

All that comes through the new organization of the new C# AAA repo. The repo now isn't one big project for all algorithms anymore, but one project per algorithm.

## fix import
Fix import in euclidean.md.

## ToDo
- [x] Should/Can namespace and type have the same name in this case? // in this case I decided to use the same name for type and namespace
- [x] Should a Program.cs be included? // yes


julianschacherpp:
Don't merge, problem may exist.

julianschacherpp:
Checked the files by comparing the files:

Seems good, but:
Should I/we include a Program.cs? -> I need to discuss/resolve that.

-> I create ToDo for this PR, to get the tasks tracked.

PR #76, opened 2018-04-17
Huffman coding in C.

Gathros:
@Gustorn did most of the work, but didn't make a pull request, so i'm doing it now for him.


Gustorn:
You don't really need `char*, size_t` pairs except for the concatenation, otherwise C strings work just fine (I also think it's a bad idea to implement a small string library in each algorithm).  I'm for making the two `concat` functions faster by also passing a capacity and length but I wouldn't replace each instance of C strings with our home-grown one. 

I'd also support using a 3rd party library like https://github.com/antirez/sds for string handling if we can decide on one to use.

ubsan:
Not home grown, just taking two parameters, `char [const]*` and `size_t`.

Gustorn:
If we're not going to use a library then I'm against it for everything except for the two concatenation functions and duplicate because:

1. It does not provide any performance benefit in any other functions
2. It provides pretty much no safety benefits either, it just pushes the problem one level up. You'll still have to call `strlen` on the input (or count the characters by hand which is also a bad solution)
3. It makes the implementation more complicated. The focus should be the algorithm, not string manipulation. Making a best effort using C's built-in tools seems like a better option to me. That being said, I can understand safety and performance concerns which is why I would support using a library (both for this and for standard data structures like a priority queue).

leios:
Hey Gathros, thanks for the Code! Ubsan, thanks for taking a look!

I don't mind the usage of `<string.h>`, but could you guys briefly explain what makes it more or less safe than the `char [const]*` and `size_t` pairs?

Gustorn:
Sure, you can track the capacity separately but it actually pollutes the algorithm implementation with a lot of unnecessary state (my original implementations of concat had length/capacity). I just want to point out that passing `max_len` is not any safer by default if you allow reallocation. The potentially unsafe part is mostly `strlen`: if you pass it a string that isn't null terminated you'll overrun the original buffer.

leios:
The discussion on the PR seemed to die down and I think the current version is fine. Thanks again!

PR #77, opened 2018-04-17
Added Huffman with Haskell

jiegillet:
Added Haskell implementation, and fixed a few typos and trailing spaces in huffman.md.

The main issue with the code is that (to simplify implementation) the tree generated can have some branches inverted compared with the algorithm from the video. The bit length doesn't change at all, so it's not a huge deal, maybe.

Thoughts?

### DIFF
--- a/chapters/data_compression/huffman/code/haskell/huffman.hs
+++ b/chapters/data_compression/huffman/code/haskell/huffman.hs
@@ -1,0 +1,57 @@
+import qualified Data.Map as M
+import           Data.List (insert, sort)
+
+data Tree a = Nil
+            | Leaf Int a
+            | Node Int (Tree a) (Tree a)
+                   deriving (Show, Eq)
+
+freq :: Tree a -> Int
+freq Nil          = 0
+freq (Leaf i _)   = i
+freq (Node i _ _) = i
+
+instance (Eq a) => Ord (Tree a) where
+  compare t1 t2 = compare (freq t1) (freq t2)
+
+getFrequencies :: Ord a => [a] -> [(Int, a)]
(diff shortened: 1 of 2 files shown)


chapters/data_compression/huffman/code/haskell/huffman.hs
```diff
@@ -0,0 +1,57 @@
+import qualified Data.Map as M
+import           Data.List (insert, sort)
+
+data Tree a = Nil
```
ubsan: Does this have to have `Nil`? It seems better not to support implicitly nullability.

chapters/data_compression/huffman/code/haskell/huffman.hs
```diff
@@ -0,0 +1,57 @@
...
+
+instance (Eq a) => Ord (Tree a) where
+  compare t1 t2 = compare (freq t1) (freq t2)
+
+getFrequencies :: Ord a => [a] -> [(Int, a)]
+getFrequencies = toSortedList . M.fromListWith (+) . flip zip (repeat 1)
+  where toSortedList = sort . map swap . M.toList
+        swap (a, i) = (i, a)
+
+buildTree :: (Ord a) => [a] -> Tree a
```
ubsan: And then this would just return a `Maybe (Tree a)`

jiegillet: Eh, why not ^^

jiegillet:
Switched to a Maybe (Tree a) on ubsan's suggestion.

leios:
This one also looks good to merge, unless there are outstanding problems.

PR #78, opened 2018-04-17
Adding C bubble sort

Gathros:


chapters/sorting_searching/bubble/code/c/bubble_sort.c
```diff
@@ -0,0 +1,41 @@
+#include <stdio.h>
+
+void bubble_sort(int *array, size_t n) {
```
ubsan: I think you should probably include `stddef.h` here.

chapters/sorting_searching/bubble/code/c/bubble_sort.c
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        if (!swapped) {
+            break;
+        }
+    }
+}
+
+int main() {
+    int array[10] = {1, 45, 756, 4569, 56, 3, 8, 5, -10, -4};
```
ubsan: Could you use a `#define`  instead of magic constants throughout the code?

Gathros: I don't think that's needed really, it's just an example.

leios:
Same as in #80 , I think `array[10]` is fine for this case. If there are no other issues, I'll go ahead and merge it.

leios:
@ubsan I do agree that a `#define n 10` would help a bit for readability in both cases, actually.

Gathros:
@leios I get that but it's a part of the example and not the implementation of bubble sort.

leios:
yeah, I get that too. I think it's probably fine this way.

leios:
With this and #80, I think the `array[10]` is fine for now specifically because it's example code.

PR #82, opened 2018-04-19
Add Huffman in C#.

julianschacher:
### ToDo
- [x] Rename `EncodeResult` to `EncodingResult`.
- [x] Remove code left over from debugging.
- [x] Replace array of children in Node-class with left and right children.
- [x] Make Node implement `IComparable`.
- [x] Improve creation of list of nodes.
- [x] Make the `Nodes` property of `NodePriorityList` private.
- [x] Improve `NodePriorityList` to provide `Pop()`.
- [x] Improve Node creation.
- [x] Use a `string` instead of a `List<bool>`.
- [x] Rename the enumerator `boolean` to `bit`.
- [x] Don't use first, when popping a `Node` from the `NodePriorityList`.
- [x] Check if children of node are null to identify leaf and create a property providing that info.
- [x] Make dictionary creation recursive.
- [x] Change input to "bibbity bobbity".
- [x] Fix Add method.
- [x] Use new mention template.


chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
+// submitted by Julian Schacher (jspp)
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace HuffmanCoding
+{
+    public class EncodeResult 
+    {
+        public List<bool> BitString { get; set; }
```
Gustorn: This isn't much more efficient than a regular `String` and it's much more annoying to debug and print. If you seriously want to produce a packed binary result then I'd go with `System.Collections.BitArray` but I think a regular `String` works better for educational purposes.

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+            this.Tree = tree;
+        }
+    }
+
+    public static class HuffmanCoding
+    {
+        // The Node class used for the Huffman Tree.
+        public class Node
+        {
+            public Node[] Children { get; set; } = new Node[2];
```
Gustorn: Just have two children, `Left` and `Right`. This only makes it easier to misuse the class

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+            public List<bool> BitString { get; set; } = new List<bool>();
+            public int Weight { get; set; }
+            public string Key { get; set; }
```
Gustorn: `char` would work here

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+        class NodePriorityList
+        {
+            public List<Node> Nodes { get; private set; } = new List<Node>();
+
+            public NodePriorityList() { }
+            public NodePriorityList(List<Node> nodes) => Nodes = nodes.OrderByDescending(n => n.Weight).ToList();
+
+            public void AddNode(Node newNode)
+            {
+                if (Nodes.Count == 0)
```
Gustorn: Just have `Node` implement `IComparable<Node>` then use something like:

```cs
var insertAt = Math.Max(Nodes.BinarySearch(newNode), 0);
Nodes.Insert(insertAt, newNode)
```

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+        }
+
+        // Node with biggest value at the top.
+        class NodePriorityList
+        {
+            public List<Node> Nodes { get; private set; } = new List<Node>();
+
+            public NodePriorityList() { }
+            public NodePriorityList(List<Node> nodes) => Nodes = nodes.OrderByDescending(n => n.Weight).ToList();
```
Gustorn: Have `Node` implement `IComparable` and just do:

```cs
{
    Nodes = nodes.ToList();
    Nodes.Sort();
}

File: chapters/data_compression/huffman/code/cs/HuffmanCoding.cs [hunk 6]
```diff
@@ -0,0 +1,185 @@
+// submitted by Julian Schacher (jspp)
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace HuffmanCoding
+{
+    public class EncodeResult
```
Gustorn: Maybe a better name would be `EncodingResult`

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+            foreach (var boolean in result.BitString)
+            {
+                // Go down the tree.
+                if (!boolean)
+                    currentNode = currentNode.Children[0];
+                else
+                    currentNode = currentNode.Children[1];
+
+                // Check if it's a leaf node.
+                if (currentNode.Key.Count() == 1)
```
Gustorn: Just check if all of the children are null. That's how you determine the leaves in a binary tree

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+                }
+            }
+            return output;
+        }
+
+        private static Node CreateTree(string input)
+        {
+            // Create a List of all characters and their count in input by putting them into nodes.
+            var nodes = new List<Node>();
+            foreach (var character in input)
```
Gustorn: There's a nicer way to do this. Pseudo-ish code with a toy example:

```cs
            var input = "Hello, World";
            var nodes = input
                .GroupBy(c => c)
                .Select(n => new Node { Key = n.Key, Weight = n.Count() })
                .ToList();
```

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+                    result.Weight++;
+            }
+            // Convert list of nodes to a NodePriorityList.
+            var nodePriorityList = new NodePriorityList(nodes);
+            nodes = nodePriorityList.Nodes;
+
+            // Create Tree.
+            while (nodes.Count > 1)
+            {
+                var parentNode = new Node("", 0);
```
Gustorn: You should have a default constructor so you don't need to pass dummy parameters.  I would actually have something like:

```cs
public class Node
{
    public static Node CreateLeaf(char key, int weight) { return new Node(...); }
    public static Node CreateBranch(Node left, Node right) { return new Node(...); }
    private Node(...) { /* regular Node constructor with every parameter */ }
}
```

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+                for (int i = 0; i < 2; i++)
+                {
+                    parentNode.Children[i] = nodes.Last();
+                    parentNode.Key += nodes.Last().Key;
+                    parentNode.Weight += nodes.Last().Weight;
+
+                    nodes.RemoveAt(nodes.Count - 1);
+                };
+                nodePriorityList.AddNode(parentNode);
+                if (parentNode.Weight > 100)
```
Gustorn: This shouldn't be here, the weights are not percentages

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+            nodes = nodePriorityList.Nodes;
+
+            // Create Tree.
+            while (nodes.Count > 1)
+            {
+                var parentNode = new Node("", 0);
+                // Add the two nodes with the smallest weight to the parent node and remove them from the tree.
+                for (int i = 0; i < 2; i++)
+                {
+                    parentNode.Children[i] = nodes.Last();
```
Gustorn: Make a nice method for this on the priority queue (and don't use the regular list at this point). Something like:

```
var left = nodeQueue.Pop();
var right = nodeQueue.Pop();
nodeQueue.AddNode(Node.CreateBranch(left, right)); // the weight will be calculated in `CreateBranch` as `left.Weight + right.Weight`

File: chapters/data_compression/huffman/code/cs/HuffmanCoding.cs [hunk 12]
```diff
@@ -0,0 +1,185 @@
...
+                var result = nodes.Where(n => n.Key[0] == character).SingleOrDefault();
+
+                if (result == null)
+                    nodes.Add(new Node(character.ToString(), 1));
+                else
+                    result.Weight++;
+            }
+            // Convert list of nodes to a NodePriorityList.
+            var nodePriorityList = new NodePriorityList(nodes);
+            nodes = nodePriorityList.Nodes;
```
Gustorn: Don't do this, just use the queue. In fact, just make `Nodes` private

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,185 @@
...
+                };
+                nodePriorityList.AddNode(parentNode);
+                if (parentNode.Weight > 100)
+                    throw new Exception();
+            }
+
+            return nodePriorityList.Nodes[0];
+        }
+
+        private static Dictionary<char, List<bool>> CreateDictionary(Node root)
```
Gustorn: I know I'm usually a big proponent for performance, but seriously, just use the recursive method here. With C# 7's local functions you can even make a nice, local function for the recursive part

julianschacherpp:
Thanks for the review @Gustorn ! I will enhance the code in the following days.

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
Gustorn: The point isn't compression, it's showcasing thr algorithm. And there was no compression achieved with `List<bool>` either. I'd just go with `String` (as almost all other implementations in the AAA do).

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
Gustorn: Yes. Maybe `char?` do you can set it to `null` in the branches

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
julianschacherpp: You mean since we usually save whole bytes anyway?

Gustorn: Strings in C# are UTF16 encoded which is 2 bytes. If you save the bitstring as a `List<bool>` that's `N` bytes, where `N` is the length of the bitstring. That means you literally only have any compression if the bitstring is a single bit long.

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
julianschacherpp: So the `Node` class?

Gustorn: No, the `Nodes` property on `NodePriorityList`

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
julianschacherpp: Hmm, I'll take a look, but I found the non-recursive version easy to understand. Didn't tried the recursive one tho.

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
julianschacher: I could also create two constructors for the `Node` class. One that takes `key` and `weight` and another one that takes a `leftChild` and `rightChild`. Your approach seems more declarative tho. What do you think?

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
julianschacher: The key of a node is string, so that a parent node's/branch's key is a combination of all it's children's keys.

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,181 @@
...
+            var bitString = CreateBitString(input, dictionary);
+
+            return new EncodingResult(bitString, dictionary, root);
+        }
+
+        public static string Decode(EncodingResult result)
+        {
+            var output = "";
+            Node currentNode = result.Tree;
+            foreach (var boolean in result.BitString)
```
Gustorn: You should probably name this `bit`

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,181 @@
...
+        // The Node class used for the Huffman Tree.
+        public class Node : IComparable<Node>
+        {
+            public Node LeftChild { get; set; }
+            public Node RightChild { get; set; }
+            public List<bool> BitString { get; set; } = new List<bool>();
+            public int Weight { get; set; }
+            public string Key { get; set; }
+            
+            // Creates a leaf. So just a node is created with the given values.
```
Gustorn: These comments are totally superfluous

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,181 @@
+// submitted by Julian Schacher (jspp) with help by gustorn.
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace HuffmanCoding
+{
+    public class EncodingResult 
+    {
+        public List<bool> BitString { get; set; }
```
Gustorn: I still think just having this be a `string` would make things a lot easier

chapters/data_compression/huffman/code/cs/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var result = HuffmanCoding.Encode("aaaabbbccd");
+            // Print dictionary.
+            foreach (var entry in result.Dictionary)
+            {
+                var bitString = "";
```
Gustorn: This is just: `string.Join("", entry.Value.Select(bit => bit ? '1' : '0'))`

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,181 @@
...
+            public NodePriorityList() { }
+            public NodePriorityList(List<Node> givenNodes)
+            {
+                this.nodes = givenNodes.ToList();
+                this.nodes.Sort();
+            }
+
+            public void Add(Node newNode)
+            {
+                var index = ~this.nodes.BinarySearch(newNode);
```
Gustorn: This doesn't seem correct. I think it should be along the lines of:

```csharp
var index = this.nodes.BinarySearch(newNode);
if (index < 0)
    this.nodes.Insert(~index, newNode);
else
    this.nodes.Insert(index, newNode);
```

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
```diff
@@ -0,0 +1,181 @@
...
+            foreach (var boolean in result.BitString)
+            {
+                // Go down the tree.
+                if (!boolean)
+                    currentNode = currentNode.LeftChild;
+                else
+                    currentNode = currentNode.RightChild;
+
+                // Check if it's a leaf node.
+                if (currentNode.Key.Count() == 1)
```
Gustorn: You should really not check for leaf codes this way. It works when the alphabet of whatever you want to encode consists of single characters. You can also run Huffman coding on words, in which case this would faily miserably. The nice solution is to add something like this to `Node`:

```cs
public IsLeaf => LeftChild == null && RightChild == null;

// Then you can just do
if (currentNode.IsLeaf)
{
    // ...
}
```

chapters/data_compression/huffman/code/cs/HuffmanCoding.cs
julianschacher: Why, they are explaining what is done pretty well.

julianschacher:
Ok, everything should be done here now.

But I may need to change the mention, so the merging needs to be delayed a bit. @Gustorn did a great job here again and without him the code would be worse. So I don't know if I should just write "with great help by x" in the files, since the help was big or just "with help by x", so things don't get subjective.
I'm also thinking about writing "reviewed by x". This would give it a whole other touch, which would maybe be quite nice.
What do you guys think?

leios:
I think it's a good idea to mention who reviewed it, so `//reviewed by Gustorn` makes sense.

Gustorn:
I'm not terribly attached to having contributor comments on the top of the files. It's clear from the git history and from the PRs. If someone wants to give some credit, I think something like: `// submitted by [X], thanks to [A], [B], [C]` is more than enough.

julianschacher:
I think, I'll go for `// submitted by [me], thanks to [others] for the help` then. Thanks for the input!

julianschacher:
Should be done now. [put in party emoji here I guess] :)

PR #84, opened 2018-04-19
Jarvis March in Haskell

jiegillet:
Only the code, I didn't include it in the .md file.


leios:
I am sorry it took me so long to get to this PR. I don't see anything wrong with it and can follow the code; however, I have to slightly update the bulk text.

Issue #69, opened 2018-03-01
Should we be able to "link" other code?


ubsan:
For example, in [convolutions][conv], the FFT code is out of date. Should we be able to link in other code? Preferably with checks for when the API breaks?

leios:
Oh, I see what you mean. Yeah, this is a good point of discussion.

I would say that we should link to previous code (if only to point the reader in the direction for more information); however, in the case of fft's I feel it's probably better to use a standard fft library (like fftw), because that's what people will see in practice. If there is no good library to perform the operation already, then we should link to previous code... And I'm not sure the best way to do that.

jiegillet:
I don't think we should link to other code, I like each chapter to be independent. Whenever relevant,  we should link to other AAA chapters, but not rely on specific code.

PR #85, opened 2018-04-19
Add Jarvis March in C#.

julianschacherpp:
### ToDo
- [x] Simplify name of Run-method.
- [x] Make the static class Jarvis March non static.
- [x] Add Scalar product method to vector.
- [x] Make vector a struct.
- [x] Make properties of vector readonly.
- [x] Overload operators for Vector.
- [x] Refactor code and make it simpler.
- [x] Override the Equals and the GetHashCode method for Vector.
- [x] Discuss stuff.
- [x] Improve explanation of `IsLeftOf`.
- [x] Remove redundant operators.
- [x] Don't remove points / simplify code.


chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,81 @@
+// submitted by Julian Schacher (jspp)
+using System.Collections.Generic;
+
+namespace JarvisMarch
+{
+    public class Vector
```
Gustorn: Make this a struct and drop the properties. It has better performance and the properties are completely superfluous for a `Vector` type:

```cs
public struct Vector
{
    public int X;
    public int Y;

    public Vector(int x, int y)
    {
        this.X = x;
        this.Y = y;
    }
}
```

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,81 @@
...
+            var previousPoint = new Vector(initialPoint.X, initialPoint.Y - 1);
+            var currentPoint = initialPoint;
+
+            var notWrapped = true;
+            // Continue searching for the next point of the wrap until the wrap is completed.
+            while (notWrapped)
+            {
+                // Search for next Point.
+                // Set the first vector, which is currentPoint -> previousPoint.
+                var firstVector = new Vector(previousPoint.X - currentPoint.X, previousPoint.Y - currentPoint.Y);
```
Gustorn: I'm not the biggest fan of this naming. First, the `Point` and `Vector` suffix is pretty redundant. Second, `firstVector` means nothing. I'll have to read the Jarvis march algorithm again to come up with a better name.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,81 @@
...
+        public Vector(int xValue, int yValue)
+        {
+            this.X = xValue;
+            this.Y = yValue;
+        }
+    }
+
+    public static class JarvisMarch
+    {
+        public static List<Vector> RunJarvisMarch(List<Vector> points)
```
Gustorn: Redundant naming:

```cs
JarvisMarch.Run(...);
```

is pretty self-explanatory. If you want to follow best practices then don't make these static classes. That way you could potentially implement an interface `IGiftWrapping` and have multiple algorithms that you can swap out at runtime (static classes/methods make this impossible)

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            // Set the intial point to the first point of the list.
+            var initialPoint = points[0];
```
Gustorn: I think we should have a shared library somewhere in the AAA for example with `List`/`IEnumeable` extensions methods (similarly for other languages that need it). These lines just take up a bunch of space in the implementation when by all right they should just be:

```cs
var iniitalPoint = points.MinBy(p => p.X);
```

We should talk about @leios about it

leios: Yeah, this is somewhat related to #69, but I am not sure about the best way to do this in practice.

This is actually why the language choice chapter exists. The hope was to introduce any extra libraries and such people would need to use the language for the purpose of this book, while also benchmarking the languages against each other so people could see their strengths and weaknesses.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
julianschacherpp: The name is not the best, I agree. But I don't know, if getting rid of `Point` and `Vector` is the best. I think they provide quite some clarity. If better names were used tho, the suffix might be redundant. I have to think about the names.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
julianschacherpp: So just use normal classes in this case?

julianschacherpp:
Thanks for the review @Gustorn ! I will target your reviews in the following days as well :)

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
Gustorn: Yeah

julianschacherpp:
I added a ToDoList for this PR.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
julianschacherpp: Yeah, that's something to discuss.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,89 @@
+// submitted by Julian Schacher (jspp) with help by gustorn
+using System;
+using System.Collections.Generic;
+
+namespace JarvisMarch
+{
+    public struct Vector
+    {
+        public readonly int x;
```
Gustorn: Public fields should be Pascal case

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            var notWrapped = true;
+            // Continue searching for the next point of the wrap until the wrap is completed.
+            while (notWrapped)
+            {
+                // Search for next Point.
+                // Set the first vector, which is currentPoint -> previousPoint.
+                var firstVector = previousPoint - currentPoint;
+
+                Vector? nextPoint = null;
```
Gustorn: This implementation feels a bit unelegant

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,69 @@
...
+                currentPoint = nextPoint;
+
+                // Check if the gift wrap is completed.
+            } while (nextPoint != convexHull[0]);
+
+            return convexHull;
+        }
+
+        // Returns true, if p is more left than b, if viewed from a.
+        private bool IsLeftOf(Vector a, Vector b, Vector p)
```
julianschacherpp: Does someone know better names for this parameters? And should I provide a more in-depth explanation of what the method does?

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,69 @@
...
+            this.y = yValue;
+        }
+
+        public override bool Equals(object obj) => obj is Vector v && this.x == v.x && this.y == v.y;
+        public override int GetHashCode() => (17 * 23 + this.x) * 23 + this.y;
+
+        public static bool operator==(Vector a, Vector b) => a.Equals(b);
+        public static bool operator!=(Vector a, Vector b) => !(a == b);
+        public static Vector operator+(Vector a, Vector b) => new Vector(a.x + b.x, a.y + b.y);
+        public static Vector operator-(Vector a, Vector b) => new Vector(a.x - b.x, a.y - b.y);
```
julianschacherpp: Should we leave those operators in here (`-` and `+`). They are nice to have, but not necessary.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
Gustorn: The comment is incorrect. It returns true if `p` is left of the line defined by `a` and `b`.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
Gustorn: If they're not used I say remove them

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
julianschacherpp: I mean it's somewhat correct, but far from perfect. I'll change it.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,69 @@
...
+    public class JarvisMarch
+    {
+        public List<Vector> Run(List<Vector> points)
+        {
+            var convexHull = new List<Vector>();
+
+            // Set the intial point to the point of the list, where the x-position is the lowest.
+            var initialPoint = points.Aggregate((leftmost, current) => leftmost.x < current.x ? leftmost : current);
+
+            convexHull.Add(initialPoint);
```
Gustorn: Just move this inside the do-while loop, it doesn't need to be special cased

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
julianschacherpp: Ok, I see it. It's incorrect.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,69 @@
...
+                // Search for the next point by looking which of the remaining points is the next most outer point (or left point if viewed from currentPoint).
+                nextPoint = points.Aggregate((potentialNextPoint, current) =>
+                {
+                    if (IsLeftOf(currentPoint, potentialNextPoint, current))
+                        return current;
+                    return potentialNextPoint;
+                });
+
+                convexHull.Add(nextPoint);
+                points.Remove(nextPoint);
```
Gustorn: You don't need to remove the point if we don't care about collinear points (and the consensus was that we don't). There are some other reasons why you shouldn't have it here:

1. It kind of messes with the complexity of the algorithm. It won't affect big O notation but you're adding another linear search.
2. It's *very* bad manners do modify an incoming parameter

If you do want to remove the point add something like:

```cs
var remainingPoints = new HashSet<Vector>(points);
```

And change stuff to use that one. But again, removing the point isn't necessary.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                // Search for the next point by looking which of the remaining points is the next most outer point (or left point if viewed from currentPoint).
+                nextPoint = points.Aggregate((potentialNextPoint, current) =>
+                {
+                    if (IsLeftOf(currentPoint, potentialNextPoint, current))
+                        return current;
+                    return potentialNextPoint;
+                });
+
+                convexHull.Add(nextPoint);
```
Gustorn: As said in the other comment, you can just move this to the beginning and then you don't need the `Add` outside of the `do...while` loop

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
Gustorn: No, you were right, it's technically correct. I still think it's better to use more rigorous terms when defining operations like this.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
julianschacherpp: Well, I removed the point from the points we search, since that point isn't needed anymore and won't be used anyway. But you're right with the linear search, it creates a bigger disadvantage than advantage... The question is tho, if I should go for the `HashSet`.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
Gustorn: I don't think it's very useful to include the starting point in the output. It doesn't really give you anything and it makes the algorithm a bit harder to read.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/cs/JarvisMarch.cs
Gustorn: I would just not remove the points

julianschacherpp: Ok, I just exclude that now.

leios:
Thanks for all the work on this guys!

PR #87, opened 2018-04-20
Bogo Sort in Haskell

jiegillet:
Added bogo sort, cleaned out trailing spaces too.
I left the implementation for shuffling because it's not part of core libraries in Haskell.


leios:
I think this implementation is fine. We need to implement Fisher-Yates shuffle in the AAA still, but it's now on the list.

PR #89, opened 2018-04-22
fix the style and highlighting of huffman.cpp

ubsan:
change `class` -> `typename` in template parameter lists; this brings the file in line with other C++ files in the project.

change `lang:"cpp"` -> `lang:"c_cpp"`, as the code highlighter uses `"c_cpp"` for highlighting both kinds of files.


leios:
Man, it is taking me forever to go through these PR's. Sorry about that!

PR #99, opened 2018-05-01
Add a margin at the bottom of every page

Butt4cak3:
When I brought up the AAA a few days ago, a friend of mine brought up that he doesn't like how the chapter text ends at the very bottom of the page. [The Introduction chapter is a good example](https://algorithm-archivists.github.io/chapters/introduction.html) - the last line of text is at the very bottom of the monitor when viewing the page in fullscreen. I see where he's coming from, so I tried adding a margin at the bottom and I think it's nicer this way. What do you all think?

### DIFF
--- a/styles/website.css
+++ b/styles/website.css
@@ -7,5 +7,5 @@
 }
 
 .book .book-body .page-wrapper {
-  margin-bottom: 15em;
+  margin-bottom: 12em;
 }
--- a/styles/website.css
+++ b/styles/website.css
@@ -3,5 +3,9 @@
 }
 
 .book .book-body .page-wrapper .page-inner section.normal .hljs-meta-string {
-  color: #718c00; 
+  color: #718c00;
+}
+
+.book .book-body .page-wrapper {
+  margin-bottom: 15em;
 }


leios:
I agree, I don't like how the chapters end so abruptly. I think this is a good change, but I'll leave it up for a bit in case there is discussion.

Butt4cak3:
It's also about the amount of margin. Should I make it wider or narrower or is it just right?

Butt4cak3:
I don't know. I made it 10em first but I thought it wasn't enough. Let's hear some more opinions.

julianschacherpp:
Maybe 12 or 13 em? I find 15 still a bit too much.

julianschacherpp:
Here's a direct comparison.

**12em**
screen shot 2018-05-05 at 21 03 01_page_margin_12em

**15em**
screen shot 2018-05-05 at 21 03 53_page_margin_15em

In the end it doesn't really matter, but 12em looks a bit better imo.

Butt4cak3:
@julianschacherpp I guess 12em is alright, too. I just thought maybe someone else would comment on it, but nobody seems to care :x

julianschacherpp:
Yeah, disappointing. :/

leios:
I care and agree on 12em

leios:
I think this is a case where everyone agrees we should have some space, but no one really knows how much space.

leios:
Perfect! Merge time?

Butt4cak3:
My suggestion was to wait until one week has passed to allow more opinions, but at this point I'm pretty sure it's not important enough. Go ahead and merge if you want.

julianschacherpp:
I guess so. We could wait to Monday, but I don't expect significant input till then.

PR #106, opened 2018-05-11
Adding graham.c

Gathros:


chapters/computational_geometry/gift_wrapping/graham_scan/code/c/graham.c
```diff
@@ -0,0 +1,114 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+struct point {
+    double x, y;
+};
+
+int cmp_points(const void *a, const void *b) {
+    struct point point1 = *(struct point*)a;
```
Gustorn: You shouldn't copy the points, it's completely unnecessary. Just cast to `struct point*` and compare the points that way

chapters/computational_geometry/gift_wrapping/graham_scan/code/c/graham.c
```diff
@@ -0,0 +1,114 @@
...
+    if (point1.y > point2.y) {
+        return 1;
+    } else if (point1.y < point2.y) {
+        return -1;
+    } else {
+        return 0;
+    }
+}
+
+double ccw(struct point a, struct point b, struct point c) {
```
Gustorn: I don't think `ccw` is the best name here. That would make sense if it returned a `bool`. I'd probably call it `orientation` or something like that.

Gathros:
@Gustorn I changed `ccw` to `is_left_of` and changed `cmp_points`.

chapters/computational_geometry/gift_wrapping/graham_scan/code/c/graham.c
```diff
@@ -0,0 +1,112 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+struct point {
+    double x, y;
+};
+
+int cmp_points(const void *a, const void *b) {
+    if (((struct point*)a)->y > ((struct point*)b)->y) {
```
Gustorn: Having them as variables wasn't a bad idea, I just didn't like the copy:

```c
struct point* pa = (struct point*) a;
struct point* pb = (struct point*) b;

if (pa->y > pb->y) {
// ...
```

julianschacherpp:
Since @Gustorn approved it, it should be fine :)

PR #107, opened 2018-05-11
Removing useless code in tree traversal.

Gathros:


Butt4cak3:
I would love to review/approve this PR and I'm sure I'm not the first to look at it, but the lack of a description makes it kind of difficult to understand what's going on. Why was the code there in the first place? What made it unnecessary?

Butt4cak3:
In addition to what I said above, you seem to have removed the wrong file. The Julia code now says this for every function:
```
# This has not been implemented in your chosen language, so here's Julia code
```

Gathros:
There were two files both identical but one didn't have a main function, so it could not be executed. So I chose to remove it. There really is nothing to add beyond that. I can remove the comments if you would like.

Butt4cak3:
Oh, wow. They're actually identical other than the fact that one has these comments in it. Why not just remove the one *with* comments and keep the .md file unchanged? That would be much cleaner than removing the currently used source file and changing the other one so that it's exactly like the one you just deleted.

Gathros:
I would still need to change the .md file.

Butt4cak3:
@Gathros Sorry, I was a little slow updating my last comment.

Both files are linked in the .md file because they're both used and they both serve different purposes. One is to fill in missing code (e. g. for Python 3) and the other one is the actual Julia example. One needs the comments, the other one doesn't. The file is not unnecessary.

Gathros:
I don't see why there needs to be two implementations with the same language one is fine.

Gathros:
I've updated the .md file to add the warning of `This has not been implemented in your chosen language, so here is the Julia code` @Butt4cak3 .

leios:
Ok. I now know what's going on, sorry.

So there are several languages without implementations for them in the bulk text. In these cases, we are substituting Julia code. That is the purpose of `Tree_example.jl`. In that code, I had to make it clear that the language shown was not the language they wanted, but was instead Julia and I opted to put this in a comment instead of in the bulk text because it seemed easier for people to find.

If we only have one file, then when we list the code at the bottom of the page, the comment "This has not been implemented in your chosen language, so here is the Julia code" will be in the code, which doesn't make sense in that context. It looks to me like both files are necessary unless we update the bulk text to have the phrase "This has not been implemented in your chosen language, so here is the Julia code" just before the code snippets for almost every language that is not julia for certain cases. 

I am fine with either solution. Keeping 2 files with comments is less work, but it ultimately doesn't matter.

Gathros:
@Butt4cak3 I made a [project](https://github.com/algorithm-archivists/algorithm-archive/projects/8) for all the missing examples in Tree Traversal.

leios:
Thanks for doing that guys!

PR #112, opened 2018-05-14
Added Graham Scan in Haskell

jiegillet:
As the title says + some trailing whitespaces in graham_scan.md


chapters/computational_geometry/gift_wrapping/graham_scan/code/haskell/grahamScan.hs
```diff
@@ -0,0 +1,31 @@
+import Data.List (sortOn, minimumBy)
+import Data.Function (on)
+
+type Point = (Double, Double)
+
+angle :: Point -> Point -> Point -> Double
```
Gathros: Is this the same as the old julia function called `graham_angle`?

jiegillet: I'm not sure, I wrote it without looking at the Julia code. Why?
I know that it's little overdoing it since when it's used in the Graham scan the first two points are always fixed, so I could get away with something more ad-hoc. Although it could make the code slightly harder to read?

Gathros: Look at my C code as an example.

jiegillet: An example of what? I see you use polar_angle.

Gathros: Yes in polar_angle I get the polar angle of the point from the point called origin. Then I sort by that angle.

jiegillet: Yes, it looks like it would work. I re-used the function that I wrote in Jarvis March for simplicity, but I suppose I could rework this one.

jiegillet:
Got rid of the angle function altogether, I'm simply using atan2 inside the sort function now.

Gustorn:
You should be able to just use the dot product instead of explicitly computing the angle. This method is not only significantly faster, it's also less likely to be affected by floating point precision issues.

It's also a very well known optimization of the Graham scan algorithm, ideally I think we should mention it in the chapter as well.

leios:
@Gustorn Fair point. I'll make the appropriate changes.

jiegillet:
Suggestion for later: write a chapter where you cover angle measurements :)
For now, I don't think it matters so much, very little of the code in the AAA is optimized anyway. I think I'll leave it like this, you read atan2, you know it's an angle. Discussing optimization possibilities in the chapter is good though.

PR #113, opened 2018-05-15
adding smallscale changes to graham implementation, fft, and thomas.

leios:
Here are a few quick updates to the implementation of graham scan. I also slightly modified the spacing in the fft algorithm and removed a comment that shouldn'e be there and fixed some stuff with the `thomas.md` file.

As a note: we indicated that the readme could use some work and should be split so it's not on algorithm-archive.org anymore. The reason we have the readme as the first page is in part because of how gitbook worked when we were using it. Since we are no longer on gitbook, this shouldn't be an issue anymore.

### DIFF
--- a/chapters/FFT/code/julia/fft.jl
+++ b/chapters/FFT/code/julia/fft.jl
@@ -5,7 +5,7 @@ function DFT(x)
     # We want two vectors here for real space (n) and frequency space (k)
     n = 0:N-1
     k = n'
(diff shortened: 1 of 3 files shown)


jiegillet:
There are many ways of calculating angles, I also used atan2 (although I used 2, I see now that I could simplify it)

Gustorn:
`atan2` is alright for educational purposes but using a `dot` product would be significantly faster

PR #117, opened 2018-05-29
Fix links to not point to the old repository anymore and remove a file that shouldn't be here.

julianschacher:
Every instance of "leios/algorithm-archive" is being replaced with "algorithm-archivists/algorithm-archive".


julianschacher:
@leios What's up with [version_control.md2](https://github.com/algorithm-archivists/algorithm-archive/blob/master/chapters/principles_of_code/version_control.md2)? Is this file important/used?

leios:
@julianschacher That was a mistake. It can definitely be deleted. Thanks!

julianschacher:
@leios Do you want to delete it or should I just include it in this PR?

leios:
I think it's fine to add it to this one

leios:
Should we leave a link to the discord server in the how to contribute chapter?

julianschacher:
Why not? The discord server is the place the main AAA discussion happens. It's also nicer to write there about something, that would otherwise spam a PR. Like a discussion about a detail, etc.
Aside from that, it doesn't hurt to have more people on Discord. It shouldn't be a requirement of course, but it should be a good option.

PR #119, opened 2018-06-03
Added Monte Carlo in Haskell

jiegillet:
Monte Carlo in Haskell + a few typos in monte_carlo.md

### DIFF
--- a/chapters/monte_carlo/code/haskell/monteCarlo.hs
+++ b/chapters/monte_carlo/code/haskell/monteCarlo.hs
@@ -1,0 +1,14 @@
+import System.Random
+
+monteCarloPi :: RandomGen g => g -> Int -> Float
+monteCarloPi g n = count $ filter inCircle $ makePairs 
+  where makePairs = take n $ toPair (randomRs (0, 1) g :: [Float]) 
+        toPair (x:y:rest) = (x, y) : toPair rest
+        inCircle (x, y) = x^2 + y^2 < 1
+        count = (*4) . (/fromIntegral n) . fromIntegral . length 
+
+main = do
+  g <- newStdGen
+  let p = monteCarloPi g 100000 
+  putStrLn $ "Estimated pi: " ++ show p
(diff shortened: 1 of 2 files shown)


Butt4cak3:
Do you use different editors for .md and .hs files? You removed some trailing whitespaces from the .md file, but your Haskell code has trailing whitespace in it. Just wondering how that happened.

chapters/monte_carlo/code/haskell/monteCarlo.hs
```diff
@@ -0,0 +1,14 @@
+import System.Random
+
+monteCarloPi :: RandomGen g => g -> Int -> Float
+monteCarloPi g n = count $ filter inCircle $ makePairs 
+  where makePairs = take n $ toPair (randomRs (0, 1) g :: [Float]) 
+        toPair (x:y:rest) = (x, y) : toPair rest
+        inCircle (x, y) = x^2 + y^2 < 1
+        count = (*4) . (/fromIntegral n) . fromIntegral . length
```
Butt4cak3: This looks super backwards. Is it common to define functions in this way? As someone who doesn't know too much about Haskell, this would be much more intuitive:

```haskell
        count l = fromIntegral (length l) / fromIntegral n * 4
```

jiegillet:
Thanks for the review. I did use different editors, I was teaching a class about vim at the time, so I used it to edit my code get warmed up. I'll review it.

chapters/monte_carlo/code/haskell/monteCarlo.hs
jiegillet: Yes, you're right, I sometimes get carried away with point-free style (not writing the parameters explicitly), I usually like it better, but in this case your suggestion is much better.

Butt4cak3: If you want to change that line, go ahead and add another commit. If not, say the word and I'll merge.

PR #121, opened 2018-06-04
Adding a C Monte Carlo implementation.

Gathros:


Gathros:
@jiegillet I dont mind either way.

leios:
@jiegillet I added in a variable radius for monte carlo because with the `radius` variable, the formula in the end makes more physical sense. I was afraid that just leaving it as 1 would make it harder to understand what we are actually doing.

That said, I am fine with either way.

jiegillet:
Leaving `radius` in the expression for `in_circle` makes sense if you want to show the meaning, but using a value of `0.5` doesn't, since precision is lost. Just use `1`, I don't think it makes it harder to understand.

leios:
Yeah, I don't disagree. We could also just set the radius function parameter `function in_circle(pos_x, pos_y, radius = 1)` as a default unless otherwise specified.

leios:
thanks a bunch, sorry it took me so long!

PR #122, opened 2018-06-06
Fix minor typos in introductory pages

chrisb2244:
There are some (typically one letter) typos in the introductory pages of the book. I've tried to fix the ones I can see here.

I also think the discussion of floating point numbers is a little unclear, and that in several places the base is given as 10 but the number of digits for exponent and mantissa match the base 2 binary IEEE754-1985 standard (i.e. the base should be 2...). I've corrected this in a separate commit in case it's not what you intended.

Some of the further examples list the biased exponent straight away - I only discussed the difference explicitly in the first case.


chrisb2244:
I'm not sure the best way to go about this - I can close this PR and open another, but I can't seem to change the branch I want to merge. I expect that probably creates more work, but perhaps it's the appropriate choice?

The changes specific to this issue are now in https://github.com/chrisb2244/algorithm-archive/tree/pr-122

Gathros:
I would just keep it as it is since the commit is in the topic of what the PR wants to do.

For the future, you would close the PR and the branch then make two new branches based of the master branch, changes the thinkgs you want then make two new PRs.

chapters/principles_of_code/building_blocks/functions.md
```diff
@@ -1,7 +1,7 @@
 ### Functions
 
 Functions make sense from a mathematical point of view. 
-$$f(x) = 2x+5$$ returns a value for x at any point you want. 
+$$f(x) = 2x+5$$ returns a value for $$f(x)$$ at any point i$$x$$ that you want.
```
Butt4cak3: You introduced a new typo in this line. There's an i before `$$x$$`. Are you a vi(m) user by any chance?

chrisb2244:
Added some additional changes to the convolutions and sorting chapter texts. Hopefully no additional typos here. I don't think the change from "convolutional theorem" to "convolution theorem" should be controversial, but I can remove it if requested. There's still a use of "convolutional" to describe "convolutional methods", which may or may not be desired.

Butt4cak3:
Looks good! Given that it's usually @leios' job to manage the actual chapter texts, I'll let him have the honor of actually merging this PR. That said, there really is no reason not to do it. Every change is an improvement.

@chrisb2244 As this is the first time I'm seeing you here: Welcome to the team!

PR #124, opened 2018-06-06
Fixing stable_marriage.c and convolutions.c

Gathros:


chapters/decision_problems/stable_marriage/code/c/stable_marriage.c
```diff
@@ -1,118 +1,121 @@
...
-        for (size_t i = 0; i < grp_size; ++i) {
-            if (men[i].prtnr == (grp_size + 1)) {
-                cont = true;
-                break;
-            }
+    while (bachelors_size > 0) {
+        struct person *man = bachelors[bachelors_size - 1];
+        struct person *woman = &women[man->prefers[man->index]];
+
+        if (woman->partner == NULL) {
```
Gustorn: I think a more conventional way to check for this would be `if (!woman->partner)`

Gathros: I prefer your version but what is better?

PR #125, opened 2018-06-06
Fix example to use 1-based indexing in Julia example

### LINKED ISSUE #123: Complexity notation page inconsistent with array indexing
Some parts of the complexity notation page (chapters/principles_of_code/notation/notation.md) list arrays with base 0 indexing, others use things like `a[1]` to refer to the first element.

Although it's possible they could be from different languages, the examples are less likely to be misleading if all arrays are 0 based.

The linear time example also lists i = 0 : length, which would probably lead to an out of bounds error in languages providing them (not sure if this includes Julia) or runtime exception/similar.

chrisb2244:
Fixes #123 according to description/instructions given in the issue.

### DIFF
--- a/chapters/principles_of_code/notation/notation.md
+++ b/chapters/principles_of_code/notation/notation.md
@@ -73,7 +73,7 @@ Let's consider the following function:
 
 ```julia
 function linear(a::Array{Float64})
-    for i = 0:length(a)
+    for i = 1:length(a)
         println(a[i])
     end
 end


Gathros:
You fixed this in #122.

chrisb2244:
Huh... My bad. Guess I misunderstood how the PR would work. Didn't know an existing PR would automatically update with new commits. Deliberately cherry-picked this one onto a branch checked out from algorithm-archivists to avoid making it contain the commits in #122.

Gathros:
if you add a commit to a branch that has a PR then the commit gets added to the PR. I'll close this PR for you.

PR #127, opened 2018-06-10
Fix broken link on convolutions chapter

tylercd100:


leios:
Yeah, this is a simple fix, so I don't think there is much discussion needed before merging it. Thanks a bunch for catching that!

PR #128, opened 2018-06-16
Added Monte Carlo in Clojure

earthfail:
sorry guys for not contributing much but some topics are really hard like FFT


earthfail:
btw could someone explain why the division by r<sup>2</sup>
here [julia monte carlo](https://github.com/algorithm-archivists/algorithm-archive/blob/e0ece15cbfc1ef66623c724dcb63f88ae4a364c1/chapters/monte_carlo/code/julia/monte_carlo.jl#L23)

leios:
You are essentially solving for `pi` in the formula: `Area = pi*r^2` so `pi = Area / (r^2)`. Because you are only using a quarter, the final formula is `pi = Area / (4*r^2)`

earthfail:
[haskell monte carlo](https://github.com/algorithm-archivists/algorithm-archive/blob/e0ece15cbfc1ef66623c724dcb63f88ae4a364c1/chapters/monte_carlo/code/haskell/monteCarlo.hs#L8) yeah but What is confusing is the haskell code it doesn't even have r<sup>2<\sup> , and honesly I tried both formulas and got different results

leios:
That's because he set the `radius` to be 1 and excluded it from the formula.

earthfail:
sorry @leios  it  doesn't seem to work with me when using different radiuses so I'm sticking with my code

earthfail:
hey @Gathros  how did you add a label

Gathros:
It's on the right hand side, just click labels and select the label.

earthfail:
well from what I understand `Area` is `ration * Area of the square`
`ratio= pi_count/n`
`Area of the square = r^2`

earthfail:
My Code runs fine and gets close results to pi. Just wanted to point out what I didn't understand

leios:
Hey @earthfail Sorry it's taking so long for this PR, we don't have enough people who understand clojure to review it.

earthfail:
That got me think should I link to a guide where people can get a repl or something to play with code.

jiegillet:
@earthfail do you still have trouble understanding the normalization? I can try to explain if you want. In my Haskell code I hardcoded the radius to 1, since it's the best value anyway, but I can justify it ^^

jiegillet:
After looking at your code, I understand why you don't need to renormalize. leios is using a random number between 0 and 1, you are using a random number between 0 and r. The scale of your box and your circle are matching so you don't need to renormalize.  Does that make sense?

earthfail:
Thanks @jiegillet , you guys are the best

earthfail:
sorry @Butt4cak3  I tried your method from discord but couldn't make it less than two commits
@jiegillet   tell me if that what you meant

have a nice weekend guys

Butt4cak3:
Uhm... What's that "image modification" commit? Are you running Windows? Could you tell me what the output of `git --version` is? I assume this has to do with our new *.gitattributes* and a bug in an old version of git.

jiegillet:
Thanks for the modification, @earthfail, that's what I wanted.
What no one wants is the modification of 30+ images, so you have to find a way to get rid of these :)

earthfail:
Do you think if I remove my remote repo it will work?

leios:
@earthfail You will probably need to revert your changes to a previous state before the images were modified, and then pull the latest changes from the AAA upstream repo. Removing your repo doesn't change the commit history.

Butt4cak3:
@earthfail Yeah, that version of git is a bit old. We changed our `.gitattributes` file and there's a bug in older versions of git (including the one you have installed) that would change the line endings of binary files, which is nonsense. That means you added a commit that converts all the "line endings" of the images to `lf`.

Removing your repo won't work as long as you have the old version of git installed. It will always mark the images as modified. I'm afraid the only way to get around this is to update git.

PR #130, opened 2018-06-26
Implement bubble sort in Ruby

ArunSahadeo:


Butt4cak3:
@jiegillet It looks like you know Ruby well enough to take this PR, so I assigned it to you.

andreaswachowski:
For the sake of consistency, I'd suggest to modify the code such that it is accepted by [rubocop](https://github.com/rubocop-hq/rubocop). Of course, this is a matter of preference and you might have intentionally abstained from using it. (I am also just passing by after finding this project via Youtube's 3Blue1Brown channel, so I don't know where this project stands with regard to Ruby code conventions.)

The parentheses for the `main()` function (ie a function without parameters) strike me as particularly unconventional in Ruby. Also, the `return` keyword may be omitted, and Rubocop suggests to prefer `each` over `for`. Eventually, the result would be:

```ruby
#!/usr/bin/env ruby

def bubble_sort(arr)
  (0..arr.length - 1).each do
    (0..arr.length - 2).each do |k|
      next unless arr[k] > arr[k + 1]
      arr[k + 1], arr[k] = arr[k], arr[k + 1]
    end
  end

  arr
end

def main
  range = [200, 79, 69, 45, 32, 5, 15, 88, 620, 125]
  puts "The range before sorting is #{range}"
  range = bubble_sort(range)
  puts "The range after sorting is #{range}"
end

main
```

jiegillet:
Thanks for the input @andreaswachowski, I haven't written Ruby code in a while, so your suggestions are welcome. I also prefer`each` over `for`, although I'm not sure that `next unless` is better than a simple `if` in this case. No big deal though ^^

chapters/sorting_searching/bubble/bubble_sort.md
```diff
@@ -20,6 +20,8 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 [import, lang:"haskell"](code/haskell/bubbleSort.hs)
 {% sample lang="cpp" %}
 [import, lang:"c_cpp"](code/c++/bubblesort.cpp)
+{% sample lang="rb" %}
+[import, lang:"ruby"](code/ruby/bubble.rb)
```
jiegillet: Let's skip the first two line, an merely `import:3-7`

julianschacher: Shouldn't it be `import:3-12`? So that the whole function is included.

jiegillet: Now that I look again, it's `import:3-15` haha. Anyway, the line count will change after more commits, the point is to just import the function :)

andreaswachowski:
@jiegillet I agree with you that the `next` is debatable. I followed rubocop's suggestion here, but as mentioned often that's a matter of preference.

ArunSahadeo:
@andreaswachowski I implemented your changes except for the next, which I did not think was necessary.

ArunSahadeo:
Hey @jiegillet I made a forced update to my feature branch, should be up to date now (been busy at work haha)

PR #131, opened 2018-06-27
Added the java version of the bubble algorithm

xam4lor:


Gathros:
Can you add an example like [this](https://github.com/algorithm-archivists/algorithm-archive/blob/master/chapters/sorting_searching/bubble/code/c/bubble_sort.c)

xam4lor:
In Java we must have all of the code inside a class like this example :
`
package test;

public class Main {
	public static void main(String[] args) {
             // blabla
        }
}
`

Should I add the class structure or just the functions inside it ?
(Sorry for my bad English, I'm French ;) )

Gathros:
Just wrap a class around `main` and `BubbleSort` so it's executable.

chapters/sorting_searching/bogo/code/java/bogo.java
```diff
@@ -0,0 +1,47 @@
...
+			System.out.print(test[i] + " ");
+		}
+
+
+	    bogoSort(test);
+
+
+	    System.out.println("\n\nSorted array :");
+	    for (int i = 0; i < test.length; i++) {
+			System.out.print(test[i] + " ");
```
Butt4cak3: You use `System.out.print()` here, which is fine to print the list, but you should add an empty `System.out.println()` or `System.out.print("\n")` at the end so that there's a line break in the end. Otherwise the output looks like this (at least on Linux):
```
marius@DasPC:java> java Bubble
Unsorted array :
20 -3 50 1 -6 59

Sorted array :
-6 -3 1 20 50 59 marius@DasPC:java>
```

See how the prompt `marius@DasPC:java>` is on the same line as the last line of output?

chapters/sorting_searching/bogo/code/java/bogo.java
```diff
@@ -0,0 +1,47 @@
+public class Bogo {
+	// The shuffle() function can be found in code/java/bogo.java
```
Butt4cak3: We are already in code/java/bogo.java.

Butt4cak3:
@xam4lor Hold on there for a second! I recommend not adding more and more stuff to this pull request. The more algorithms you add, the more clustered the conversation will be and the harder it will be to review the pull request.

Bubble Sort and Bogo Sort in one PR is fine, but I recommend creating separate PRs for separate algorithms. Otherwise it will take ages to get this merged.

Alternatively, you can add a few algorithms and we'll wait until you're done with all of them. Then we can review all of them at once. In that case, just add stuff and when you're done with everything, mention me with @butt4cak3 and I'll get back to you.

Your call.

julianschacher:
@xam4lor Do you mind to move the C# commit into a separate PR?

chapters/sorting_searching/bogo/code/java/bogo.java
xam4lor: That was for the website implementation ! The user don't see the shuffle() methode !

xam4lor:
Okay okay 🤣
I'm really new with opensource projects !!
Can somebody clean and split all of the requests and things and make the good changes please ??

julianschacher:
@xam4lor do you mind joining https://discord.gg/2kWgSWB ?
I don't know, if we can actually split things up or if you need to do so. (@Butt4cak3 ?)

chapters/sorting_searching/bogo/code/java/bogo.java
```diff
@@ -0,0 +1,48 @@
...
+
+
+        bogoSort(test);
+
+
+        System.out.println("\n\nSorted array :");
+        for (int i = 0; i < test.length; i++) {
+            System.out.print(test[i] + " ");
+        }
+	System.out.println("\n");
```
Butt4cak3: You don't need both `println` and `"\n"`. `println` already automatically adds a `"\n"` at the end of the string. Using `System.out.println("\n")` will actually print *two* line breaks.

Go for either `System.out.print("\n")` *or* `System.out.println()`. I prefer the latter, but it's up to you.

chapters/sorting_searching/bubble/code/java/bubble.java
```diff
@@ -0,0 +1,31 @@
+public class Bubble {
+	static void bubbleSort(int[] arr) {
```
Butt4cak3: This entire file is still indented with tabs. Again, not inherently bad, but you should replace these with 4 spaces each for consistency.

chapters/sorting_searching/bubble/code/java/bubble.java
```diff
@@ -0,0 +1,31 @@
...
+	    for (int i = 0; i < test.length; i++) {
+			System.out.print(test[i] + " ");
+		}
+
+		bubbleSort(test);
+
+	    System.out.println("\n\nSorted array :");
+	    for (int i = 0; i < test.length; i++) {
+			System.out.print(test[i] + " ");
+		}
```
Butt4cak3: Add an empty `System.out.println()` here to fix the output, just as in bogo.java.

chapters/sorting_searching/bogo/code/java/bogo.java
xam4lor: I don't really know if it's better with one or two spaces ...

xam4lor:
So I think it's better if all of the codes are the same in the way of thinking, because the user can compare different languages !

Butt4cak3:
Everything is looking good now. Let's merge and then realize what a massive mistake we have done! \o/

PR #135, opened 2018-06-27
Added .java files to .editorconfig

Butt4cak3:
I don't know how I managed to forget Java, but here it is. I set it to 4 spaces for indentation for all .java files.


julianschacher:
Seems good, you can merge.

PR #136, opened 2018-06-27
Added Rust implementation of bubble sort.

penpalperson:
Added a Rust implementation of bubble sort using the `rand` crate as found at https://crates.io/crates/rand.

If the random number generation for making a vector looks really esoteric, I'm willing to add another example with an easier-to-read sample vector!


Butt4cak3:
I don't think we have many Rust people who can review PRs on a regular basis. Prepare for this one to last for a while. It will get merged eventually, though.

chapters/sorting_searching/bubble/code/rust/bubble_sort.rs
```diff
@@ -0,0 +1,30 @@
...
+use rand::{thread_rng, Rng}; // Used for random number generation
+use rand::distributions::Uniform; // Used for a uniform distribution
+
+fn bubble_sort(a: &mut Vec<u32>) {
+    let n = a.len();
+
+    for _ in 0..n {
+        for j in 1..n {
+            if a[j - 1] > a[j] {
+                // Swaps values at indices j - 1 and j
```
Gustorn: Rust has a `swap` method on slices, I would use that instead of the explicit swap written here. It's faster and it's more descriptive

chapters/sorting_searching/bubble/code/rust/bubble_sort.rs
```diff
@@ -0,0 +1,30 @@
+extern crate rand; // External crate that provides random number generation tools
+
+use rand::{thread_rng, Rng}; // Used for random number generation
+use rand::distributions::Uniform; // Used for a uniform distribution
+
+fn bubble_sort(a: &mut Vec<u32>) {
```
Gustorn: This should take `&mut [u32]` since it doesn't any functionality on Vec, and taking slices is strictly more general (for example it could allow you to partially sort a vector).

chapters/sorting_searching/bubble/code/rust/bubble_sort.rs
```diff
@@ -9,8 +9,7 @@ fn bubble_sort(a: &mut [u32]) {
     for _ in 0..n {
         for j in 1..n {
             if a[j - 1] > a[j] {
-                // Swaps values at indicies j - 1 and j 
-                a.swap(j, j - 1);
+                a.swap(j, j - 1); // Swaps values at indicies j - 1 and j
```
Butt4cak3:
```
doSomething(); // does something
```

I'm sorry. I had to. :D

Gustorn: Yeah, this comment is not necessary

Issue #134, opened 2018-06-27
Error in Tree Traversal code c#


xam4lor:
The function DFSStack() for the Tree Traversal's example is missing a parameter of type "Node" called "root".

julianschacher:
I'll look into it and come back to you! :)

julianschacher:
You're right. Currently the `root` gets set by the constructor, which will just make up a "fake" Tree.
Making the methods accessible from the outside and really use the algorithm is a good idea actually. We should probably switch the algorithm over to do so.

xam4lor:
I think so ! It's not very clear, the way it's made right now !

julianschacher:
Yeah it's definitely a flaw! I'll go ahead and change it. Or do you want to (I don't want to take away your issue)?

xam4lor:
Let's go do it by myself ;)

julianschacher:
Yes, but I currently have a PR changing TreeTraversal open. Since you know C#, would you mind to review it and work on the new master? Otherwise you can of course just work on the current master and I update my PR. (PR: https://github.com/algorithm-archivists/algorithm-archive/pull/120)

xam4lor:
I'm sorry but I don't really know how to do that :/

PR #137, opened 2018-06-27
Added root parameter

xam4lor:
Added root parameter instead of having a global variable. @see #134.

### DIFF
--- a/chapters/tree_traversal/code/cs/Tree/Program.cs
+++ b/chapters/tree_traversal/code/cs/Tree/Program.cs
@@ -7,14 +7,15 @@ namespace TreeTraversal
     {
         static void Main(string[] args)
         {
+            Node root;
             Console.WriteLine("TreeTraversal");
             var tree = new Tree(3, 3);
             Console.WriteLine("StartDFSRecursive:");
-            tree.StartDFSRecursive();
+            tree.StartDFSRecursive(root);
             Console.WriteLine("DFSStack:");
-            tree.DFSStack();
+            tree.DFSStack(root);
             Console.WriteLine("DFSQueue:");
-            tree.BFSQueue();
+            tree.BFSQueue(root);
         }
     }
 }
--- a/chapters/tree_traversal/code/cs/Tree/Tree.cs
+++ b/chapters/tree_traversal/code/cs/Tree/Tree.cs
@@ -12,14 +12,12 @@ namespace TreeTraversal
             public int Id { get; set; }
         }
 
-        private Node root;
-
-        public Tree(int depthCount, int childrenCount)
+        public Tree(int depthCount, int childrenCount, Node root)
         {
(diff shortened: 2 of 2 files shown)


xam4lor:
May be do that for the full example but I think that It's more comprehensive for the website which shows functions 1 by 1.

Gustorn:
I disagree, every language should implement the algorithms in a way that's idiomatic for _that_ specific language. For example you won't see parallels between the C and Haskell implementations (and you really shouldn't).

xam4lor:
The problem is that if we do what you said, on the website a method like `CreateTree(int depthCount, int childrenCount)` is going to use a parameter called `root` but the user cannot know what mean this parameter, where it was instanciate, ...
But I agree with you that it's not the better way for an object-oriented language but this will do the job, at least for the functions isolated on the website !

Gustorn:
I still disagree. If anyone wonders where `root` comes from they can just check the full code example. The main selling point of OO languages is encapsulation, we should follow best practices regardless of the format of the book. Also people who are at least somewhat competent in C# will probably guess where that ariable comes from. If you really want you can use it as `this.root` if you want to make it _really_ obvious.

xam4lor:
I think that the users reading this book should understand how it works directly without having to check the full code ! Using `this.` is really obvious so why not create a comment just above the first use of the parameter `root` ?

julianschacher:
@Gustorn We could maybe move `Node` in a separate file, make an interface for it and let the algorithm use the interface; that way it can get used by all `Node` classes using the interface.
This is the only way I can think of on how to implement the functionality of having a root parameter.
The question is tho, if we want to make the algorithm really that complex (but then also usable).

Gustorn:
The thing is you _don't_ want the root parameter. It makes no sense to have it (and as I said, you don't really need the `Node` class either).

xam4lor:
If think it makes more sense to have a node parameter than a root node variable.

Gustorn:
You don't need a root node variable.

```cs
public class Tree {
    public int Value { get; set; }
    private List<Tree> Children { get; set; }
}
```

Gustorn:
I'll obviously vote for my solution :) My argument for keeping the name `Tree`: it's a recursive datastructure anyway, so it makes sense to call it `Tree`. It also reflects that no matter which child/node/etc. you take you still have a valid tree.

But that particular detail is pretty inconsequential, I'm fine with either `Node` or `Tree`.

xam4lor:
I have an other solution which is made just of one 'code' (basically two file but that's because of the language) in Java for this example. The root parameter is a variable in the class himself but is also passed as a parameter to a specific function. Look at the issue #149 and tell me what do you think of that !

julianschacher:
@xam4lor Ok, but why should we have the parameter, if the root is already in the class & public? That doesn't make much sense to me. Either one or the other I would say.
What do you think @Gustorn ?

PR #138, opened 2018-06-28
Adding Python 3 euclidean example

KerimovEmil:
Adding Python 3 euclidean example

### DIFF
--- a/chapters/euclidean_algorithm/code/python3/euclidean_example.py
+++ b/chapters/euclidean_algorithm/code/python3/euclidean_example.py
@@ -1,0 +1,28 @@
+# submitted by KerimovEmil
+def euclid_mod(a, b):
+
+    a = abs(a)
+    b = abs(b)
+
+    while b > 0:
+        a, b = b, a % b
+
+    return a
+
+
+def euclid_sub(a, b):
+
+    a = abs(a)
+    b = abs(b)
+
+    while a != b:
+        if a > b:
+            a = a - b
+        else:
+            b = b - a
+
+    return a
+
+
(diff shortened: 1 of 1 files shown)


chapters/euclidean_algorithm/code/python3/euclidean_example.py
```diff
@@ -0,0 +1,28 @@
+# submitted by KerimovEmil
+def euclid_mod(a, b):
+
```
Gustorn: I don't think it's very usual to start a function with an empty line (at least not according to pep8). Other than that the implemrntation looks good.

VikingScientist: If you do fix this (which I think you should), make sure to change it in the python2-implementation as well

VikingScientist:
This is now in conflict with #141 and #145

chapters/euclidean_algorithm/code/python3/euclidean_example.py
KerimovEmil: I can make this change in both files.

jiegillet:
You can remove the python 2 code and replace mentions of python3/py3 by python/py

jiegillet:
Honestly, after #168, all this PR does is remove a blank line and add your name. I'm sorry @KerimovEmil the Python side of the AAA was taken by a sudden storm and chaos ensued. I think I will close this PR, hopefully you can find some other chapter to contribute to.

leios:
@KerimovEmil I think #168 just changed a few lines, but we are no longer splitting python 2 and python 3, so there is no need for two separate python versions.

PR #139, opened 2018-06-28
Add d language example to bubble sort

scarface-one:
This will add an example implementation of bubble sorting in d, based on the C++ implementation.

### DIFF
--- a/book.json
+++ b/book.json
@@ -78,6 +78,10 @@
 				{
 					"lang": "LabVIEW",
 					"name": "LabVIEW"
+				},
+				{
+					"lang": "d",
+					"name": "D"
 				}
 
 			],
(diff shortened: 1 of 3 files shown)


book.json
```diff
@@ -79,6 +79,10 @@
 					"lang": "LabVIEW",
 					"name": "LabVIEW"
 				}
```
Butt4cak3: You have to add a comma after this `}` or it's not valid JSON.

scarface-one: fixed, thanks!

Butt4cak3:
I don't think we have any D programmers here who could properly review your pull request, but I will take a look at it later today (I'm in Europe) and see what I can do. Bubble Sort is not that complicated.

scarface-one:
Thats fine by me. You can find info on the D language here : https://dlang.org/ , and the documentation for the standard library is here : https://dlang.org/phobos/index.html . To build and the file, you can use `dmd bubble_sort.d && ./bubble_sort`, or to make it easier, `rdmd bubble_sort.d`

PR #143, opened 2018-06-28
Added rust version of bogo sort

Jess3Jane:


chapters/sorting_searching/bogo/code/rust/bogosort.rs
```diff
@@ -0,0 +1,35 @@
...
+fn is_sorted(arr : &Vec<i32>) -> bool {
+    for i in 1..arr.len() {
+        if arr[i-1] > arr[i] {
+            return false;
+        }
+    }
+    true
+}
+
+fn bogo_sort(arr : &mut Vec<i32>) {
```
Gustorn: This should take a `&mut [i32]` instead since you don't need any of the `Vec` functionality.

chapters/sorting_searching/bogo/code/rust/bogosort.rs
```diff
@@ -0,0 +1,35 @@
...
+    while !is_sorted(arr) {
+        shuffle(arr)
+    }
+}
+
+fn main() {
+    let mut v = vec!(1,2,3,4,5);
```
Gustorn: I think most everyone uses `let mut v = vec![1, 2, 3, 4, 5]` instead of the regular parens. Also make sure to have spaces after the commas to conform to the style guide (or you can just run `rustfmt`)

chapters/sorting_searching/bogo/code/rust/bogosort.rs
```diff
@@ -0,0 +1,35 @@
+// Submitted by jess 3jane
+
+extern crate rand;
+
+fn shuffle(arr : &mut Vec<i32>) {
```
Gustorn: `rand` already has a shuffle function, just use that and delete this one

Gustorn:
You were faster the slice changes than I was with the review :) Still,

chapters/sorting_searching/bogo/code/rust/bogosort.rs
```diff
@@ -0,0 +1,33 @@
+// Submitted by jess 3jane
+
+extern crate rand;
+
+fn shuffle(arr : &mut [i32]) {
```
Gustorn: You still don't really need this function, just use the one that `rand` provides.

Butt4cak3:
@Gustorn Pretty sure you're the only active contributor who knows Rust well enough.

Gustorn:
Fair enough

PR #144, opened 2018-06-28
Added python3 to Monte Carlo Integration

jonasvandervennet:


chapters/monte_carlo/code/python3/monte_carlo.py
```diff
@@ -0,0 +1,19 @@
...
+    for i in range(n):
+        count += in_circle(random.uniform(0,radius),random.uniform(0,radius),radius)
+    estimated_area = 4*(count/n)*radius**2
+    exact_area = math.pi*radius**2
+    print("percent error: %.10f%%"%(abs(estimated_area-exact_area)/exact_area*100))
+    return estimated_area
+
+#estimate pi by integrating the unit circle
+estimated_area = monte_carlo(10**6, 1)
+print("Estimation for pi: %.10f"%estimated_area)
```
jiegillet: Isn't it preferred to use string interpolation? `"Estimation for pi: {}".format(estimated_area)`

chapters/monte_carlo/code/python3/monte_carlo.py
```diff
@@ -0,0 +1,19 @@
...
+def in_circle(x, y, radius):
+    return x**2 + y**2 < radius**2
+
+# function to integrate a circle with given radius via monte carlo integration
+def monte_carlo(n, radius):
+    count = 0
+    for i in range(n):
+        count += in_circle(random.uniform(0,radius),random.uniform(0,radius),radius)
+    estimated_area = 4*(count/n)*radius**2
+    exact_area = math.pi*radius**2
```
jiegillet: I would do the solution comparison outside of `monte_carlo`, it's not part of the algorithm.

chapters/monte_carlo/code/python3/monte_carlo.py
Gustorn: Even better, we could use format strings: `f"Estimation for pi: {esimated_area}"`. I don't remember off the top of my head how to specify precision but it shouldn't be that difficult.

jiegillet:
I believe the code is incorrect. If you pick your random numbers between 0 and radius, you don't need to renormalize by the radius later on, you only need to do so if you pick between 0 and 1.
Overall, I don't like having this radius parameter laying around, everyone is getting confused about it and it doesn't add anything except for making the `in_circle` function more general.

jonasvandervennet:
If the only goal is to approximate pi, multiplying with radius squared is indeed unnecessary. My thought was  that the example for monte carlo integration should contain a function to determine the area of the circle, rather than just pi. Also the use of the radius parameter is to stay consistent with the Julia code already present.

chapters/monte_carlo/code/python3/monte_carlo.py
VikingScientist: `print( "Estimation for pi: {:.10f}".format(estimated_area) )`

[string format syntax](https://docs.python.org/3/library/string.html#format-string-syntax)

leios:
I prefer the radius to be in the code (but it ultimately doesn't matter). Having the `radius` variable makes it clearer what the formula is supposed to be and doesn't look like a bunch of magic numbers.

the `in_circle(...)` function is also there so people see that the code is easily extended to other shapes and objects.

jiegillet:
@jonasvandervennet No, you misunderstand, beyond the question of using `radius`, your code is incorrect. I just ran it with a radius of `0.1` and I got

``` 
percent error: 98.999539%
Estimation for pi: 0.031430
```

It's a terrible approximation ^^

You can fix it by either replacing `random.uniform(0,radius)` by `random.uniform(0,1) ` (but in this case, you can only pick radiuses smaller than 1, and if you do you will lose precision, BTW @leios your Julia code has the same pitfall) or changing `return 4*(count/n)*radius**2` by `return 4*(count/n)` (but then why even bother with a free parameter radius?).

My suggestion: keep the parameter `radius` in `in_circle`, but don't include it in `monte_carlo` as a free parameter (and you always call `in_circle` with radius 1). @leios, you should do the same.

leios:
hey @jonasvandervennet I apologize for this. The initial julia code was confusing. I had a discussion with @jiegillet and he convinced me to re-write the code without the `radius` parameter in #159. 

If it is easier to use that code as a base, please feel free to do that. Sorry again for the confusion!

GuyPozner:
I've submitted a pull request to @jonasvandervennet fork, mainly fixing readability issues.

chapters/monte_carlo/monte_carlo.md
```diff
@@ -41,6 +41,8 @@ each point is tested to see whether it's in the circle or not:
 [import:2-8, lang:"julia"](code/julia/monte_carlo.jl)
 {% sample lang="hs" %}
 [import:7-7, lang:"haskell"](code/haskell/monteCarlo.hs)
+{% sample lang="py3" %}
```
leios: We just changed some things and settled on `py` instead of `py3`, basically just requiring code to be submitted in python3. Would you be able to change this to `py`?

chapters/monte_carlo/monte_carlo.md
```diff
@@ -78,6 +80,9 @@ Feel free to submit your version via pull request, and thanks for reading!
 {% sample lang="hs" %}
 ### Haskell
 [import, lang:"haskell"](code/haskell/monteCarlo.hs)
+{% sample lang="py3" %}
```
leios: Same here, change to `py`

chapters/monte_carlo/code/python3/monte_carlo.py
```diff
@@ -0,0 +1,20 @@
+#example submitted by Jonas Vander Vennet, Python 3.5.2
+
+import random, math
+
+# function to determine whether an x, y point is inside a circle with given radius
+def in_circle(x, y, radius):
```
leios: We slightly modified most implementations to hide the `radius` variable because it was confusing some people and we specified that we were working on the unit circle. Would you be able to remove the `radius` variable?

chapters/monte_carlo/code/python3/monte_carlo.py
```diff
@@ -0,0 +1,20 @@
...
+    return x**2 + y**2 < radius**2
+
+# function to integrate a circle with given radius via monte carlo integration
+def monte_carlo(n, radius):
```
leios: Also remove `radius` here

leios:
Actually, I just checked the PR from @GuyPozner . It seems to make all the necessary updates. If you merge that PR into your own fork @jonasvandervennet , we should be mostly good to go here.

jiegillet:
Dear @jonasvandervennet, you've submitted your PR about 8 days ago, but haven't responded to our reviews for 7 days. In the mean time 3 new PRs were submitted for Monte Carlo in Python, we've turned them all down, even though they were closer to a finished product than your PR. I even believe on PR was submitted to your branch.
We'd like to add the algorithm on the AAA, so please show some sign of life within 24h of this comment. If you don't reply, we will move forward with someone else's PR. Cheers.

jiegillet:
Time's up.

Issue #141, opened 2018-06-28
Should we have seperate python2 and python3 codes?


VikingScientist:
The current state of python code:

* Tree Traversal (python2 and unfinished python3)
* Euclidean Algorithm (python2)
* Bubble Sort (python3, but unlisted)
* FFT (python2)

My question is if we really need to split the code into separate implementations for python2 and python3. It would probably be a much better design to just have a `python` implementation, and write it so it runs on both versions. For practical purposes this means that it is the `print`-statement that is going to be the most obvious problem to combine, but this very neatly fixed by including the following code at the start
``` python
from __future__ import print_function
```
which allows python2 to write `print('Hello World')`. The second most obvious difference is that the division operator `/` is interpreted as integer-division in python2 while always floating point division in python3. Again: we can write compatible code by including the `__future__` library
``` python
from __future__ import division
```

Butt4cak3:
I would honestly not have a problem with going for Python 3 completely. As far as I know, unless you're working with legacy code, there is not much reason to use Python 2 at all. Let's hear some more opinions.

jiegillet:
Honestly, my opinion on this is: dump Python 2 altogether. [It's retiring anyway](https://pythonclock.org). There is so little difference, any beginner should start learning Python 3 and we should encourage that.

Gustorn:
Yeah, I agree, we should do Python 3 only

VikingScientist:
I think a sensible solution would be to pretend that there are no python versions. At least project this to the outside audience. We write "python" and never "python 3". This means that it will stand the test of time, in that it makes sense after 2020 when py2 is discontinued. All python files should start with `from __future__ import print_function` which means that almost everything should work for both versions. Anyone which have either version installed will have no problems running the examples and it is very non-intrusive.

In the off-chance that the single `import` statement does not suffice in making the code python2/python3 compatible (no such codes as of yet), then we take the stance that we only support the python3 version, but there is still no need to cumber any reader with the distinction as we continue to just write "python" when we talk about it.

Butt4cak3:
I thought about this one and I now strongly support the decision to just drop Python 2 altogether and go with Python 3 *without* added compatibility for Python 2. Here is why.

## Get with the times
Python 2 support will be dropped soon-ish. There is no reason to keep it around unless you're dealing with legacy code. Beginners should learn Python 3 right away.

## Python 2 and 3 are similar enough
[...]

PR #145, opened 2018-06-28
General python support

VikingScientist:
In awaiting the conclusion from #141, I tried to show how a general version-agnostic approach could look. In practice, you only need to add the line `from __future__ import print_function` to all python files and they will in 99% of the times work for both versions. At least when considering small algorithms such as this project here and not huge complex interconnected libraries that have lots of dependencies.


jiegillet:
I think the consensus is that we don't want to have a compromise on versions. Getting rid of `py2` and `py3` and use `py` is great, but I don't think we want to include `from __future__ import print_function ` at all. People who want to use python 3 code using python 2 should add that themselves.

chapters/euclidean_algorithm/euclidean.md
```diff
@@ -89,7 +89,7 @@ Program.cs
 [import, lang="javascript"](code/javascript/euclidean_example.js)
 {% sample lang="py2" %}
```
jiegillet: You missed a py2 here

chapters/tree_traversal/tree_traversal.md
```diff
@@ -41,7 +39,7 @@ Because of this, the most straightforward way to traverse the tree might be recu
 {% sample lang="js" %}
 [import:15-23, lang:"javascript"](code/javascript/Tree_example.js)
 {% sample lang="py2" %}
```
jiegillet: py 2 here

chapters/tree_traversal/tree_traversal.md
```diff
@@ -41,7 +39,7 @@ Because of this, the most straightforward way to traverse the tree might be recu
...
-[import:8-16, lang:"python"](code/python2/Tree_example.py)
+[import:7-15, lang:"python"](code/python/Tree_example.py)
 {% sample lang="py3" %}
```
jiegillet: py3 here (although there is no code)

leios:
As a quick note the `py` label should be defined in the `book.json` file in the root directory. this will make sure that all `py` labels will have the correct tag at the top of the screen ("Python" instead of "Python 2" or "Python 3"). I can do this and submit a PR to your branch @VikingScientist , or you can change 

```
                                {
                                        "lang": "py2",
                                        "name": "Python 2"
                                },
                                {
                                        "lang": "py3",
                                        "name": "Python 3"
                                },

```

to

```
                                {
                                        "lang": "py",
                                        "name": "Python"
                                },

```

and then just check to make sure the language is correct and syntax highlighting is right by running `gitbook serve`.

jiegillet:
@leios is there a place we could mention explicitly our stance on python version?

leios:
@jiegillet I think I we already did in #141. Maybe we should make language-specific style guides and mention it in here?

jiegillet:
No I mean in the book. The topic is bound to come up. It should be stated somewhere, maybe?

leios:
Yeah, that's what I meant too. We should add some style guidelines to the *How to Contribute* section and for python, we just say that we are sticking to python 3 for this.

I am not sure where else to put it, but @julianschacher was mentioning he wanted to create a *How to Contribute* guide... So that would be a good place too. 

Maybe we should move this to a separate issue?

jiegillet:
No #141 is the right place to discuss that. This is slightly wrong though, so let's get back to code reviewing, sorry @VikingScientist

VikingScientist:
Hehe. This is fine. I initiated the discussion, and if it is discussed here or in #141 really doesn't bother me. I agree with @leios though: we should put the "python 3 only"-stance somewhere (i.e. in the How to contribute section) so that it is available for new contributers. The issue tracker will fade; in fact auto-close after this PR is merged, so it's not a good long-term documentation.

There's a few upcoming commits more where I make the requested changes to `book.json` as well as fixing the last python-code in the FFT. After this, all version-specific code is gone, so I'll remove py2 and py3 from `book.json`.

jiegillet:
Coolz. Ping me when you feel it's ready for a merge.

leios:
Oh yeah, if this PR removes all version specific stuff from the whole book, it should be fine to remove `py2` and `py3`. I just thought there were more python implementations hiding somewhere.

VikingScientist:
I did `grep` for any mention of py2 or py3, but can't find any trace of it anymore.

I like fixup commits such as 2356130896210f6f3ebe69c6d8a38407f4902c31 as it makes the reviewers life easier, but I want these to be squashed with it's referencing commit before being merged into trunk. So if I get a thumbs up on this so far then I'll run `git rebase -i HEAD~6 --autosquash` and then it's ready to be merged.

Butt4cak3:
We always squash and merge PRs (through the GH interface) anyway. There's no need to manually squash. But thanks for the consideration!

VikingScientist:
Rebased to resolve some merge conflicts. Ready to be looked at @jiegillet

leios:
Great!, so once this is merged, we can have a look at the python 3 implementations #146 #144 #138 , all of which should have modifications in their markdown files. 

I am mentioning this now so we remember on those later.

PR #146, opened 2018-06-28
Add Python3 code to Verlet

HugoGranstrom:
There was already Python2 code included in the article but as newcomers to Python should learn Python3 it makes sense to also add Python3 code. Hope my code makes sense. 
In the last example I left Stormer-Verlet out because it is just Classical Verlet but you calculate the velocity but you don't use it in this example with the falling ball.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,3 +5,4 @@ Gathros
 Jeremie Gillet (- Jie -)
 Salim Khatib
 Hitesh C
+Hugo Granström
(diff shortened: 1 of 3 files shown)


HugoGranstrom:
Alright, thank you for the review :-). I'm away from home for a couple of days and I don't have my laptop with me so I can fix this on Wednesday. Is that ok?

leios:
Yeah, get to it when you have time! Thanks for contributing!

HugoGranstrom:
@leios thank you sir :-) Hopefully I won't be leaving AAA for a long time ;-)

HugoGranstrom:
I just noticed that the old python2 version of verlet was changed yesterday to python3. I guess I should closed this one then?

Butt4cak3:
Oh no, you're right. I'm so sorry. I totally didn't think of that. I'm sure you will find another cool algorithm you can implement! The timing was just a bit unfortunate because we just made a big decision regarding Python. I hope that's okay.

PR #148, opened 2018-06-28
Python3 monte carlo

RonPeleg:
clean code for monte carlo algo

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -5,3 +5,4 @@ Gathros
 Jeremie Gillet (- Jie -)
 Salim Khatib
 Hitesh C
+name
--- a/chapters/monte_carlo/code/Python3/monte_carlo.py
+++ b/chapters/monte_carlo/code/Python3/monte_carlo.py
@@ -1,0 +1,34 @@
+import random
+import math
+
+
+def in_circle(x, y, r):
+    """
+    return True if (x,y) is inside a circle with radius r
+    :param x: int or float, x position of a point
+    :param y: int or float, y position of a point
+    :param r: int or float, radius of circle
+    :return: True or false
+    """
+    return x ** 2 + y ** 2 < r ** 2
+
+
+def monte_carlo(n, r):
+    """
+    calculate pi
+    :param n: int, number of points
+    :param r: radius of circle
+    :return: estimate of pi,
+    """
+    points_in_circle = 0
+    for _dummy in range(n):
+        x = random.random()
+        y = random.random()
+        if in_circle(x, y, r):
+            points_in_circle += 1
+
+    pi_estimate = 4.0 * points_in_circle / (n * r ** 2)
+    pi_erro = 100*(math.pi - pi_estimate)/math.pi
+    return pi_estimate, pi_erro
+
+print(monte_carlo(10000000, 0.5))


Butt4cak3:
I'm afraid you're a bit late. People are already working on Monte Carlo in Python over in #144. If you want, you can join the conversation and suggest some changes. Sorry for that!

jiegillet:
Hi @RonPeleg, the person that sent the PR that came right before you is unresponsive, so we are switching to you. Are you up for it? I only have a few points for you to change.

CONTRIBUTORS.md
```diff
@@ -5,3 +5,4 @@ Gathros
 Jeremie Gillet (- Jie -)
 Salim Khatib
 Hitesh C
+name
```
jiegillet: It's probably better if you put in your actual name :)

chapters/monte_carlo/code/Python3/monte_carlo.py
```diff
@@ -0,0 +1,34 @@
...
+    return True if (x,y) is inside a circle with radius r
+    :param x: int or float, x position of a point
+    :param y: int or float, y position of a point
+    :param r: int or float, radius of circle
+    :return: True or false
+    """
+    return x ** 2 + y ** 2 < r ** 2
+
+
+def monte_carlo(n, r):
```
jiegillet: We have moved away from using `r` as a free parameter in `monte_carlo`, so you can remove it here, and call `in_circle(x, y, 1)` instead of `in_circle(x, y, r)`. Another, clean alternative is to change `def in_circle(x, y, r):` to `def in_circle(x, y, r = 1):` and then just use it as `in_circle(x, y)`.

chapters/monte_carlo/code/Python3/monte_carlo.py
```diff
@@ -0,0 +1,34 @@
...
+    :return: estimate of pi,
+    """
+    points_in_circle = 0
+    for _dummy in range(n):
+        x = random.random()
+        y = random.random()
+        if in_circle(x, y, r):
+            points_in_circle += 1
+
+    pi_estimate = 4.0 * points_in_circle / (n * r ** 2)
```
jiegillet: Then in this line you don't need `r ** 2` anymore.

chapters/monte_carlo/code/Python3/monte_carlo.py
```diff
@@ -0,0 +1,34 @@
...
+    pi_erro = 100*(math.pi - pi_estimate)/math.pi
+    return pi_estimate, pi_erro
+
+print(monte_carlo(10000000, 0.5))
```
jiegillet: The last thing is, it's best if you have a `main()` function that prints something like "The estimated value of pi is..." and "the Error is...". The error should be calculated in `main()` as well because it's not part of the algorithm.

jiegillet:
One more thing: you need to include your code into `chapters/monte_carlo/monte_carlo.md`. You can mimic how the other languages do it.

Butt4cak3:
Actually, you know what? I have this suspicion that @RonPeleg is already gone. So let's give them another day (it has been 5 at this time) and then move on if we get no answer.

jiegillet:
OK. We have other PRs for the same algorithm.

Butt4cak3:
I'd honestly just take the branch of the inactive PR that everyone was working on and create a new PR based on that.

PR #149, opened 2018-06-28
Implementation of Tree Traversal in java

xam4lor:


Gustorn:
The same argument as in the C# case. I don't think we need separate implementations just to include in the chapters. The final code, especially with the explicit `this` makes it incredibly clear what the reader should expect. In Java you would normally write the parameterless version, so just stick to that one.

Additionally, I'd prefer to keep merging stuff into the archive that actually compiles.

Butt4cak3:
What @Gustorn already said. Show the entire code at the bottom of the chapter and import parts of it for the smaller snippets. No need to have two versions of the same code.

xam4lor:
Allright !

xam4lor:
The problem is that if I do that, I have to options :
- Create useless methods in the Tree class who are only use on the website to explain part by part of the function (like DFS_recursive_postorder())
- Don't add this methods and let a 'hole' which says 'This has not been implemented in your chosen language, so here is the Julia code'

Butt4cak3:
I'd just go for unused methods. They don't hurt any people can take the code and play around with them if they want.

julianschacher:
Currently the C# code is split into `MdAdditional` (for `in-text` code) and normal `example` code as well. Both compile, but one is the actual example.
With the most recent PR (since things and how to do it changed a bit) they now get merged into one file.
So I support and strongly recommend one implementation instead of two and of course ones that actually compile, after all people should be able to run the code and be able use it for learning.

chapters/tree_traversal/code/java/Tree.java
```diff
@@ -0,0 +1,122 @@
...
+    // root node of the Tree
+    public Node root;
+
+    public Tree(int numberRow, int numberChild) {
+        root = createTree(numberRow, numberChild);
+    }
+
+
+
+	 public void DFSRecursive(Node node) {
```
Gustorn: You should really just remove these versions. They are not any more clear and they take up a bunch of space for no reason. People who'll want to use the Java implementation _will_ know what `this.root` means in the other implementations.

xam4lor: But we need this method to have a parameter of type Node.
I have two solutions : one is to create a function with a same name but no parameters like DFSRecursive() who just call DFSRecursive(this.root).
Or I call DFSRecursive(null) and check if the node is null, the node equals this.root.
In java we cannot create functions within other functions.

Gustorn: Why do you think it needs a parameter? We are not copying the Julia functions 1:1, we change them to fit the current language. You can just omit/change parameters

Gustorn: You could create a private function under the public one which does the recursive part and then you can include both in the code example

Butt4cak3: I was about to suggest method overloading as well. Is there a reason not to?

Gustorn:
One more thing, the methods do not follow Java's naming convention: they should be camelCase, e.g.: `dfsStack`.

chapters/tree_traversal/code/java/Tree.java
xam4lor: It was one of my solutions, I think It's the better thing to do.

Gustorn:
I'm a bit unsure about the excessive whitespace use to separate the different functions. I see why you did it but maybe change it to 2 newlines.

chapters/tree_traversal/code/java/Tree.java
```diff
@@ -0,0 +1,129 @@
...
+                    queue.add(c);
+                }
+            }
+        }
+    }
+
+
+    private void createAllChildren(Node node, int rowCount, int childrenCount) {
+        if(rowCount <= 1)
+        return;
```
Gustorn: This indentation is not quite right

chapters/tree_traversal/code/java/Tree.java
```diff
@@ -0,0 +1,129 @@
...
+                stack.push(c);
+            }
+        }
+    }
+
+    public void bfsQueue() {
+        Queue<Node> queue = new PriorityQueue<Node>();
+        queue.add(this.root);
+
+        Node temp;
```
Gustorn: You can move this inside the loop

chapters/tree_traversal/code/java/Tree.java
```diff
@@ -0,0 +1,129 @@
...
+        if(node.children.size() > 2) {
+            System.err.println("Not a binary tree at dfsRecursiveInOrderBinary()!");
+            return;
+        }
+
+        if(node.children.size() > 1) {
+            dfsRecursiveInOrderBinary(node.children.get(0));
+            System.out.println(node.id);
+            dfsRecursiveInOrderBinary(node.children.get(1));
+        }
```
Gustorn: The standard Java coding style would place the else on the same line as the closing brace:
```
if (...) {

} else {

}
```

chapters/tree_traversal/code/java/Tree.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+
+    public void dfsRecursiveInOrderBinary() {
+        dfsRecursiveInOrderBinary(this.root);
+    }
+
+    // This assumes only 2 children
+    private void dfsRecursiveInOrderBinary(Node node) {
+        if(node.children.size() > 2) {
```
Gustorn: I think the de-facto standard is to have spaces around brackets in Java, so

```
if (condition) {

}
// Instead of
if(condition) {

}
```

Similarly for while, for, etc. loops

xam4lor: You are right ! That's a bad practice that I have :/

chapters/tree_traversal/code/java/Tree.java
Gustorn: Any opinion on this one?

chapters/tree_traversal/code/java/Tree.java
Gustorn: Any opinion on this one?

xam4lor: What do you mean by 'inside a loop' ?

Gustorn: You only use `tmp` inside the following while loop, so you could write this instead:

```java
while(stack.size() != 0) {
    System.out.println(stack.peek().id);
    Node tmp = stack.pop();

    for (Node c : tmp.children) {
        stack.push(c);
    }
}
```

Gustorn:
With the newest whitespace/formatting changes the line ranges in the chapter are out of date

xam4lor:
And I think that's done !

xam4lor:
You do what you can ! No problem ! :)

PR #154, opened 2018-06-28
Add bubble_sort.py to bubble_sort.md

Mukundan314:
bubble_sort.py was already written but not linked so I linked it.


Butt4cak3:
Hello @Mukundan314! Unfortunately, this change is already present in #145. You also used the wrong file name (*bubble.py* instead of *bubblesort.py*).

We're still thankful for your contribution and hope that you will stick around to submit some more stuff!

PR #157, opened 2018-06-28
Add Go language to bubble sort

chin123:
Adds a Go implementation for bubble sort. If you have Go installed, you can do `go run bubbleSort.go`  
Edit: Also I'm not sure whether I'm supposed to add myself to the Contributors list yet, so I haven't.


chapters/sorting_searching/bubble/code/go/bubbleSort.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+		if !swapped {
+			break
+		}
+	}
+}
+
+func main() {
+	array := [10]int{1, 45, 756, 4569, 56, 3, 8, 5, -10, -4}
+	fmt.Println("Unsorted array:")
```
Gustorn: Why not just `	fmt.Println("Unsorted array: ", array)`?

chapters/sorting_searching/bubble/code/go/bubbleSort.go
```diff
@@ -0,0 +1,38 @@
...
+	for i := 0; i < len(array); i++ {
+		fmt.Print(array[i], " ")
+	}
+	fmt.Println()
+
+	bubbleSort(array[:])
+
+	fmt.Println("Sorted array:")
```
Gustorn: Similarly, why not just: 	fmt.Println("Sorted array: ", array)

chin123: Yeah, its simpler this way, I changed it. Initially it was like that because I wanted it to be like the output format of the C bubble sort.

PR #158, opened 2018-06-29
euler.m

vasilisk075:
Matlab implementation of the Forward_Euler method

### DIFF
--- a/chapters/differential_equations/euler/code/euler.m
+++ b/chapters/differential_equations/euler/code/euler.m
@@ -1,0 +1,62 @@
+clc;clear;close all
+
+% Author : vasilisk075
+
+%==========================================================================
+% Define Function
+f =@(x) -3*x;
+
+% Define Initial and Final time
+tInit=0;
+tLast=10;
+
+% Define number of points
+N=1e2;
+
+% Set Initial Conditions
+y(1)=1;
+t(1)=tInit;
+
+% Set tolerance
+tol=1e-6;
+%==========================================================================
+
+dt=(tLast-tInit)/(N-1); % Calculate dt
+i=1;                    % Initialize iter
+succes=0;               % Initialize succes variable
+
+% Loop over time
+while succes==0
+    
+    t(i+1) = t(i) + dt;             % Calculate next time
+    y(i+1) = y(i) + f( y(i) )*dt;   % Update solution
+    
+    % Calculate differce between 2 time steps
+    error=abs(( y(i+1) - y(i) )/dt); 
+    
+    % If difference is small enough or max time is reached stop
+    if (error<tol) || (t(i+1)>=tLast)
+        succes=1;
+        break
+    end
+    
+    % Update counter
+    i=i+1;
+end
+
+% Plot numerical solution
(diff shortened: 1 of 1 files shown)


chapters/differential_equations/euler/code/euler.m
```diff
@@ -0,0 +1,62 @@
...
+%==========================================================================
+
+dt=(tLast-tInit)/(N-1); % Calculate dt
+i=1;                    % Initialize iter
+succes=0;               % Initialize succes variable
+
+% Loop over time
+while succes==0
+    
+    t(i+1) = t(i) + dt;             % Calculate next time
```
jiegillet: Dynamically growing vector sizes is bad practice, they should always be predefined. For `t` is is very easy to do since you already know the bounds and the time difference.

chapters/differential_equations/euler/code/euler.m
```diff
@@ -0,0 +1,62 @@
...
+    y(i+1) = y(i) + f( y(i) )*dt;   % Update solution
+    
+    % Calculate differce between 2 time steps
+    error=abs(( y(i+1) - y(i) )/dt);
```
jiegillet: I wouldn't call this `error`, it's not an error, it's a slope. I also wouldn't use that as an indication that the computation is over. I would simply predefine the time vector, calculate the function for all time steps and plot that.

jiegillet:
Also if you want your code to be included in the website, you should include the code in the chapter `,md` file. Look at [other PRs](https://github.com/algorithm-archivists/algorithm-archive/pull/147/files) to see how to do that. Also we'll need to include matlab in the supported languages, `book.json` and `.editorconfig` files and all that. Maybe @Butt4cak3 can do that?

Butt4cak3:
I'll do the EditorConfig once this is merged.

chapters/differential_equations/euler/code/euler.m
vasilisk075: I understand your idea but ODEs usually reach a SteadyState, and this particular surely does, so why waste time computing something that practically stays the same?

jiegillet: What is your function was actually a sinusoidal function? It would stop at the first extremum.  In general you don't have prior knowledge of the function, and computational time is dirt cheap nowadays.

Butt4cak3:
#166 adds Matlab to *book.json*. Let's just merge that one first and we don't have to deal with it here.

chapters/differential_equations/euler/code/euler.m
vasilisk075: It is dirt cheap, if you talk about small problems like that! Generally when i run a big simulation based on some differential equation, i need a creteria indicating that Steady State is reached and thus terminating the simulation. Normaly a don't know the amount of time steps needed for that.
But either way, i think that it is a long discussion, so i am going to update the code according to your suggestions.
Thank you!

jiegillet: Well, yeah, real world is more complex of course. Let's keep it simple here :)

PR #159, opened 2018-06-29
simplifying julia code for monte carlo

leios:
Hey guys, a lot of people were mentioning that the julia code for monte carlo was unclear. After looking at it again, I agree. I really wanted to have some parameter `radius` in the function, just because I wanted to make it clear that the circle could be an arbitrary size. That said, @jiegillet brought it to my attention, that we clearly say we are using a unit circle in the code, and if we read in a radius of greater than the box length, the code is wrong.

For this purpose, I removed the `radius` parameter from the julia code (except in the `in_circle()` function, but I can remove it there too). I was just being overly cautious to try to make the text easier to read and understand and in the process actually made it more difficult for new contributors.

Let me know what you think. I should probably just remove the `radius` from the `in_circle()` function too, now that I think about it.

We need to put code in that makes sense, so whatever you guys think is best should be what we go with here.

### DIFF
--- a/chapters/monte_carlo/code/julia/monte_carlo.jl
+++ b/chapters/monte_carlo/code/julia/monte_carlo.jl
@@ -1,5 +1,8 @@
 # function to determine whether an x, y point is in the unit circle
-function in_circle(x_pos::Float64, y_pos::Float64, radius::Float64)
+function in_circle(x_pos::Float64, y_pos::Float64)
+
+    # Setting radius to 1 for unit circle
+    radius = 1
     if (x_pos^2 + y_pos^2 < radius^2)
         return true
     else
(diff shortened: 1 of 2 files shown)


jiegillet:
That's an improvement. Leaving `radius` in `in_circle` wouldn't bother me at all, but either is ok.

chapters/monte_carlo/code/julia/monte_carlo.jl
```diff
@@ -1,5 +1,8 @@
 # function to determine whether an x, y point is in the unit circle
-function in_circle(x_pos::Float64, y_pos::Float64, radius::Float64)
+function in_circle(x_pos::Float64, y_pos::Float64)
+
+    # Setting radius to 1 for unit circle
+    radius = 1
     if (x_pos^2 + y_pos^2 < radius^2)
```
Gustorn: This should be `return x_pos^2 + y_pos^2 < radius^2` in my opinion. I don't like unnecessary if/else blocks

Gustorn:
The overeager @leios merge strikes again :smiley:

PR #160, opened 2018-06-29
Add euclidean algorithm implementation for Go

chin123:
Run with `go run euclidean.go`


chin123:
@Gustorn Thanks, fixed now!

PR #162, opened 2018-06-29
Add D implementation of monte carlo

scarface-one:
Its a fairly straight forward implementation. Simply run with `rdmd monte_carlo.d`

### DIFF
--- a/chapters/monte_carlo/code/d/monte_carlo.d
+++ b/chapters/monte_carlo/code/d/monte_carlo.d
@@ -1,21 +1,21 @@
 ///Returns true if a point (x, y) is in the circle with radius r
-bool inCircle(double x, double y, double r = 1.0)
+bool inCircle(real x, real y)
 {
-    return x ^^ 2 + y ^^ 2 < r ^^ 2;
+    return x ^^ 2 + y ^^ 2 < 1.0;
 }
 
 ///Calculate pi using monte carlo
 real monteCarloPI(ulong n)
 {
     import std.algorithm : count;
-    import std.random : Random, uniform01, unpredictableSeed;
+    import std.random : uniform01;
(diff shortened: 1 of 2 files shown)


chapters/monte_carlo/code/d/monte_carlo.d
```diff
@@ -0,0 +1,29 @@
...
+///Calculate pi using monte carlo
+real monteCarloPI(ulong n)
+{
+    import std.algorithm : count;
+    import std.random : Random, uniform01, unpredictableSeed;
+    import std.range : generate, take;
+    import std.typecons : tuple;
+
+    auto rnd = Random(unpredictableSeed);
+    return generate(() => tuple!("x", "y")(rnd.uniform01, rnd.uniform01))
```
Gustorn: I think we should make this part a little more explicit, something like:

```d
    auto radius = 1.0;
    auto piCount = generate(() => tuple!("x", "y")(rnd.uniform01, rnd.uniform01))
        .take(n)
        .count!(a => inCircle(a.x, a.y, radius));
    return 4.0 * piCount / (n * radius * radius);
```

I definitely think radius should be a variable just or clarity's sake and with that change, jamming the return value into one expression seems a bit over the top.

scarface-one: I can do the return bit, but its worth noting that the Julia code is being refactored to remove the radius variable entirely. I think that would be better.

Gustorn: Removing `radius` also works but I'd still introduce the separate variable for `piCount`.

scarface-one: done

PR #165, opened 2018-06-29
add Go implementation of monte carlo

chin123:
It assumes that we are taking a unit circle, so it takes a radius of 1.0

### DIFF
--- a/chapters/monte_carlo/code/go/monteCarlo.go
+++ b/chapters/monte_carlo/code/go/monteCarlo.go
@@ -1,0 +1,42 @@
+// Submitted by Chinmaya Mahesh (chin123)
+
+package main
+
+import (
+	"fmt"
+	"math"
+	"math/rand"
+	"time"
+)
+
+func inCircle(x, y float64) bool {
+	if x*x+y*y < 1.0 { // the radius of an unit circle is 1.0
+		return true
+	} else {
+		return false
+	}
+}
+
+func monteCarlo(samples int) {
+	count := 0
+	s := rand.NewSource(time.Now().UnixNano())
+	r := rand.New(s)
+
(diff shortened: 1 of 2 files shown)


chapters/monte_carlo/code/go/monteCarlo.go
```diff
@@ -0,0 +1,42 @@
...
+
+	for i := 0; i < samples; i++ {
+		x, y := r.Float64(), r.Float64()
+
+		if inCircle(x, y) {
+			count += 1
+		}
+	}
+
+	var estimate float64
```
Gustorn: Why this instead of `estimate := 4.0 * float64(count) / float64(samples)`?

chapters/monte_carlo/code/go/monteCarlo.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+	"math"
+	"math/rand"
+	"time"
+)
+
+func inCircle(x, y float64) bool {
+	if x*x+y*y < 1.0 { // the radius of an unit circle is 1.0
```
Gustorn: This is a pet peeve of mine, you can just do: `return x*x+y*y < 1.0`

chapters/monte_carlo/code/go/monteCarlo.go
chin123: that was an oversight because the code was slightly different before :) Fixed now

chapters/monte_carlo/code/go/monteCarlo.go
chin123: :man_facepalming: don't know why i did that, thanks for mentioning!  Fixed now.

PR #168, opened 2018-06-29
Removed unnessesary temp variable

aquacash5:
This is a very minor change, but it is a little more pythonic


aquacash5:
This should do the trick for the documentation.

Gustorn:
Thanks for the contribution :)

PR #170, opened 2018-06-29
Fix the size of the diagrams. Add res folders.

julianschacher:
Also add res folder for building_blocks and move images into it.
This is targeting this card https://github.com/algorithm-archivists/algorithm-archive/projects/11#card-10967676


julianschacher:
Also: What do you (reviewer) think about the size of the images.
If everything is fine, let us wait for @leios s opinion, since this is a `Chapter Edit`.

julianschacher:
Ok, a few question came up for me, while I'm working on this.

First of all: should the image [here](https://www.algorithm-archive.org/chapters/principles_of_code/notation/notation.html) be centered as well? I did it in a commit, but I can easily revert it.

Also: I added a few `res` folders and also changed the syntax to include an image from
```markdown
![Image name](path)
```
to
```html
<p align="center">
    <img src="path" />
</p>
```
since the latter one is also used, when we need to specify a certain size and having all images included the same way seemed nice.
A few chapters still need to be done, so what do you (all) think? Should we use the first version whenever possible and just use the second, if needed? Or use the second version every time and include everything in a standardized way?
(Especially @leios is asked)

chapters/differential_equations/euler/euler.md
```diff
@@ -51,7 +51,9 @@ Now, solving this set of equations in this way is known as the *forward* Euler M
 In fact, there is another method known as the [*backward* Euler Method](backward_euler.md), which we will get to soon enough.
 For now, it is important to note that the error of these methods depend on the timestep chosen.
 
-<div style="text-align:center"><img src ="res/error.png" /></div>
+<p align="center">
```
Gustorn: You should not use the "align" attribute. We already have a custom CSS file to fix dome syntax highlighting, you should add a class there.

julianschacher: Can you do that? I'm not really familiar with CSS.

Gustorn: Sure, I'll send you a PR soon™ (but seriously, it might take a few days, anyone else who wants to do it, feel free)

leios:
Hey @julianschacher, I think the centered (HTML) version is best, but I am also not sure what file to update about the custom CSS and stuff. If you guys think there is a better way for different screen resolutions, please let me know. Right now, I have been manually specifying the size of each image, which is not necessarily the best.

chapters/differential_equations/euler/euler.md
julianschacher: I added something to the file, that seems to provide the desired result.

julianschacher:
Using the `width` attribute is more mobile friendly than using the `height` attribute.

styles/website.css
```diff
@@ -9,3 +9,9 @@
 .book .book-body .page-wrapper {
   margin-bottom: 12em;
 }
+
+img {
```
Butt4cak3: This is probably not what you want. This changes the style of *every* `` element, including, for example, the CC banner at the bottom of the front page.

PR #180, opened 2018-06-30
Adding Bubble Sort in Swift

CDsigma:
Adding Bubble Sort in Swift


jiegillet:
I realize we didn't ask you to do that on your previous PR, but could you add some sort of main() function which runs an example so that your contrition can be run on its own?

CDsigma:
@jiegillet Just added in a main function so that bubble sort runs on its own. I think this is what you were asking for but I'm not entirely sure. I'd be happy to make changes if I misunderstood 👍

PR #181, opened 2018-06-30
Added java implementation of Huffman compression

Sklan:


chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+    public TreeNode(int value, TreeNode left, TreeNode right) {
+        this.value = value;
+        this.left = left;
+        this.right = right;
+    }
+
+}
+
+class HuffmanTree {
+    private Map<String, Integer> map = new HashMap<>();
```
Gustorn: I think this could use a more descriptive name

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+    private Map<String, String> codeBook = new HashMap<>(), reverseCodeBook = new HashMap<>();
+    private TreeNode root;
+    private String s;
```
Gustorn: Similarly, single-letter class members are usually not a good idea

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+
+    public HuffmanTree(String s) {
+        this.s = s;
+    }
+
+    public void createTree() {
+        for (int i = 0; i < s.length(); i++) {
+            String key = Character.toString(s.charAt(i));
+
+            if (!map.containsKey(key)) map.put(key, 1);
```
Gustorn: If the else branch has braces, so should the if

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+        Comparator<TreeNode> nodeComparator = Comparator.comparingInt(o -> o.value);
+
+        Queue<TreeNode> priorityQueue = new PriorityQueue<>(nodeComparator);
+        for (Map.Entry<String, Integer> m : map.entrySet()) {
+            priorityQueue.add(new TreeNode(m.getKey(), m.getValue()));
+        }
+
+
+        while (priorityQueue.size() > 1) {
+            TreeNode temp1 = priorityQueue.remove();
```
Gustorn: `left` and `right` are probably better names here

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+            TreeNode temp2 = priorityQueue.remove();
+            TreeNode node;
```
Gustorn: At least merge this declaration with the line below but you could also just do`priorityQueue.add(new TreeNode(...));`

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+
+}
+
+class HuffmanTree {
+    private Map<String, Integer> map = new HashMap<>();
+    private Map<String, String> codeBook = new HashMap<>(), reverseCodeBook = new HashMap<>();
+    private TreeNode root;
+    private String s;
+
+    public HuffmanTree(String s) {
```
Gustorn: You should probay use better variable names for the constructor too

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+            TreeNode temp2 = priorityQueue.remove();
+            TreeNode node;
+            node = new TreeNode(temp1.value + temp2.value, temp1, temp2);
+            priorityQueue.add(node);
+        }
+        root = priorityQueue.remove();
+
+    }
+
+    private void traverse(TreeNode temp, String w) {
```
Gustorn: Once more, these argument names are not the best.

Gustorn: Also, a StringBuilder would be much better here

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+        if (temp.left == null && temp.right == null)
+            codeBook.put(temp.key, w);
+        if (temp.left != null) traverse(temp.left, w + 0);
```
Gustorn: I don't think you can add integers to strings. Make sure to always test if your program compiles and works before submitting a PR

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private void CodeBookReverse() {
+        for (Map.Entry<String, String> m : codeBook.entrySet()) {
+            reverseCodeBook.put(m.getValue(), m.getKey());
+        }
+    }
+
+    public String encode() {
+        traverse(root, "");
+        String enc = "";
```
Gustorn: A StringBuilder would be more efficoentt here

chapters/data_compression/huffman/code/java/huffman.java
```diff
@@ -0,0 +1,125 @@
...
+        for (int i = 0; i < s.length(); i++) {
+            String k = Character.toString(s.charAt(i));
+            enc += codeBook.get(k);
+        }
+        printCodeBook();
+        return enc;
+    }
+
+    public String decode(String enc) {
+        String dec = "", key = "";
```
Gustorn: Again, a StringBuilder and variable name suggestion. Since this is an educational resource I would go with encoded/decoded instead of enc/dec

chapters/data_compression/huffman/code/java/huffman.java
Sklan: The program is compiling fine. You can add integers to string in Java. I'll use StringBuilder as you say.

chapters/data_compression/huffman/code/java/huffman.java
Sklan: Ok, I'll change the variable names and make them more descriptive.

chapters/data_compression/huffman/code/java/huffman.java
Gustorn: Sorry, you're right about the addition. Weird, not many languages allow that and Java's usually pretty strict (I've never seen it used in practice either)

Sklan: Yeah, I was shocked too when I first learned about it. But you are right it creates unnecessary confusion.

PR #182, opened 2018-06-30
Add rust implementation of forward euler

hsjoihs:
Naive translation of the C implementation. Correct me if there's anything wrong, since it's my first time contributing to this project.


chapters/differential_equations/euler/code/rust/euler.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut result = [0.0;100];
+    let threshold = 0.01;
+    let timestep = 0.01;
+
+    solve_euler(timestep, &mut result, 100);
+    println!("{}", check_result(&result, 100, threshold, timestep));
+}
+
+
+fn solve_euler(timestep: f64, result: &mut [f64], n: usize) {
```
Gustorn: You don't need to pass `n`, Rust slices already know their size. Also, I would pass a `&[f64]` and return an `Option<Vec<f64>>` which would be much more idiomatic.

hsjoihs: Oops! Thanks

chapters/differential_equations/euler/code/rust/euler.rs
```diff
@@ -0,0 +1,34 @@
+fn main() {
+    let mut result = [0.0;100];
```
Gustorn: You should consider formatting your code with `rustfmt` so it conforms to the standard styleguide.

PR #188, opened 2018-06-30
Added MonteCarlo in Java

DominikRafacz:


jiegillet:
Hi @DominikRafacz, welcome to the AAA and thanks for your contribution.
My Java skills are buried too deep to be useful, but other implementations (including the original one in Julia) have moved away from using the `radius` parameter (you can check other Monte Carlo PR's discussions for that). 
You can drop the radius parameter from `monteCarlo ` and always use the value 1 instead. You can leave it in `inCircle` if you want.

chapters/monte_carlo/code/java/MonteCarlo.java
```diff
@@ -0,0 +1,34 @@
...
+        return x*x + y*y < 1;
+    }
+
+    //function to calculate estimation of pi
+    public static void monteCarlo(int samples){
+        int piCount = 0;
+
+        Random random = new Random();
+
+        for(int i = 0; i < samples; i++){
```
Gustorn: Standard Java style has parens around `for` and `if` statements: `for (...) {` instead of `for(...){`

Issue #57, opened 2018-02-19
Unintuitive Directory Scheme


leios:
Currently, the Algorithm Archive is divided into large algorithmic "domains" of `computational_physics`, `computational_mathematics` and so on. I wrote it this way so that readers could easily learn about that topic of interest when it was sufficiently filled with algorithms. For example, if someone wanted to learn about computational physics, they just go to that section. 

The problem here is that many of the algorithms in `computational_physics` could easily belong in `computational_mathematics` or other areas, so instead of providing large overarching directories, it might be better to divide those sections up into individual components (like `ODE_solvers` or something), and then provide an "algorithmic guide to computational physics" at the start, which provides commentary through all the different algorithms and methods in computational physics (kinda like a large outline for those who want to learn the topic).

These guides would be written after we have a sufficient number of algorithms in the Archive, so it will be a while, but I wanted to create an issue to keep this in mind and see if anyone has a better idea on how to organize this project. As always, I am open to suggestions!

jiegillet:
Out there suggestion: apply some unsupervised machine learning techniques to the wikipedia pages of all the algorithms to group them into 4 categories, and figure out how to call them from that :)

leios:
Huh. I kinda like that idea. It would be neat to use one of the methods we eventually cover here to organize the algorithm archive. 

This will take some time to do, though. So for now, I think we should do something simple. 

It doesn't necessarily need to be `n = 4` categories either. We can use affinity propagation (or something) to split it up more organically.

jiegillet:
Agreed, that's a fun project for later :)
Quick research on how people are doing it elsewhere: [here](https://github.com/nryoung/algorithms/tree/master/algorithms), [here](https://github.com/iiitv/algos) or [here](https://github.com/TheAlgorithms/Java)

leios:
I actually left it open because it's still not quite right. I am missing the guides and am not happy with certain sections (like convolutions) just lying about in the chapters directory. This should probably go into a mathematical methods section or something.

I just feel like it's still a work in progress.

julianschacherpp:
Ok

Butt4cak3:
Can this be closed now that #242 is merged?

PR #189, opened 2018-06-30
Reorganize the Archive

nikhiljha:
Reorganized the summary page based on [Wikipedia's list of algorithms](https://en.wikipedia.org/wiki/List_of_algorithms).

Also working on reorganizing the content itself, so WIP.


leios:
I agree on re-organizing the summary page and such; however, once we decide on which scheme we are working with, the urls will change and I will need to update all my links on youtube (and other places) to match the new archive. In addition, we'll lose our search presence because of it. 

I tried to get this conversation going in #57, because it's incredibly important... but there wasn't much discussion then. Ultimately, whatever scheme we go with, we need to stick to.

The idea of the Algorithm Archive was to provide a cohesive curriculum for all areas of computational science. We can still do this, regardless of the directory scheme by providing "guides" for (for example) computational physics, which provides links for all the chapters in the area and secondary commentary.

nikhiljha:
>  the urls will change 

Luckily, I don't think so. All I moved around was the `SUMMARY.md` which as far as I know doesn't affect the URLs. Not 100% sure though, someone should test this.

leios:
Oh, you are right. As long as we don't change the directory scheme, we are fine; however, I would like to have an introductory blurb to each area. I suppose we could have a different `SUMMARY.md` to the directory scheme. I feel like if we change the directory scheme, all the PR's will have merge conflicts.

Butt4cak3:
You added a bunch of commits that are unrelated to the PR. Please try to work with branches and keep PRs focused on one topic. Everything else makes them difficult to review and merge.

jiegillet:
I'm not in favor of this PR. It is quite rushed, and I would leave jobs like reorganizing the AAA to @leios. Start with contributing by submitting or reviewing code and getting to know other members, not by tearing the house down.

leios:
To be fair, I do think the chapters need re-ordering and I think something like this makes sense. I just don't know the best scheme to go with for now.

Butt4cak3:
I must admit I'm with @jiegillet on this one. In addition to that, we should *first* discuss the matter, come up with a solution and *then* start to work on it, not the other way around.

julianschacher:
Even if we decide to make this change, I strongly suggest to update the directory structure as well and add redirects for the old links. But before we put in all this work, let's discuss this topic.

leios:
So we have some pseudo organization (for example *sorting and searching* and *physics solvers*), which might conflict with other organizational schemes. No matter what organizational scheme we go with (maybe even allowing all algorithms to be in the main directory, while only touching the `SUMMARY.md` file for re-organization), we'll have to modify *some* directories.

If all the algorithms are just sitting in the main directory, then the urls will never change no matter how we change the `SUMMARY.md` file. I don't think that's a bad solution.

julianschacher:
I actually like the idea of a very flat directory structure. It would make the future-proof regarding potential reorganizations. A directory structure complying to the website structure is obviously nice as well.

nikhiljha:
Yup, that's what we're working with right now. It's a flat directory structure and a modified summary.md.

Our current pseudo-organization also fits with the Wikipedia organization, so that works.

Are there any other ways we could organize things?

leios:
I meant that we need to remove that organization too and create a separate `introductions/` directory or something and have an markdown introduction file for each heading.

If we are going for a flat structure, we need to move all the algorithms to the base directory, in my opinion.

julianschacher:
Yeah, so a flat directory structure could consist of all algorithms having an own directory just below base. Then having all the introduction markdowns in an `introductions` directory seems good as well.
The only special case is the `principles_of_code` thingy tho. Maybe we make everything in there flat as well? Just having `priciples_of_code` with all `md`s in it.

leios:
To be honest, the `principles_of_code` section might need a rework as well. The way I see it, we can have that as an exception for now

julianschacher:
I mean making it flat wouldn't be much of an issue.

julianschacher:
I would suggest making the reorganization after https://github.com/algorithm-archivists/algorithm-archive/pull/170 tho, since I add a few `res` folders there.

I should start finishing that...

julianschacher:
Ok, this PR implements flat hierarchy and therefore paves the way for a reorganization.

nikhiljha:
Yeah this PR is pointless now. That change can be addressed in a different PR.

PR #193, opened 2018-07-01
Added Gaussian Elimination in Haskell

jiegillet:
* Added implementation
* Corrected "Eschelon" to "Echelon" (which comes from a French word meaning step of a ladder)
* Minor typos
* Removed "Reduced Row Eschelon Form is diagonal" since it's [not true in general](https://en.wikipedia.org/wiki/Row_echelon_form#Reduced_row_echelon_form)
* Removed "You can do basically anything you want" since if you start from the upper left you'll have to do several passes. While technically you can always do what you want, maybe it's best to suggest the most efficient route.
* James refers to his Julia code as "pseudocode", I didn't want to change the text so my crazy haskell code is now "pseudocode" as well (it works for crazy FP people maybe)


jiegillet:
===> Changes pseudo-code to code
Sorry for the mess, I've had git issues. I'm still modifying a file that has nothing to do with Gaussian Elimination, but it's just removing an empty line and I have no clue how to remove that so please turn a blind eye ^^

chapters/matrix_methods/gaussian_elimination/code/haskell/gaussianElimination.hs
```diff
@@ -0,0 +1,60 @@
...
+  where
+  ((r1, c1), (rn, cn)) = bounds mat
+  echelonPath (r, c)
+    | r > rn || c>=cn = []
+    | mat!(r, c) == 0 = echelonPath (r, c + 1)
+    | otherwise       = (r, c) : echelonPath (r + 1, c + 1)
+  go (r, c) m = foldr (combRows (r, c) 1) (multRow r (1/(m!(r, c))) m) [r1..r-1]
+
+
+printM :: (Show a) => Matrix a -> String
```
lulucca12: [ghc-mod](https://github.com/DanielG/ghc-mod) suggested that this could be defaulted to Double to add a constraint:
```haskell
printM :: Matrix Double -> String
printM m =
  let ((r1, c1), (rn, cn)) = bounds m
   in unlines
        [ intercalate "\t" [take 7 $ show $ m ! (r, c) | c <- [c1 .. cn]]
        | r <- [r1 .. rn]
        ]
```

jiegillet:
* Cosmetics changes to my code using `hindent` (I tried `brittany` but hated it)
* Changes to the chapter to include singular matrices to the algorithm

@lulucca12 If you have more input, shoot me
@leios please review the chapter edits, take your time. We can try some pair programming later to modify your Julia code, if you're up for it! Other implementations will follow.

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
```diff
@@ -111,10 +105,91 @@ $$
...
+Out of all of these, only the first two one make sense if you're talking about a system of linear equations, as the last two don't even have the right dimensions. Additionally, if you translate the last row of second matrix into a system, you get $$0=1$$, which is a contradiction. This is due to the fact that the matrix is singular, and there are no solutions to this particular system. Nevertheless, all of these are in Row Echelon Form.
 
-For now, I hope the motivation is clear: we want to convert a matrix into Row Eschelon and (potentially) Reduced Row Eschelon Form to make large systems of equations trivial to solve, so we need some method to do that. What is that method called? \(Hint: It's the title of this section\)
+Now, it seems obvious to point out that if we ignore the last column, Row Echelon Form is an upper triangular matrix. This might not be important now, but it will play an important role in future discussions, so keep it buzzing in the back of your brain.
+
+As we discussed before, Row Echelon Form is not the terminus. Cue the * **Reduced** Row Echelon Form*.  A matrix is in Reduced Row Echelon form if it satisfies the following conditions:
```
leios: "terminus" sounds funky. It's not wrong, though.

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
```diff
@@ -111,10 +105,91 @@ $$
...
+0 & 0 \\
+0 & 0 \\
+0 & 0
+\end{array}
+\right]
+$$
+
+Again, only the first one (the identity matrix looking guy) is desirable in the context of solving a system of equations, but transforming a matrix in this form gives us an immediate and definitive answer at the question: can I solve my system?
+
+Beyond solving a system, reshaping a matrix in this form makes it very easy to deduce other properties of the matrix, such as the rank.
```
leios: Could you add a definition for "rank"?

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
```diff
@@ -170,25 +245,28 @@ $$
...
-2 & 3  & 4 & 6 \\
+\mathbf{2} & 3  & 4 & 6 \\
 1 & 2 & 3 & 4 \\
 \mathbf{3} & -4 & 0 & 10
 \end{array}
 \right]
 $$
-2. Swap the row with the highest valued element with the `col`th row.
+
+If that value is $$0$$, the matrix is singular and the system has no solutions. Feel free to exit here, but I'm powering through by moving on to the next column, baby!
```
leios: Haha, this is wonderful!

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
```diff
@@ -170,25 +245,28 @@ $$
...
+
+2. Swap the row with the highest valued element with the current row.
```
leios: For some reason, this is showing up as a `1` when rendered in markdown... fixed in PR to your fork.

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
jiegillet: Yeah, it's too frenchy, I'll fix it ^^

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
```diff
@@ -111,10 +105,91 @@ $$
...
+\begin{array}{cc}
+0 & 0 \\
+0 & 0 \\
+0 & 0
+\end{array}
+\right]
+$$
 
-And again has a special name * **Reduced** Row Eschelon Form*. Now, it seems obvious to point out that if we remove the values to the right of the equals sign \($$=$$\), Row Eschelon Form is an upper triangular matrix, while Reduced Row Eschelon Form is diagonal. This might not be important now, but it will play an important role in future discussions, so keep it buzzing in the back of your brain.
+Out of all of these, only the first two one make sense if you're talking about a system of linear equations, as the last two don't even have the right dimensions. Additionally, if you translate the last row of second matrix into a system, you get $$0=1$$, which is a contradiction. This is due to the fact that the matrix is singular, and there are no solutions to this particular system. Nevertheless, all of these are in Row Echelon Form.
```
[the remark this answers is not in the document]
jiegillet: Yes, that's a better way to put it.

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
jiegillet: Added.

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
jiegillet: Sorry :D

chapters/matrix_methods/gaussian_elimination/gaussian_elimination.md
jiegillet: Thank you.

jiegillet:
More changes to the text. @leios, do you want me to split the PR so that the code and chapter edits can be reviewed separately? I guess it would improve my Git-Fu

leios:
In principle, the text and the code should be separate PR's; however, I think it would be a bit messy to split the two here.

I will see what I need to update with the julia code soon.

jiegillet:
Updated my code, it's cleaner now, also added back substitution. Got rid of chapter edits.

lulucca12:
I think before merging this PR(now that #219 is merged) we could format and factor out the code, passing it through a linter (and maybe a beautifier) to verify that all the added code it's all optimized and good.

jiegillet:
I ran it through hlint and got no hints. I tried a beautifier (hindent I think) but it made it worse :)
Do you have suggestions for factoring out? Each function is pretty short already.

jiegillet:
Single letter parameters are mostly matrix indices, `r` for row and `c` for column. Changing the name will make the code less readable I think. I also use `m` for matrix, I suppose I could change that. 

Do you have concrete suggestions for making the code clean and declarative? Honestly I've done my best here.

I'll look into the Show instances, I actually started doing it that way, but it got complicated I don't remember exactly why, maybe because I need a `newtype` rather than `type`? I'll look into it.

lulucca12:
I think there is just a little issue on how this code is written, and from the fact that you created type synonyms to make the code more understandable: There is no way of someone with just the type declaration of any function rewrite that function with given context. To see what I mean with that take an example: the function compose
 
```haskell 
compose :: (b -> c) -> (a -> b) -> a -> c
```
it's really clear(even if you don't give an explicit name) that you can use techniques for programming with context (like using Category Theory, or in a much simpler way Hole-driven Haskell) and generate the function 

```haskell
compose f g x = f (g x)
```
and I understand that this is not always possible with complex algorithms like this one, but making polymorphic types and explicit notations can make this way of Haskell programming better.

lulucca12:
And for the Show instances, could you write your approach in code for me to understand it?

jiegillet:
I know it's not very readable, haskell is not super suited for matrix operations (my FFT code is horrendous). I'm very happy to try and make it better, do you have concrete suggestions?
I tried making Matrix an instance of Show, but that add verbosity everywhere, not worth it.

lulucca12:
Have you seen Data.Matrix? I think we could rewrite the code using its functions and make it more clear. For example, they have a ``` prettyMatrix``` function for displaying matrices. It could make it more clear

lulucca12:
You could use [Data.Vector](https://hackage.haskell.org/package/vector-0.11.0.0/docs/Data-Vector.html#t:Vector) with it

jiegillet:
Data.Matrix is not in the standard distribution, I don't think I should use that, even though it has some functions that I could use (switchRows, combineRows). Even if I did use these functions, the core algorithm (gaussJordan, backSubstitution) wouldn't change at all. All really, if you look at the source course for switchRows for example, it's not much more beautiful than my code. 

Data.Vector is in the standard distribution, I think, but it's based on Data.Array to begin with, and it doesn't have many function I could use.

All in all, none of my functions are longer than 11 lines. They may be dense and a bit hard to read, but it's the reality of the algorithm.

lulucca12:
Being on the standard distribution or not is not a problem if you say clearly this information in the code and I really think that you could shorten your code without defining again what was already defined and can be standard. And it's good to use Haskell libraries because it gives your code more readability. Do you prefer to use "\f g -> (\x -> f(g x))" or "."?

lulucca12:
Have you seen the functions 
```haskell
inverse :: (Fractional a, Eq a) => Matrix a -> Either String (Matrix a) 
```
and
```haskell
rref :: (Fractional a, Eq a) => Matrix a -> Either String (Matrix a) 
```
the last one converts a matrix to reduced row echelon form using Gaussian elimination and thus solving a system of linear equations.
If you don't want to use the built-in algorithm you can implement it using it's [source](https://hackage.haskell.org/package/matrix-0.3.6.1/docs/src/Data-Matrix.html#rref) (I think it's worse in readability, though). You can take a look at all of them and notice that you can shorten your code by (at least) 3 functions implemented (switchRows, prettyMatrix and yours printV that's not necessary because of Vector in Data.Vector being an instance of Show)

jiegillet:
Of course I can't use a built-in algorithm, that defeats the purpose. Overall I'm pretty reluctant to rely too much on outside source, even if it makes my code shorter. I like the idea of having everything you need in one file. And so far, that's been the philosophy of the AAA for all languages.

leios:
I would like to say that the point of this chapter is to show how the algorithm could be implemented in haskell, using an in-built `rref` or `inverse` function kinda defeats the purpose.

I don't know how many people use the `Data.Matrix`, but if it's standard enough (like numpy or scipy), then I suppose we can use it. I don't mind using external libraries (like fftw) if they are common enough and necessary to perform an action.

That said, using a vector/matrix library it might be overkill for this case.

lulucca12:
The purpose was just to shorten the code, I mentioned this functions for you to look at their source and how they are implemented and to get an idea, but for factoring the code I think you can use the library just for using their print and prettier functions and other that can help, like the switchRows and switchCols that you don't have to implement by yourself.

jiegillet:
Yeah I understand your purpose. I also love short, point-free style functions. It's the beauty of Haskell. But I can't refactor a 10 lines function down to 2 even with other people's build in functions. I've already said that I prefer using my own code, especially that my functions are not especially worse than the ones in Data.Matrix.

You haven't criticized a single specific line of code, just the general look of it. You gave me no reason to believe  that you actually read my code at all past the signatures. If it all comes down to esthetics and personal preference, I ask that you respect my choices as the author of the code. If you read the code and have no specific improvements in mind, provide a positive review. If you haven't, leave it to someone else to do the review.

PR #195, opened 2018-07-01
Adding Gaussian Elimination and Thomas Algorithm in C.

Gathros:


chapters/matrix_methods/thomas/code/c/thomas.c
```diff
@@ -0,0 +1,43 @@
...
+    y[0] = c[0] / b[0];
+    x[0] = x[0] / b[0];
+
+    for (size_t i = 1; i < size; ++i) {
+        double scale = 1.0 / (b[i] - a[i] * y[i - 1]);
+        y[i] = c[i] * scale;
+        x[i] = (x[i] - a[i] * x[i - 1]) * scale;
+    }
+
+    for (int i = size - 2; i > -1; --i) {
```
VikingScientist: My favorite way of writing reversed for loops in C is

`for(int i=size-2; i-->0; ) {`

Becuase it looks like it reads "i goes to zero". It is functional identical to your implementation though.

Gustorn: I _strongly_ disagree. The whole "goes to thing" is cute, but it's just obfuscating the fact that it's two separate operators. I like the concept but I'd much rather see something like:

```c
for (int i = size - 2; i >= 0; --i) {

}
```

Since we're not dealing with unsigned numbers this'll just work and it's a lot more clear.

PR #198, opened 2018-07-02
Remove unused "differential_equations.md" and "math.md".

julianschacher:


julianschacher:
This files don't seem to be used or serve any purpose.

leios:
I use `math.md` regularly. In order to get mathjax to work, I need that set of commands somewhere on the chapter, so I kept the set of commands in a central directory to copy from. There might be a way to directly import `math.md` to other chapters instead of just copying it's contents.

If we delete `differential_equations.md`, I'll just end up adding it back later when we need it, but we can delete it for now, if you want to clean up the directory structure.

julianschacher:
Ok, the `math.md` is there again. The PR is just now a minor organizational change.
You can either delete or merge it; what you like better.

PR #200, opened 2018-07-02
Added python gaussian elimination

VikingScientist:
With no use of numpys `array` datastructure. This might be a contested decision.

### DIFF
--- a/chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
+++ b/chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
@@ -1,0 +1,71 @@
+import numpy as np
+
+def gaussian_elimination(A):
+
+    rows = len(A)
+    cols = len(A[0])
+    n    = min(rows, cols)
+
+    # Main loop going through all columns
+    for k in range(n):
+
+        # Step 1: finding the maximum element for each column
+        max_index = np.argmax(np.abs([A[i][k] for i in range(k,n)])) + k
+
+        # Check to make sure matrix is good!
+        if (A[max_index][k] == 0):
+            print("matrix is singular! End!")
+            return
+
+        # Step 2: swap row with highest value for that column to the top
+        A[max_index], A[k] = A[k], A[max_index]
+
+        # Loop for all remaining rows
+        for i in range(k+1,rows):
+
+            # Step 3: finding fraction
+            fraction = A[i][k] / A[k][k]
+
+            # loop through all columns for that row
+            for j in range(k+1,cols):
+
+                 # Step 4: re-evaluate each element
+                 A[i][j] -= A[k][j]*fraction
+
+            # Step 5: Set lower elements to 0
+            A[i][k] = 0
+
+
+def back_substitution(A):
+
+    rows = len(A)
+    cols = len(A[0])
+
+    # Creating the solution Vector
+    soln = [0]*rows
+
+    # initialize the final element
+    soln[-1] = A[-1][-1] / A[-1][-2]
+
+    for i  in range(rows - 2,-1,-1):
+        sum = 0
+        for j in range(rows-1,-1,-1):
+            sum += soln[j]*A[i][j]
+        soln[i] = (A[i][-1] - sum) / A[i][i]
+
+    return soln
+
+def main():
+    A = [[ 2,  3, 4,  6],
+         [ 1,  2, 3,  4],
+         [ 3, -4, 0, 10]]
+
+    gaussian_elimination(A)
+    print(A)
+    soln = back_substitution(A)
+
+    for element in soln:
+        print(element)
+
+if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Main loop going through all columns
+    for k in range(n):
+
+        # Step 1: finding the maximum element for each column
+        max_index = np.argmax(np.abs([A[i][k] for i in range(k,n)])) + k
+
+        # Check to make sure matrix is good!
+        if (A[max_index][k] == 0):
+            print("matrix is singular! End!")
```
jiegillet: Singular matrices have nothing to do with Gaussian elimination. If your pivot is zero, you continue with the next column. It's still a valid matrix in the echelon form.

chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
```diff
@@ -0,0 +1,71 @@
...
+            return
+
+        # Step 2: swap row with highest value for that column to the top
+        A[max_index], A[k] = A[k], A[max_index]
+
+        # Loop for all remaining rows
+        for i in range(k+1,rows):
+
+            # Step 3: finding fraction
+            fraction = A[i][k] / A[k][k]
```
jiegillet: The pivot is not always in the diagonal, in the case that a previous pivot was zero. You have to consider this case.

chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
```diff
@@ -0,0 +1,71 @@
...
+def back_substitution(A):
+
+    rows = len(A)
+    cols = len(A[0])
+
+    # Creating the solution Vector
+    soln = [0]*rows
+
+    # initialize the final element
+    soln[-1] = A[-1][-1] / A[-1][-2]
```
jiegillet: Again here, this is not general, in case the Echelon form is not a pure upper triangular matrix.

jiegillet:
Oh boy. All the mistakes I found in your code are also in @leios's Julia code. This might take a while :)

VikingScientist:
Yup. This is a direct port of the Julia code. Might create an issue to fix that? BTW, the exact same errors appear in #195

jiegillet:
Yeah I'll open an issue. Let's keep this PR in the fridge for a bit if you don't mind.

VikingScientist:
I sniped you and just opened an Issue

chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
leios: https://stackoverflow.com/questions/23985004/gaussian-elimination-of-a-singular-matrix

In my understanding a singular matrix was indicative of an improper set of systems of equations and should be underdetermined. This algorithm shouldn't work for this case (I don't think, anyway). Most example code I can find stops once the matrix is considered singular; however, I think you are right. We can also take the general case. I am just not sure the best version of the code here.

chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
leios: This is actually related to the above discussion, I think. The code I have in the algorithm archive is meant to solve a system of equations, so these cases will be caught by the "singular matrix" condition.

chapters/matrix_methods/gaussian_elimination/code/python/gaussian_elimination.py
leios: Again, this is caught by the "not singular" condition.

Currently looking at the code and seeing how we should update things.

leios:
@jiegillet is right, the form we have in the AAA is not general; however, it is the most common implementation of the algorithm I could find. The current implementation solves a system of equations, which is what the chapter was set up to do. It can also find the determinant; however, it lacks the capability of finding the ranks and bases of a matrix, which means that it cannot be used for all cases.

I'm currently looking at how to update the code from here.

jiegillet:
@leios I was thinking about it yesterday, and yes, if the problem you are working on is solving a system of equations, then yes, the algorithm is perfectly adequate. However, Echelon and Reduced Echelon Forms have more applications than that, for example finding the rank of a matrix. It seems a little bit wasteful to implement the general case 90% of the way and stop there.

leios:
@jiegillet Yeah, the issue #201 is probably the best place to discuss this.

jiegillet:
Update: the Julia code and chapter were updated and merged. Have a look and change your code accordingly.

jiegillet:
Still alive? I will close this PR if it's stale.

PR #203, opened 2018-07-02
Added forward euler in python3

Sklan:


chapters/differential_equations/euler/code/python/euler.py
```diff
@@ -0,0 +1,27 @@
+import math
+
+
+def forward_euler(time_step, n):
+    result = [0] * n
+    result[0] = 1
+    for i in range(1, n): result[i] = result[i - 1] - 3.0 * result[i - 1] * time_step
```
Butt4cak3: Why `3.0` instead of just `3`?

chapters/differential_equations/euler/code/python/euler.py
```diff
@@ -0,0 +1,27 @@
...
+    approx = True
+    for i in range(len(result)):
+        solution = math.exp(-3. * i * time_step)
+        if abs(result[i] - solution) > threshold:
+            print(result[i], solution)
+            approx = False
+        return approx
+
+
+time_step = 0.01
```
Butt4cak3: Most of the Python code we have in the archive uses a `main()` function. To keep everything consistent, you should put this stuff inside a `def main():` and call `main()` like, for example, [in this file](https://github.com/algorithm-archivists/algorithm-archive/blob/a9ef5ff179c2cba47fb564cc04043da2c46c03df/chapters/data_compression/huffman/code/python/huffman.py#L93-L118).

chapters/differential_equations/euler/code/python/euler.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def check(result, threshold, time_step):
+    approx = True
+    for i in range(len(result)):
+        solution = math.exp(-3. * i * time_step)
+        if abs(result[i] - solution) > threshold:
+            print(result[i], solution)
+            approx = False
+        return approx
```
Butt4cak3: Your `return` statement is inside the for-loop.

Butt4cak3:
That should do it!

PR #204, opened 2018-07-02
Modified euclidian algorithm haskell code to be more clear and readable

lulucca12:
Avoided hiding values inside "where" bindings.

### DIFF
--- a/chapters/euclidean_algorithm/code/haskell/euclidean_example.hs
+++ b/chapters/euclidean_algorithm/code/haskell/euclidean_example.hs
@@ -1,25 +1,22 @@
 -- contributed by Nicole Mazzuca (ubsan)
-
 euclidSub :: Integer -> Integer -> Integer
-euclidSub a b = inner (abs a) (abs b) where
-  inner a b =
-    if a == b then
-      a
-    else if a < b then
-      euclidSub a (b - a)
-    else
-      euclidSub (a - b) b
+euclidSub a b = inner (abs a) (abs b)
+  where
+    inner x y
+      | x == y = x
+      | x < y = euclidSub x (y - x)
+      | otherwise = euclidSub (x - y) y
 
 euclidMod :: Integer -> Integer -> Integer
-euclidMod a b = inner (abs a) (abs b) where
-  inner a 0 = a
-  inner a b = inner b (a `mod` b)
+euclidMod a b = inner (abs a) (abs b)
+  where
+    inner x 0 = x
+    inner x y = inner y (x `mod` y)
 
 main :: IO ()
 main = do
   let chk1 = euclidMod (64 * 67) (64 * 81)
       chk2 = euclidSub (128 * 12) (128 * 77)
-  putStrLn (show chk1)
-  putStrLn (show chk2)
+  print chk1
+  print chk2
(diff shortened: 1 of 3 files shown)


Butt4cak3:
Welcome to the AAA and thanks for your contribution!

In my opinion, your changes to this code are definitely improvements, especially those to `euclidSub`. There are a few people here who know Haskell better than me, though, so I'll leave the actual code review to them.

There is one thing you definitely have to change. When you look at [the chapter](https://www.algorithm-archive.org/chapters/euclidean_algorithm/euclidean.html) you will notice that not the entire file is visible, but only the function the current paragraph is about. Since you changed quite a bit, you will have to adjust the line numbers in the code import. That's not too difficult.

Open euclidean.md in your editor and look for the code imports for Haskell. They look like this:
```
{% sample lang="haskell" %}
[import:3-11, lang="haskell"](code/haskell/euclidean_example.hs)
```
You will have to change that one to
```
{% sample lang="haskell" %}
[import:2-8, lang="haskell"](code/haskell/euclidean_example.hs)
```
because lines 2-8 contain `euclidSub`. I'm sure you will figure the rest out!

lulucca12:
I've solved it now

chapters/euclidean_algorithm/euclidean.md
```diff
@@ -53,7 +53,7 @@ Modern implementations, though, often use the modulus operator (%) like so
 {% sample lang="py" %}
 [import:1-9, lang="python"](code/python/euclidean_example.py)
 {% sample lang="haskell" %}
-[import:13-24, lang="haskell"](code/haskell/euclidean_example.hs)
+[import:10-22, lang="haskell"](code/haskell/euclidean_example.hs)
```
Butt4cak3: This should only import the `euclidMod` function without `main`. That was already wrong before you edited it, but if you look at all the other languages, you'll see that they all only import the one function. In other words, this should be `10-14`.

lulucca12: ok, sorry :smile:

Butt4cak3: It's alright. As I said, it was wrong *before* you changed it, so no worries.

chapters/euclidean_algorithm/code/haskell/euclidean_example.hs
```diff
@@ -1,25 +1,22 @@
...
 
 main :: IO ()
 main = do
   let chk1 = euclidMod (64 * 67) (64 * 81)
       chk2 = euclidSub (128 * 12) (128 * 77)
-  putStrLn (show chk1)
-  putStrLn (show chk2)
+  print chk1
+  print chk2
   return ()
```
jiegillet: print is already `IO ()`, is adding `return ()` a thing?

lulucca12: Since you already added the IO() in the type, you can omit it, I'll fix it

jiegillet: Again, I didn't write this :p (I wrote all other Haskell files though, I think)

PR #163, opened 2018-06-29
Add Merge Sort algo chapter and Python example

Mukundan314:

PR #205, opened 2018-07-02
Creating a python example of quicksort

KerimovEmil:
Creating a python example of quicksort, also based on the previous closed PR: https://github.com/algorithm-archivists/algorithm-archive/pull/138  I also removed two unused lines in another python file.


leios:
Firstly, I really appreciate the contribution! However, as I mentioned in #163, the Algorithm Archive is (first and foremost) a book, which means that the text is an *incredibly important* part of this project. At this point in time, I am the only one who has written the bulk text for this project, and though I don't mind additional authors, additional chapters will be somewhat rigorously reviewed to make sure they follow at least a somewhat consistent style and make complete sense. 

Since you did not write a chapter yet, I assume that you are waiting for the chapter to be written, which is fine. We will probably hold on to this PR until the quicksort chapter is written.  We have 3 options:

1. You keep re-writing the text until we are satisfied
2. I rewrite the text and send a PR to you
3. I rewrite the text, put it on the AAA and then you resubmit your code then

That said, this would be a test case for other "authors" in this project, so we are not sure how the process would work, exactly.

KerimovEmil:
Hi @leios , Thanks for your comment! I was actually using https://github.com/algorithm-archivists/algorithm-archive/pull/163 as a reference before creating this PR. In it I saw that the last comment proposed contributing the code and creating a stub .md file. So that when the chapter would be written, some sample python code would already exist. 
 
With that said, I'm open to any suggestion on how you want to proceed. All three options are fine to me.

On a possibly unrelated note, is there a way to suggest chapters to be made?

Butt4cak3:
The one who suggested to leave a stub article and add the code was me. It was probably not very clear, but it was meant as an exception, because the PR already existed and we didn't want to just throw the code away. It was not meant to be the new process for every chapter in the future.

Butt4cak3:
What's the status of this? Can I close this, @leios? The original author seems to be inactive and the code isn't good according to @Gustorn.

PR #206, opened 2018-07-02
R monte carlo

GuyPozner:
Hey I've added an R implementation of the monte carlo, I didn't use loops because they are not welcomed in R.


chapters/monte_carlo/code/R/monte_carlo.R
```diff
@@ -0,0 +1,30 @@
+
+in_circle <- function(x, y){
+        # Return True if the point is in the circle and False otherwise.
+        radius = 1.0
```
jiegillet: It doesn't make sense to hardcode `radius` to 1 here, you can keep it as a parameter of `in_circle`. Since you set it to one in `monte_carlo` anyway, it will have the same effect.

leios: I think this is mimicking the current AAA julia code.

jiegillet: I know (I don't like it but I understand the reason) , but this code hardcodes `radius` to 1 twice! Might as well do it only once in `monte_carlo` and pass it to `in_circle(x,y,r)`

leios: Oh, I didn't see the second hardcoded radius.

chapters/monte_carlo/monte_carlo.md
```diff
@@ -112,6 +114,8 @@ Feel free to submit your version via pull request, and thanks for reading!
 {%sample lang="go" %}
 ### Go
 [import, lang:"golang"](code/go/monteCarlo.go)
+### R
```
jiegillet: Ah, just noticed you missed `{% sample lang="R" %}` here.

PR #208, opened 2018-07-03
Fix Swift and JavaScript code imports.

julianschacher:


leios:
Ah, man. I was just about to do that. We need to figure out what the appropriate label is for swift. It should be in this file: https://github.com/algorithm-archivists/highlight.js/blob/master/src/languages/swift.js (I think)

julianschacher:
Hmm yeah, I try to solve it... Sorry for your PR!

leios:
As a note: once we get these tags fixed, I was planning on pushing to the github pages repo. No rush, though. I cannot even do that until I get home tonight anyway.

PR #209, opened 2018-07-03
Update Tree Traversal C#.

julianschacher:
Restructure code by removing the `Node` class. Remove `CreateAllChildren` method and make the constructor do it instead.

This is applying the changes suggested @Gustorn in https://github.com/algorithm-archivists/algorithm-archive/pull/137 in discussion with @xam4lor .
The PR also removes the `CreateAllChildren` method and just makes the constructor recursive and do the creation.

### DIFF
--- a/chapters/tree_traversal/code/cs/Tree.cs
+++ b/chapters/tree_traversal/code/cs/Tree.cs
@@ -6,110 +6,107 @@ namespace TreeTraversal
 {
     public class Tree
     {
-        private class Node
-        {
-            public List<Node> Children { get; set; } = new List<Node>();
-            public int Id { get; set; }
-
-            public Node(int id) => this.Id = id;
-        }
-
-        private Node root;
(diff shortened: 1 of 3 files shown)


chapters/tree_traversal/code/cs/Tree.cs
```diff
@@ -6,111 +6,108 @@ namespace TreeTraversal
...
-            stack.Push(root);
-            Node temp;
+            var stack = new Stack<Tree>();
+            stack.Push(this);
+            Tree temp;
 
             while (stack.Count != 0)
             {
                 Console.WriteLine(stack.Peek().Id);
                 temp = stack.Pop();
```
Gustorn: I would just do `Tree temp = stack.Pop()` instead of the separate declaration/initialization

julianschacher: Fixed. :)

PR #212, opened 2018-07-03
Added DFS_stack to Haskell examples

Thurii:
Ran the code this time

### DIFF
--- a/chapters/tree_traversal/code/haskell/TreeTraversal.hs
+++ b/chapters/tree_traversal/code/haskell/TreeTraversal.hs
@@ -14,30 +14,26 @@ dfsInOrder (Node x [l])    = dfsInOrder l ++ [x] -- Single branch assumed to be 
 dfsInOrder (Node x [l, r]) = dfsInOrder l ++ [x] ++ dfsInOrder r
 dfsInOrder _               = error "Not a binary tree"
 
+dfsStack :: Tree a -> [a]
+dfsStack (Node x ts) = x : concatMap dfsStack ts
+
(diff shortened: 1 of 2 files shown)


chapters/tree_traversal/code/haskell/TreeTraversal.hs
```diff
@@ -14,30 +14,26 @@ dfsInOrder (Node x [l])    = dfsInOrder l ++ [x] -- Single branch assumed to be
 dfsInOrder (Node x [l, r]) = dfsInOrder l ++ [x] ++ dfsInOrder r
 dfsInOrder _               = error "Not a binary tree"
 
+dfsStack :: Tree a -> [a]
```
jiegillet: This function is _literally_ the same as `dfs`

jiegillet:
I know you're trying to make the code closer to the other implementations, but there is nothing new in this PR. You duplicated a function and you replaced my example tree that I chose to highlight the features of the different tree reversals by something more basic.

chapters/tree_traversal/code/haskell/TreeTraversal.hs
Thurii: ... I apologize, that was not intentional.

Thurii:
Note to self, actually read the code that you're adding to.
I hope I haven't offended you at all.

Though I do think that the examples should provide identical outputs as much as possible. That makes it better for validity/sanity checks. Your tree is definitely the superior tree due to the unique elements. It's always a pain when you aren't sure that a zero is the correct zero.

PR #215, opened 2018-07-04
Added Forward Euler to Haskell

Thurii:


chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,13 @@
+solve_euler timestep n = take n $ iterate f 1
+    where f x = x - 3 * x * timestep
+
+check_result results threshold timestep =
+    and $ zipWith check' results [exp $ -3 * i * timestep | i <- [0..]]
+    where check' result solution = abs (result - solution) < threshold
+
+main = let timestep = 0.01;
```
lulucca12: It is highly recommended to explicit the main type signature, even when it's just 
```haskell 
main :: IO() 
```

Thurii:
Hindent is pretty cool!
Set the types to `Doubles`, didn't realize how generalized `checkResult` had become.

chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,20 @@
+solveEuler :: Double -> Int -> [Double]
+solveEuler timestep n = take n $ iterate f 1
+    where
+        f x = x - 3 * x * timestep
```
lulucca12: I am not sure whats more readable,
```haskell
solveEuler :: Double -> Int -> [Double]
solveEuler timestep n = take n $ iterate f 1
    where
        f x = x - 3 * x * timestep

```
or

```haskell
solveEuler :: Double -> Int -> [Double]
solveEuler timestep n = take n $ iterate (\x -> x - 3 * x * timestep) 1
```

What do you think?

jiegillet: The second one

jiegillet:
This is just a personal preference, but Haskel being a functional programming, I think we should emphasize the functions. I believe that the function that calculates the acceleration should be a parameter in `solveEuler`. This way, the Euler algorithm is general, and we can choose the Newton equations along with the time steps and other parameters down in `main` .  I submitted code for the [Verlet algorithm](https://www.algorithm-archive.org/chapters/physics_solvers/verlet/verlet.html) (way at the bottom, not in the text, I should fix that) if you want to see what I mean.

lulucca12:
@jiegillet I see what you mean, but I asked about implementing the code as a lambda expression or a 'where' expression. Either way I think it could be simpler implementing the function as a parameter, like you said.

chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,20 @@
+solveEuler :: Double -> Int -> [Double]
```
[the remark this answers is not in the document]
jiegillet: I'm maybe overdoing it, but if you want to be more general, something like
`solveEuler :: (Particle -> Acceleration) -> Time -> Particle -> [Particle]`
would be good, where `Particle` represents a particle in space-time, something like like
`type Particle = (Position, Speed, Acceleration, Time)`
the first parameter is the physical model, the second parameter is the time step, because it shouldn't be part of the physical model, and the third parameter is the initial condition.
I'll admit that this kind of general thinking it *not* represented in any of the other implementations.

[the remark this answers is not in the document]
lulucca12: I think using Doubles for now is good, but how would you implement the timestep given to the solveEuler function? I think it is unnecessary if we implement it in the acceleration function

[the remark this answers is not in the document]
jiegillet: Well, in my Verlet implementation (see link in my other comment), everything is either double or list of doubles (to use in more than 1 dimension). Giving them a name only makes it easy to reason about them.
We can't use the tilmestep in the acceleration function, the acceleration function the f in: dy(t)/dt=f(t,y(t)).  You give f any time, place and speed, it tells you what the acceleration is. Once you know the acceleration, you calculate the new position using the tilmestep and whichever method you like, in this case Euler. So these are used in different places.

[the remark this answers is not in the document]
lulucca12: I think this is what you meant:
```haskell
solveEuler :: (Double -> Double) -> Double -> Int -> [Double]
solveEuler accFun timestep n =
  take n $ iterate (\x -> x + accFun x * timestep) 1

``` 

as now you can pass the timestep and the acceleration function and it calculates the solution. the function (\x -> x + accFun x * timestep) should calculate the change in x for a given time and acceleration generated by the acceleration function.
Is this correct or I misunderstood what you are trying to say with the acceleration function?

chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,20 @@
...
+checkResult results threshold timestep =
+    and $ zipWith check' results [exp $ -3 * i * timestep | i <- [0 ..]]
+    where
+        check' result solution = abs (result - solution) < threshold
+
+main :: IO ()
+main =
+    let timestep = 0.01
+        n = 1
+        threshold = 0.01
```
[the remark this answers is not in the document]
lulucca12: Sould the acceleration function be:
```haskell
accFun x = -3 * x
```

chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,20 @@
...
+    in  putStrLn $
+        if checkResult (solveEuler timestep n) threshold timestep
```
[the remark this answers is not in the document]
lulucca12: So now with the timestep:
```haskell
if checkResult (solveEuler accFun timestep n) threshold timestep
```

chapters/differential_equations/euler/code/haskell/euler.hs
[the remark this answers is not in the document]
Thurii: This is what I have right now:
```haskell
solveEuler :: Num a => (a -> a) -> a -> Int -> [a]
solveEuler func initVal n = take n $ iterate func initVal
```
Just pure forward euler and `func` can be whatever you need.

chapters/differential_equations/euler/code/haskell/euler.hs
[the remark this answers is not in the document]
Thurii: Do you think it's better to leave timestep as an explicit parameter in solveEuler, or have it be a part of accFun?
I'm thinking timestep is more closely related to solveEuler and should be a parameter so that you can try different timesteps without having to modify accFunc.

chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,23 @@
+solveEuler :: Num a => (a -> a) -> a -> Int -> [a]
+solveEuler func initVal n = take n $ iterate func initVal
+
+checkResult :: (Num a, Ord a, Enum a) => [a] -> (a -> a) -> a -> Bool
```
jiegillet: You can simplify this
```
checkResult :: (Ord a, Num a, Num t, Enum t) => a -> (t -> a) -> [a] -> Bool
checkResult thresh check = and . zipWith (\i k -> abs (check i - k) < thresh) [0..]
```

chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,23 @@
+solveEuler :: Num a => (a -> a) -> a -> Int -> [a]
```
jiegillet: I know that the Julia code returns `n` values, but the is Haskell, have `solveEuler` return an infinite list, we can limit ourselves to `n` values down in `main`

Thurii: True. Though now the whole function is feeling kinda off because it's (now more obviously) just `solveEuler = iterate`. At that point why even define it?

jiegillet: It's a good idea to define it anyway because giving it a name gives it meaning.
In a more general setting, it's really supposed to do more than that, it would have another parameter, the time step, but you included it in `kinematics`. It's OK like this, though.

Thurii: After watching Leios's video I think I have a somewhat better idea of what this *should* look like.
The timestep should definitely be a solveEuler parameter.

chapters/differential_equations/euler/code/haskell/euler.hs
```diff
@@ -0,0 +1,20 @@
...
+
+kinematics :: Double -> Double
+kinematics x = -3 * x
+
+main :: IO ()
+main =
+    let timestep = 0.01
+        n = 100
+        threshold = 0.01
+        checkResult' = checkResult threshold $ exp . (\x -> x - 3 * x * timestep)
```
jiegillet: I think here it should be `(\x -> - 3 * x * timestep)`

PR #216, opened 2018-07-04
insertion-sort-chapter

GuyPozner:
Hi, Ive created an insertion sort chapter, hope you'll like it. I'll add more languages soon.

PR #217, opened 2018-07-04
added selection sort python algorithme

SudoChmodAll:


Gathros:
I'm putting this on hold till #216 gets the green light.

SudoChmodAll:
this is not insertion sort sorry, this is selection sort I've corrected it.

Gathros:
I got rid of the on hold label since it's not the same algorithm.

julianschacher:
The algorithm will not be in the AAA soon and in general algorithms should be written with the chapter in mind imho. Also they need to get integrated in the chapter and if there is no chapter, the algorithm can't be integrated. I see no reason, why this PR even exists.
I appreciate the contribution tho! We're trying to build out the AAA as fast as possible while holding up with our quality standards.

chapters/sorting_searching/selection/code/python/selection.py
```diff
@@ -0,0 +1,7 @@
+def selectionList(lst):
+	for i in range(0, len(lst)):
+		if i < len(lst):
```
jiegillet: I don't know what this `if` statement is doing. Could you please explain?

oussamabouchareb: I really can't remember what was the purpose of that, and it seems wrong from looking at it now, I'm ashamed of this tbh, I code better than this now I promise

Issue #201, opened 2018-07-02
Gauss elimination code (Julia) is wrong 


VikingScientist:
Basically
* Don't check for singularity as Gauss Elimination can be done on singular matrices
* Pivots are not neccessarily on the diagonal
* Cover the general case and don't assume triangular matrices

This issue affects PR #200 (contains more details on the bugs) and PR #195

jiegillet:
#183 (which is already merged) is also wrong

jiegillet:
The best is 4. In general, almost nothing you state in the chapter is wrong so it doesn't need that much rewrite. Back substitution DOES make sense in the case of singular matrices, you just need to travel through the pivots, not the diagonal. Again, very little rewrite, I believe. And yes, putting a couple more examples of non-physical examples will help with intuition.

I'm very happy to try and add something to my PR tonight, I've already modified the chapter anyway.

leios:
Yup, sounds good. Please feel free to do what you can when you have the time.

jiegillet:
I've modified the chapter over in #193

PR #219, opened 2018-07-05
updating gaussian elimination to be more general

leios:
Hey guys, following the discussion in #193 and #201 , we looked at making the Gaussian Elimination code more general. This change does not change the computational complexity of the gaussian_elimination function, but it does for the back-substitution. 

In the previous julia version, we were literally plugging the values in for the upper diagonal matrix. In this version, we are actually doing the full Gauss-Jordan elimination and creating a reduced row eschelon matrix. The definition of "back-substitution" can mean either case.

As we mentioned before, the previous code we had submitted was not "wrong"; however, this is technically more general. I am curious to see what you guys think about the back-substitution step. In the worst-case, we can add a condition to switch between the methods if the matrix is not singular (which means that it admits a regular set of solutions). As a note: the solution vector of this method is the last column of the output matrix.

Please let me know what you think about this implementation. We can always change it.


jiegillet:
I want to apologize for not fully grasping your original implementation. I had not realized that the complexity was so different. 
I support the idea of keeping both functions. Checking if a matrix is singular after it goes through `gaussian_elimination` is as simple as checking `A[rows, cols-1] == 0` (or by stopping the computation as you were doing before). After that, one could decide to use your old `back_substitution` or, say, `to_reduced_echelon`.

leios:
I'm alright with having multiple implementations that switch off depending on the matrix. It should be easy enough to implement and we can highlight the difference in the text.

jiegillet:
I don't even think you need to implement it, you just need to make both functions available and let people decide which method they want. Someone who only wants to solve a system wouldn't be interested in looking at the reduced echelon form anyway.

leios:
Yeah, I just meant that we re-write the code like you suggested and mention the differences in the text.

jiegillet:
God I really wish I had not put text and code in the same PR. Do you want me to do it or you want to PR my branch?

leios:
I will PR your branch (tomorrow, though)

leios:
Alright, sorry this took so long. I was busy this weekend with other commitments. Here is an attempt to solve the issues and PR's for Gaussian Elimination.

I split the chapter to have both Gauss-Jordan elimination and back-substitution, which seemed to be the fairest thing to do. This allows the programmer to choose which method best suits their use-case. I am sure there will be discussion, so this is a "rough" edit.

jiegillet:
This is a good change. I'll get rid of the chapter edits in my PR. 
A small nitpick is the inconsistent capitalization of Echelon Form/echelon form.

jiegillet:
Sorry, closed it by mistake

leios:
No prob dawg. I'll review for typos and update this PR soon.

chapters/algorithms/gaussian_elimination/code/julia/gaussian_elimination.jl
```diff
@@ -1,43 +1,47 @@
+using DataStructures
```
jiegillet: This is not needed anymore.

chapters/algorithms/gaussian_elimination/gaussian_elimination.md
```diff
@@ -80,20 +80,63 @@ $$
...
-
-and it has a particular name: _Row Eschelon Form_. Basically, any matrix can be considered in row eschelon form if
+This matrix form has a particular name: _Row Echelon Form_.
+Basically, any matrix can be considered in row echelon form if
 
 1. All non-zero rows are above rows of all zeros
 2. The leading coefficient or _pivot_ (the first non-zero element in every row when reading from left to right) is right of the pivot of the row above it.
 
-Now, Row Eschelon Form is nice, but wouldn't it be even better if our system of equations looked simply like this
+All the following examples are in the Row Echelon Form:
```
jiegillet: The capitalization of [R|r]ow [E|e]echelon is all over the place. After some thoughts, I suggest dropping the capital letter everywhere (since row echelon is not a name, it's like saying diagonal), except maybe when it's first introduced. Same deal for RRE.

leios:
Cool, I'll make the changes as soon as I get home tonight!

PR #223, opened 2018-07-06
Python implementation for Jarvis March

wmboyles:
I struggled a bit, so some of the individual commits are unnecessary.


chapters/computational_geometry/gift_wrapping/jarvis_march/code/py/jarvits.py
```diff
@@ -0,0 +1,41 @@
+# Is the turn counter clockwise?
+def CCW(p1, p2, p3):
+    return (p3[1]-p1[1])*(p2[0]-p1[0]) >= (p2[1]-p1[1])*(p3[0]-p1[0])
+
+
+#Find the leftmost point in the list
+def leftmost(gift):
```
jiegillet: You can use a `key=` with `min` to reduce this function to one line: `min(gift, key=lambda point: point[0])`

chapters/computational_geometry/gift_wrapping/jarvis_march/code/py/jarvits.py
```diff
@@ -0,0 +1,41 @@
...
+    leftmost = gift[0]
+    for point in gift[1:]:
+        if point[0]<leftmost[0]: leftmost=point
+    return leftmost
+
+
+#Are two points the same point?
+def equal(p1,p2):
```
jiegillet: Can't you simply use `p1==p2`?

chapters/computational_geometry/gift_wrapping/jarvis_march/code/py/jarvits.py
```diff
@@ -0,0 +1,41 @@
...
+    return p1[0]==p2[0] and p1[1]==p2[1]
+
+
+def jarvisMarch(gift):
+    n = len(gift)  #Number of points in list
+    hull = [(None,None)] * n #The hull, initially empty points
```
jiegillet: I know that initializing lists is a good idea in general, but here it looks clunky. I would just append points to an initially empty hull as they come.

jiegillet:
hi @wmboyles, welcome to the AAA and thanks for your contribution.
On top of my other comments in text, could you add a `main()` function that actually runs on a set of points?

chapters/computational_geometry/gift_wrapping/jarvis_march/code/py/jarvits.py
jiegillet: Well, actually, in this case you can just use `min(gift)` directly, because when comparing a list, python compares by the first element first.

wmboyles:
Thanks for the input @jiegillet . I have implemented everything you asked for. Do you think the hull should include the starting point both at the beginning and end of the hull? I think having it emphasizes the fact that the convex hull is a closed shape, but understand the possible confusion of having the same point in the hull twice.

chapters/computational_geometry/gift_wrapping/jarvis_march/code/py/jarvits.py
```diff
@@ -0,0 +1,42 @@
+# Is the turn counter clockwise?
+def CCW(p1, p2, p3):
+    return (p3[1]-p1[1])*(p2[0]-p1[0]) >= (p2[1]-p1[1])*(p3[0]-p1[0])
+
+
+def jarvisMarch(gift):
+    n = len(gift)  #Number of points in list
+    pointOnHull = min(gift) #leftmost point in gift
+    hull = [pointOnHull] #leftmost point guranteed to be in hull
```
jiegillet: tiny typo `guranteed `

chapters/computational_geometry/gift_wrapping/jarvis_march/code/py/jarvits.py
```diff
@@ -0,0 +1,42 @@
...
+    
+    i = 0
+    while True:
+        endpoint = gift[0]
+        for j in range(1,n):
+            if endpoint==pointOnHull or not CCW(gift[j],hull[i],endpoint):
```
jiegillet: `hull[i]` is the only point where you used `i`, correct? Could you get away with using `hull[-1]`?

jiegillet:
Thanks for the quick answer! Your changes are great.
- I think the starting point should only be there once. We don't need to emphasize that a hull is a closed shape, it's the definition. 
- I left a couple of small comments.
- Could you remove your commit on the Euclidian Algorithm? It gets really messy when a PR has more than one topic. Feel free to open a new PR about that if you want.

wmboyles:
Changes made. I added myself as a contributor in a separate pull request.

jiegillet:
It's almost done, but during your PR, the euclidian algorithm has changed, and here you are still modifying it. That's exactly why you shouldn't mix PRs :)
Get rid of any change to`euclidean_example.py` and I'll merge directly.

wmboyles:
Hopefully that should fix all of my mistakes.

PR #224, opened 2018-07-06
Renamed some language code dirs for consistency

Butt4cak3:
Some languages, namely JavaScript and Haskell had code directories both called "javascript" and "js" or "haskell" and "hs" respectively. That kind of bothered me for no practical reason, so I changed it.

Because the code directories of all other languages are named after the language, not the file extension (e. g. "julia" rather than "jl"), I went for "javascript" and "haskell" over "js" and "hs".

I also renamed "R" to "r" because everything else is lowercase.

One thing I thought about chaning, but didn't, are the C# code directories. They are called "cs", after the C# file extension. I thought about renaming them to "csharp" to go with the theme. What are your thoughts on it?


Gustorn:
I vote for `csharp` as well, it's just more readable.

julianschacher:
csharp sounds good

leios:
Yeah, this is good. Merging.

Should we also note what languages correspond to different directories somewhere?

Butt4cak3:
It's not a bad idea, but I don't know where we could note that. I guess as long as the dev team knows the naming scheme we're fine for now.

PR #227, opened 2018-07-06
Cleaned up the code a bit

SudoChmodAll:
made it use a one print instead of two and deleted the function that was handling the two prints and made it print directly and used a -= b, b -= a instead of a = a - b, b= b - a

### DIFF
--- a/chapters/euclidean_algorithm/code/python/euclidean_example.py
+++ b/chapters/euclidean_algorithm/code/python/euclidean_example.py
@@ -22,4 +22,5 @@ def euclid_sub(a, b):
     return a
 
 if __name__=="__main__":
-    print('Eucledean Mod: {}\nEucledean sub: {}'.format(euclid_mod(64 * 67, 64 * 81), euclid_sub(128 * 12, 128 * 77)))
+    print('Euclidean mod: ', euclid_mod(64 * 67, 64 * 81))
+    print('Euclidean sub: ', euclid_sub(128 * 12, 128 * 77))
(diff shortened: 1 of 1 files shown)


jiegillet:
The `a-=b` is OK, I guess, the but the `main` is less readable with your version. It's also misspelled.

jiegillet:
I am aware of how `main` works in python. For our purposes it doesn't really matter if the code is a script or a module, but I agree that using that using `if__name__==" __ main __ ":` is a general improvement. 
However, my criticism was that you made `main` less readable by changing two short and readable lines into one monster of a line. Also you misspelled "Euclidian".

chapters/euclidean_algorithm/code/python/euclidean_example.py
```diff
@@ -15,14 +15,11 @@ def euclid_sub(a, b):
...
 
     return a
 
-def main():
-    print(euclid_mod(64 * 67, 64 * 81))
-    print(euclid_sub(128 * 12, 128 * 77))
-
-main()
+if __name__=="__main__":
+    print('Eucledean Mod: {}\nEucledean sub: {}'.format(euclid_mod(64 * 67, 64 * 81), euclid_sub(128 * 12, 128 * 77)))
```
jiegillet: 1. It's spelled Euclidian
2. split that line into 2 lines to make it easier to read, something like
```
print('Eucledean Mod: {}'.format(euclid_mod(64 * 67, 64 * 81))
print('Eucledean sub: {}'.format(euclid_sub(128 * 12, 128 * 77))
```

Butt4cak3: Pretty sure it's "Euclidean".

PR #229, opened 2018-07-07
Added myself as a contributor

wmboyles:


lulucca12:
I think you should do the same as in PR #230

PR #230, opened 2018-07-07
Update .md to include Python code

wmboyles:


lulucca12:
Why opening a new pull request

lulucca12:
I think you should add this in the PR #223 and open a new PR for the Euclidian Algorithm

PR #234, opened 2018-07-08
Changes to Euclidean Algorithm in Java

CDsigma:
I think that the Euclidean Algorithm should be implemented and run within a single class rather than two. So I modified the implementation of the Euclidean Algorithm in Java to only be one class. I made the following changes:

- Removed MainClass.java file
- Added main() function to EuclideanAlgo class
- Modified euclidean.md to reflect changes

This may just be a personal preference though so I understand if it isn't merged 👍

### DIFF
--- a/chapters/algorithms/euclidean_algorithm/code/java/EuclideanAlgo.java
+++ b/chapters/algorithms/euclidean_algorithm/code/java/EuclideanAlgo.java
@@ -24,4 +24,9 @@ public class EuclideanAlgo {
 
         return a;
     }
+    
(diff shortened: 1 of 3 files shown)


Butt4cak3:
I agree. These are code examples, not libraries, so it's alright to put everything into one file/class. I'll wait another day or two and then merge it if nobody has a good reason not to.

Butt4cak3:
> another day or two
... or four.

CDsigma:
haha its all good :)

PR #240, opened 2018-07-10
Adding Example Code to Sorting Algorithms

CDsigma:
My current understanding is that code in the AAA is supposed to show a main() function which runs the algorithm. So I updated the java sorting algorithms to reflect this.

- Changed bubble_sort.md so main() is now shown
- Changed bogo_sort.md so main() is now shown

### DIFF
--- a/contents/bogo_sort/bogo_sort.md
+++ b/contents/bogo_sort/bogo_sort.md
@@ -14,19 +14,55 @@ In code, it looks something like this:
 
 {% method %}
 {% sample lang="jl" %}
-[import:1-14, lang:"julia"](code/julia/bogo.jl)
+[import:10-14, lang:"julia"](code/julia/bogo.jl)
 {% sample lang="cs" %}
 [import:9-15, lang:"csharp"](code/csharp/BogoSort.cs)
 {% sample lang="clj" %}
-[import:2-11, lang:"clojure"](code/clojure/bogo.clj)
+[import:7-11, lang:"clojure"](code/clojure/bogo.clj)
 {% sample lang="c" %}
-[import:4-29, lang:"c_cpp"](code/c/bogo_sort.c)
+[import:25-29, lang:"c_cpp"](code/c/bogo_sort.c)
 {% sample lang="java" %}
-[import:2-17, lang:"java"](code/java/bogo.java)
+[import:2-6, lang:"java"](code/java/bogo.java)
 {% sample lang="js" %}
-[import:1-16, lang:"javascript"](code/javascript/bogo.js)
+[import:11-15, lang:"javascript"](code/javascript/bogo.js)
 {% sample lang="py" %}
-[import:4-12, lang:"python"](code/python/bogo.py)
+[import:10-12, lang:"python"](code/python/bogo.py)
+{% sample lang="hs" %}
+[import:17-20, lang:"haskell"](code/haskell/bogoSort.hs)
+{% sample lang="m" %}
+[import:21-28, lang:"matlab"](code/matlab/bogosort.m)
+{% sample lang="cpp" %}
+[import:33-38, lang:"c_cpp"](code/c++/bogosort.cpp)
+{% sample lang="rs" %}
+[import:16-20, lang:"rust"](code/rust/bogosort.rs)
+{% sample lang="swift" %}
+[import:32-39, lang:"swift"](code/swift/bogosort.swift)
+{% endmethod %}
+
+That's it.
+Ship it!
+We are done here!
+
+## Example Code
+
+{% method %}
+{% sample lang="jl" %}
+[import, lang:"julia"](code/julia/bogo.jl)
+{% sample lang="cs" %}
+BogoSort.cs
+[import, lang:"csharp"](code/csharp/BogoSort.cs)
+Program.cs
+[import, lang:"csharp"](code/csharp/Program.cs)
+{% sample lang="clj" %}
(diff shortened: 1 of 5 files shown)


Butt4cak3:
To be honest, I don't think we're even sure when to show the main function and when not to. Some chapters intentionally only show the algorithm in question, others show full code examples with a main function.

We should probably agree on something or come up with a rule. Let me summon @leios, @julianschacher, @Gathros, @jiegillet and @Gustorn real quick. They probably have opinions on this.

julianschacher:
It's true that we don't have a `Code Example` section in some algorithm chapters. Only parts of the actual code are shown there.
For chapters with the `Code Example` section, the whole code should be shown imo and I also handle it like this with my C# imports.

With the `Bubble Sort` and `Bogo Sort` chapters we don't have a `Code Example` section, but we should add one maybe. It would make the chapter more comprehensive.

Butt4cak3:
We have a few chapters in which only the algorithms themselves are imported. Maybe we should *always* add one code block at the very bottom that contains a working, compilable/runnable example.

julianschacher:
@Butt4cak3 yeah, that was what I was thinking of and I support that

CDsigma:
I totally agree with both of you that there should be one section just showing the algorithm and then one example code section which has a working example.

I just added in an example code section for Bubble sort and Bogo sort

julianschacher:
@CDsigma Hey, I found a little problem with the C# imports of the `Example Code` sections and there were also a few other adjustments to be made. Therefore I made a PR, to your `javaSorting`-Branch, which should be safe to merge: https://github.com/CDsigma/algorithm-archive/pull/1

julianschacher:
The restructuring of the chapters seems good now; so from my side everything is ok :)
I just want to let @Butt4cak3 look over it as well and then he can decide to merge the PR.

Butt4cak3:
By the way, why do we even still have the language specific headlines for the full example codes (`### Java`, `### Haskell`, ...)? Also related to #245.

leios:
Ah, I guess you are right. I had them in initially because we didn't have the theme api. There's no harm in removing them now since we can indicate what language you are reading with the theme-api.

CDsigma:
Just removed the language headlines and changed 'Example Code' to using `##`

Butt4cak3:
You also kinda just copied the full code imports so that we now have the same, full code twice for some languages like C++.

I think now that we're already working on it, we should do it right. I think the first code block should show *only* the `bubbleSort` function, no `shuffle` no `isSorted` and no `main` - in *every* language. Then, there should be the `## Example Code` section with all the full examples.

julianschacher:
I agree with @Butt4cak3 there, that makes the most sense for these chapters. For functions like `isSorted()` it's pretty clear what they do even without seeing the implementation. And if people want to see, how it actually works, they can look at the bottom.

CDsigma:
I agree it makes sense to only show the bubble sort function for the first part and then in the 'Example Code' section to show all the code. I adjusted the .mds accordingly.

leios:
Sorry it has taken me so long to get to this PR. I agree. The intent was to have the full code in the `Example Code` section and have snippets in the bulk text. For bubble and bog sort, this idea was not fully realized because it was early in the process and we hadn't committed to the "multi-lang" feature.

chapters/algorithms/bogo_sort/bogo_sort.md
```diff
@@ -14,35 +14,66 @@ In code, it looks something like this:
...
 
 That's it.
 Ship it!
 We are done here!
 
+## Example Code
+
+{% method %}
+{% sample lang="jl" %}
+[import:1-14, lang:"julia"](code/julia/bogo.jl)
```
leios: Why do some of the languages have line numbers to import and others don't? Shouldn't the example code section just import an entirely working code sample?

julianschacher: It should. I guess that needs a change.

Butt4cak3: Yup. That's pretty much what I tried to say in my most recent comment.

CDsigma: Sorry about that, I accidentally messed up on the Julia code for both Bubble Sort and Bogo Sort in the Example Code section. I just made a commit to fix this. 

The rest of the line numbers given are to hide unnecessary comments at the beginning of the programs saying who created them. It seems unnecessary to show these comments since it doesn't provide useful information to the reader. Additionally, there is already the contributors.md file so the programmers are receiving credit elsewhere. I think there are three options for what can be done here:

1. Keep it as is (meaning hide those lines by modifying the .md file)
2. Modify the code to remove the unnecessary comments
3. Show the comments and then sort out what should be done if anything in a different PR

Or, another option I may have missed. 
Give me your thoughts on what you think should be done 👍

CDsigma:
I disagree that we should be showing those comments but I also don't think that this is a key issue for the project.

So I just made a commit to show all comments in the 'Example Code' section. This should help to simplify this pull request.

In the future I may make an issue to help further the discussion around these comments, which show the contributor at the top of the program, since I think they add unnecessary clutter.

julianschacher:
Thank you :) It's the way it's handled until now and for the sake of holding to standards it's good to include it here as well.

Aside from that: I think showing the contributors and giving them credit is a nice gesture, we should keep up.

Butt4cak3:
Good work!

PR #242, opened 2018-07-11
Implement flat hierarchy. Fix some broken links.

julianschacher:
This PR implements a flat hierarchy for all algorithm and algorithm related chapters.
There are discussion points left tho:

- [x] Should the `Principles of Code` chapters also be implemented in flat hierarchy or should we just leave them as is? // Leios will change that at some point, so I just don't do anything in here
- [x] @leios what should I do about the `Runge Kutta Methods` chapter? It links to `Euler Methods`, but this doesn't exists. Should I link to something else or can this stub just get deleted? // Runge Kutta got removed
- [x] Should the `Principles of Code` chapter be a sub-chapter of the new `introduction` chapter? // I won't do this in this PR anymore, also the `Principles of Code` will get reworked by Leios
- [x] What do you guys think in general of the flat hierarchy? I really like it, because algorithms are super easy to find in the directory overview now. // Leios approved the PR


leios:
Nice. This was a lot of work and you even got the in-chapter links.

I will be gutting a good amount of the `Principles of Code` chapter soon. For now, we can leave it and I'll put it where it belongs after the update. 

Feel free to delete the stubs. I'll add the chapters back when I actually start working on them. I was  little too ambitious in certain places.

I am generally in-favour of the flat hierarchy, as long as the summary can be modified correctly.

julianschacher:
Ok, I deleted the `Runge Kutta Chapter` for now, since it's causing some dead links. I just leave the other stubs for now tho, since they shouldn't be part of this PR.

julianschacher:
The summary can be modified without consequences, so the AAA site looks like the old one with the flat hierarchy. The flat hierarchy also allows easier restructuring and is therefore more robust for the future. Let's wait what the others say to the flat hierarchy tho.

julianschacher:
If anyone is worried about broken links, we can easily fix that with redirects. I just found a plugin enabling us to do so.

leios:
It there is no more discussion on this topic, I'll merge it before making my changes to the split-op chapter for this weekend's video.

julianschacher:
Notes (especially for @leios ): 
- All chapters follow a special naming scheme now: chapter names and the names of their included `.md`s are written all lowercase with the following characters: `a-z` + `_`. With these chars the name of the chapter folder and of the `.md` should be the same name as the chapter title.
So the chapter `Forward Euler Method` uses the name `forward_euler_method` for it's `.md` and it's folder.
By doing so, less confusion exists and maximum structure can be achieved.
- All algorithm chapters should be located in `chapters/algorithms/`, all general chapters should be located in `chapters/general/`, all introduction chapters should located in `chapters/introduction/` and the `Principles of Code` part of the AAA also has its own directory called `chapters/principles_of_code`. There shouldn't be more than these 4 `chapters/` sub-directories. Adding a new sub-directory only happens with a big new feature the AAA gets.
- The structure displayed on the sidebar of the AAA site can be changed by modifying the `SUMMARY.md`. There's no need to change the actual directory structure.
- If a chapter gets renamed or deleted and legacy support for the old link is wanted, an entry should be added to the `redirects.json`. Looking at the style of the other entries, a new redirect should be easy to set up.

If you (reading this) encounter any problems or have questions, feel free to ask me here or on [Discord](https://discord.gg/pb976sY).

PR #243, opened 2018-07-12
Improve readability of examples in verlet.cpp

antonte:
- make functions return time
- assign value to vars at declaration
- use next_pos var name instead of tmp_pos (delete comment)


antonte:
BUMP

PR #244, opened 2018-07-12
Cleanup bubble sort code for C++

antonte:
- remove random generator function
- use nested loop instead of is_sorted var

### DIFF
--- a/contents/bubble_sort/bubble_sort.md
+++ b/contents/bubble_sort/bubble_sort.md
@@ -25,7 +25,7 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 {% sample lang="hs" %}
 [import, lang:"haskell"](code/haskell/bubbleSort.hs)
 {% sample lang="cpp" %}
-[import, lang:"c_cpp"](code/c++/bubblesort.cpp)
+[import:13-23, lang:"c_cpp"](code/c++/bubblesort.cpp)
 {% sample lang="rs" %}
 [import:6-19, lang:"rust"](code/rust/bubble_sort.rs)
 {% sample lang="d" %}
--- a/contents/bubble_sort/code/c++/bubblesort.cpp
+++ b/contents/bubble_sort/code/c++/bubblesort.cpp
@@ -2,70 +2,34 @@
 #include <cstddef>
 #include <iostream>
 #include <iterator>
-#include <random>
 
-using std::begin;
-using std::end;
-
-template <class Rng>
-std::vector<float> generate_input(std::size_t size, Rng& rng) {
-  auto dist = std::uniform_real_distribution<>(0.0, 1.0);
-
-  auto ret = std::vector<float>();
-  std::generate_n(std::back_inserter(ret), size,
-    [&rng, &dist] { return dist(rng); });
-
-  return ret;
+template <class Iter>
+void print_range(Iter first, Iter last) {
+  for (auto it = first; it != last; ++it)
+    std::cout << *it << " ";
+  std::cout << std::endl;
 }
 
 template <class Iter>
-void print_range(std::ostream& os, Iter const first, Iter const last) {
-  os << '{';
-
-  if (first != last) {
-    os << *first;
-    std::for_each(first + 1, last, [&os] (double d) { os << ", " << d; });
-  }
-
-  os << "}\n";
-}
-
-template <class Iter>
-void bubble_sort(Iter const first, Iter const last) {
-  if (first == last) {
-    // the empty range is vacuously sorted
-    return;
-  }
-
-  for (;;) {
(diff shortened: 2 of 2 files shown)


Butt4cak3:
I don't have much to do with the C and C++ code in this project, but I know that most other language implementations (C#, Java, JavaScript, Julia, Matlab, Python, Racket?, Rust, Swift) have no optimization whatsoever. The chapter also talks about the pure, unoptimized version of it, so I don't know why it's there in the first place.

> And that's it for the simplest bubble sort method. Now, as you might imagine, computer scientists have optimized this to the fiery lakes of Michigan and back, so we'll come back to this in the future and talk about how to optimize it. For now, it's fine to just bask in the simplicity that is bubble sort.

I think all implementations should follow the same procedure, so I would actually vote to *remove* the optimization from all implementations to make it match both the chapter text and the Julia version, which is sort of serves as the reference implementation for all chapters.

Other than that, there was no reason to remove the random generator function. Random inputs to test the programs are fine.

I also don't understand why you felt the need to rename the .cpp file. That seems to me like a pure preference thing. We don't have a convention for how to name sources files and I don't think we need one. In addition to that, you renamed it so that Git wasn't even able to track it. For future reference, you should try to always move files with `git mv` to avoid that.

I think that's all I have to say.

antonte:
My PR actually closer to Julia implementation, Julia implementation does not have is_sorted var. I removed random generation because implementation is confusing for inexperienced programmers and does not provide to much value. I renamed the file to make the name consistent with c implementation.

Butt4cak3:
It's different, not closer. It doesn't have the `is_sorted` variable, but it has a third nested loop, which the Julia implementation also doesn't have. In addition to that, I'm pretty sure your implementation is even slower than the standard one without any optimization.

The implementation of the random generator is not subject of this chapter. It doesn't really matter as long as it works.

Most of my other points still stand, but I'll let someone who actually knows C and C++ go ahead and properly review this.

antonte:
Third nested loop is just a bug. I pushed updated version.

- fix the bug with netsted loop
- renamed back bubblesort.cpp
- match algorithm with Jilia

chapters/algorithms/bubble_sort/code/c++/bubblesort.cpp
```diff
@@ -2,70 +2,34 @@
...
+void bubble_sort(Iter first, Iter last) {
+  for (auto it1 = first; it1 != last; ++it1) {
+    for (auto it2 = first; it2 + 1 != last; ++it2) {
       // these are unsorted! gotta swap 'em
-      if (*(it + 1) < *it) {
-        using std::swap;
-        swap(*it, *(it + 1));
-        is_sorted = false;
+      if (*(it2 + 1) < *it2) {
+        std::swap(*it2, *(it2 + 1));
```
Gustorn: You could use `std::iter_swap(it2, it2 + 1);`

antonte:
BUMP

leios:
Hey @antonte, sorry for taking so long to get to this PR. For the record BUMPs are not really necessary and lead to spam for everyone following the algorithm archive. Please avoid using them in the future, if you can. We have a lot of PR's and are getting to them when we have time.

As far as this PR is concerned, it is difficult to review because it is basically an update to your personal preference here, which means we need to discuss style. It is hard to do so for both C and C++ at the same time.

In this case, the C++ code is notably cleaner; however, I am not sure about the C code. Would it be possible to split the PR into 2?

antonte:
This PR now is only C++.
C implemenation you can find here: https://github.com/algorithm-archivists/algorithm-archive/pull/254

antonte:
@gastorn can you please re-approve this PR, because I rebased it to fix automerge

contents/bubble_sort/bubble_sort.md
```diff
@@ -14,6 +14,8 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 [import:9-27, lang:"csharp"](code/csharp/BubbleSort.cs)
 {% sample lang="c" %}
 [import:11-21, lang:"c_cpp"](code/c/bubble_sort.c)
+{% sample lang="cpp" %}
```
Butt4cak3: There is already a C++ code import further down. No need to create another one. In fact, this causes the page to render incorrectly.

antonte: Should be good now.

Butt4cak3:
The import propery works now. For the future, please don't amend your commits and force push after you made requested changes. It's helpful for us as reviewers to see only what you changed after the review, or else we have to look through the entire code again.

Since @Gustorn already approved the code itself, I will go ahead and merge this one.

antonte:
@Butt4cak3 Got it, from now I'll make commits without squish and rebase.

PR #245, opened 2018-07-12
Fixing 'Example Code'

CDsigma:
- Made 'Example Code' a consistent size throughout document
- Added in 'Example Code' on several pages
- Added in missing programming language titles

### DIFF
--- a/chapters/computational_geometry/gift_wrapping/graham_scan/graham_scan.md
+++ b/chapters/computational_geometry/gift_wrapping/graham_scan/graham_scan.md
@@ -61,6 +61,7 @@ In the end, the code should look something like this:
 ### C
 [import, lang:"c_cpp"](code/c/graham.c)
 {% sample lang="js" %}
+### Javascript
 [import, lang:"javascript"](code/javascript/graham-scan.js)
 {% endmethod %}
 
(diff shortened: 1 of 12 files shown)


Butt4cak3:
I checked the changes in this PR for correctness and I can say there are no errors in it. I'd still like to pull @leios into this PR to discuss which size headline we want to use for "Example Code". In my opinion, it should be a H2 (`##`) because it should be one level lower than the chapter headline, which is a H1 (`#`) since the example code is part of the chapter and not a chapter in itself.

leios:
Yeah, agreed. If we could use `##` instead of `#`, that would be nice. I use `##` for all other section headings (I think)

Butt4cak3:
I just asked myself: Why do we even have the language headlines (`### Java`, `### Haskell`, ...)? There is always only one programming language visible at a time anyway and we don't add headlines to the smaller code snippets. Honestly, I don't think we need those, but I want to hear what others think.

Butt4cak3:
Okay. The same question has been answered in [another PR](https://github.com/algorithm-archivists/algorithm-archive/pull/240#issuecomment-404424748). I'd say we finish up this one first (change "Example Code" to `##` instead of `#`) and then remove all the language headlines in another PR.

leios:
Yeah. Agreed. I want to merge #242 soon, too. I think it will break some PR's, though.

Butt4cak3:
It shouldn't break too much. Git properly detected all the file moves, so it should be able to handle it.

CDsigma:
@Butt4cak3 I'll make that PR to remove language headlines :)

Issue #33, opened 2018-01-14
Useless chapters.


Gathros:
The AAA has a couple of useless chapters like Mathematical Background and Principle of Code. The Chapter A Personal Note is weird but I think it should be renamed preface and should be less personal.

julianschacherpp:
To sum up, what I already said in Discord regarding that topic and the changes being in discussion (and to present my vision of implementing them):
I find the Mathematical Background chapter pretty useful and would combine it with another chapter, if needed. If we go full only-algorithm we can get rid of it for my part.
The Principle of Code chapter is something I would like to keep in a separate book or guide and link to it in an Introduction / HowToContribute page, to satisfy a more algorithm-focused book.
About the A Personal Note page: I really like it. It adds a personal touch to the project and I wouldn't  discard it.

Gathros:
It's really long. BTW I haven't read it.

julianschacherpp:
@Gathros The "A Personal Note"?

Gathros:
Yes, it's better if it was shorter and renamed to prelude or something like that.

julianschacherpp:
@Gathros The naming doesn't matter that much I think. Who wants to read it, does, who doesn't want to, doesn't. I read it and found it really interesting and lovely.

Gathros:
If the goal of the AAA is to be a book then preludes would make sense since books have preludes.

julianschacherpp:
@Gathros That's true. A personal note fits as well I think. It doesn't really matter what name you give to the prelude.

leios:
@Gathros As it currently stands, I do not think there are many useless chapters, except for *A Personal Note*. The mathematical background and principles of code are obviously not the purpose of this book, but are kinda necessary for newcomers. I kept the material there super short and don't intend to make it any longer than necessary.

Some things, like Taylor Series and Convolutions belong in the mathematical background because they are used in a number of algorithms all over the place.

julianschacherpp:
@leios I would keep _A Personal note_. It adds something to the AAA.

leios:
As a note, the language discussion can probably be omitted; however, I really wanted to do language benchmarks later, which I feel would fit into AAA discussion pretty well. 

The AAA is not meant to be an introductory text, but we'll have a bunch of people who don't know how to code look into it because of how I advertise it on youtube. I think a ~10 page crash course to a bunch of different topics is fine and common in most books, it's just often skipped by the reader.

Gathros:
[...]

PR #250, opened 2018-07-15
removing principles of code chapter, updating book

leios:
This PR addressed #33 . It removes the principle of code section (which was kinda aimless rambling anyway) and transforms it into mathematical background and data structures chapters, both of which will be continually updated with time.

It also moves other chapters to their relevant location.

As a note, this also hides work-in-progress sections from the site. In principle, these should all be hidden away in branches on my local fork, but I'll get to that soon. It also removed the personal note. It was basically a discussion about linux anyway, which is not relevant to algorithms.


julianschacher:
For all the chapters that get moved and not deleted: we should maybe add redirects for them.

leios:
yup, they should. I'll get the chapters consistent and then see about the redirects.

julianschacher:
The restructuring with the latest changes and the redirects looks good :)
Do you plan to introduce more changes on this PR?

leios:
Nope, that was it, I think. It's good for a review, if you are up for it.

julianschacher:
`As a note, this also hides work-in-progress sections from the site.` what do you mean by that? There are still chapters that just say "coming soon".

julianschacher:
I think we should always have the Title of each page in the biggest style i.e. `#`.
If you agree, `how_to_contribute.md`, `version_control.md`, `notation.md` and `differential_equations.md`, `physics_solvers.md`, `backward_euler_method.md` would need a change regarding the title.

leios:
I think these latest commits fix those problems.

julianschacher:
Everything seems fine, I have no more complains.

PR #256, opened 2018-07-16
Reimplemented Verlet Integration in Python

Butt4cak3:
During a recent discussion about #243 I noticed that the Python version of our Verlet Integration example code is vastly different from all other implementations. It creates a `Simulation` class and the 3 Verlet algorithms are classes that inherit from that `Simulation` class.

I changed the code to more closely match the other language implementations because I think that introducing classes and inheritance to a piece of example code like this is a bit excessive. Tell me what you think.

### DIFF
--- a/contents/verlet_integration/code/python/verlet.py
+++ b/contents/verlet_integration/code/python/verlet.py
@@ -4,7 +4,8 @@ def verlet(pos, acc, dt):
 
     while pos > 0:
         time += dt
-        prev_pos, pos = pos, pos * 2 - prev_pos + acc * dt * dt
+        next_pos = pos * 2 - prev_pos + acc * dt * dt
+        prev_pos, pos = pos, next_pos
 
     return time
 
@@ -15,7 +16,8 @@ def stormer_verlet(pos, acc, dt):
 
     while pos > 0:
         time += dt
-        prev_pos, pos = pos, pos * 2 - prev_pos + acc * dt * dt
+        next_pos = pos * 2 - prev_pos + acc * dt * dt
+        prev_pos, pos = pos, next_pos
         vel += acc * dt
 
     return time, vel
--- a/contents/verlet_integration/code/python/verlet.py
+++ b/contents/verlet_integration/code/python/verlet.py
@@ -1,72 +1,53 @@
-# KingFredrickVI
-class Ball:
+def verlet(pos, acc, dt):
+    prev_pos = pos
+    time = 0
 
-    def __init__(self, pos=0, vel=0, acc=0):
-        self.pos = pos
-        self.vel = vel
-        self.acc = acc
+    while pos > 0:
+        time += dt
+        prev_pos, pos = pos, pos * 2 - prev_pos + acc * dt * dt
 
+    return time
 
-class Simulation:
+def stormer_verlet(pos, acc, dt):
+    prev_pos = pos
+    time = 0
+    vel = 0
 
-    def __init__(self, obj, dt = 0.01):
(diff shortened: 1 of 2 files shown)


chapters/algorithms/verlet_integration/code/python/verlet.py
```diff
@@ -1,72 +1,53 @@
...
+    prev_pos = pos
+    time = 0
 
-    def __init__(self, pos=0, vel=0, acc=0):
-        self.pos = pos
-        self.vel = vel
-        self.acc = acc
+    while pos > 0:
+        time += dt
+        prev_pos, pos = pos, pos * 2 - prev_pos + acc * dt * dt
```
Butt4cak3: I don't know about this one. I thought about changing it to the following, but I wasn't sure, so feel free to tell me your opinion.
```python
next_pos = pos * 2 - prev_pos + acc * dt * dt
prev_pos, pos = pos, next_pos
```

chapters/algorithms/verlet_integration/code/python/verlet.py
```diff
@@ -1,72 +1,53 @@
...
-    print("Velocity Verlet:", sim.time)
-
-
-if __name__ == "__main__":
+    time = verlet(5, -10, 0.01)
+    print("Verlet")
+    print("Time: {:.10f}".format(time))
+    print()
+
+    time, vel = stormer_verlet(5, -10, 0.01)
```
leios: I think other methods just return `time`?

chapters/algorithms/verlet_integration/code/python/verlet.py
leios: Your method is definitely easier to read.

chapters/algorithms/verlet_integration/code/python/verlet.py
Butt4cak3: That's correct. I thought that if I already have the `vel` variable, I might as well do *something* with it. Especially in `stormer_verlet` in Julia, the `vel` variable is assigned a value but it is never actually used anywhere.

But I can get it in line with everything else. I probably should, since I was talking about bringing language implementations closer to each other.

chapters/algorithms/verlet_integration/code/python/verlet.py
Butt4cak3: They're both my methods. I just wasn't sure which one to use. Which one do you mean?

chapters/algorithms/verlet_integration/code/python/verlet.py
leios: Ah, you are right. The stormer-verlet is basically just verlet with a velocity term. Maybe we should output it?

chapters/algorithms/verlet_integration/code/python/verlet.py
leios: Oh, the second one.

Butt4cak3: Yeah, I agree.

chapters/algorithms/verlet_integration/code/python/verlet.py
Butt4cak3: I mean... My answer to that is in the code =P

leios: Yeah, leave it. I think it's fine and an improvement.

jiegillet:
Gosh, imagine that, conflicts cause by files being moved around.  @Butt4cak3, can you fix these?

PR #258, opened 2018-07-16
Implement Gaussian Elimination in C++

antonte:


chapters/algorithms/gaussian_elimination/gaussian_elimination.md
```diff
@@ -259,6 +259,8 @@ In code, this looks like:
 [import:1-42, lang:"julia"](code/julia/gaussian_elimination.jl)
 {% sample lang="c" %}
 [import:13-44, lang:"c_cpp"](code/c/gaussian_elimination.c)
+{% sample lang="cpp" %}
+[import:6-51, lang:"c_cpp"](code/c++/gaussian_elimination.cpp)
```
antonte: Link to back_substitution function is missing

leios:
Hey, sorry for this @antonte (I know you have been doing a lot of updates and appreciate it), but we just changed the gaussian elimination code to include a gauss-jordan function to be a bit more general, see #219 

We're getting to all the PR's sorry it takes so long!

antonte:
@leios no problems, I'll redo implementation this night. Closing the PR. I'll reopen it after redoing the implementation.

antonte:
Pushed version with gaussian_jordan

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,133 @@
...
+void gaussian_elimination(std::vector<double>& a, int cols) {
+  assert(a.size() % cols == 0);
+  int rows = a.size() / cols;
+
+  int row = 0;
+
+  // Main loop going through all columns
+  for (int col = 0; col < cols - 1; ++col) {
+    // Step 1: finding the maximum element for each column
+    int max_index = [&]() {
```
Gathros: The lambda function doesn't help with readability and there is no need too.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,133 @@
...
+      int res = row;
+      int max_element = a[col + row * cols];
+      for (int r = row + 1; r < rows; ++r)
```
Gathros: The for loop has no curly bracket.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,133 @@
+#include <algorithm>
+#include <cassert>
+#include <cmath>
+#include <iostream>
+#include <vector>
+
+void gaussian_elimination(std::vector<double>& a, int cols) {
+  assert(a.size() % cols == 0);
```
Gathros: no need for `assert`.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,133 @@
...
+    std::cout << "[";
+    for (int j = 0; j < cols; ++j) {
+      std::cout << a[j + i * cols] << " ";
+    }
+    std::cout << "]\n";
+  }
+}
+
+int main() {
+  std::vector<double> a = {2, 3, 4, 6, 1, 2, 3, 4, 3, -4, 0, 10};
```
Gathros: Make this the same as in the julia code.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+}
+
+void print_matrix(const std::vector<double>& a, int cols) {
+  assert(a.size() % cols == 0);
+  int rows = a.size() / cols;
+  for (int i = 0; i < rows; ++i) {
+    std::cout << "[";
+    for (int j = 0; j < cols; ++j) {
+      std::cout << a[j + i * cols] << " ";
```
Gathros: use `"\t"` and not a space for better formatting.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,133 @@
...
+void gauss_jordan_elimination(std::vector<double>& a, int cols) {
+  assert(a.size() % cols == 0);
+  // After this, we know what row to start on (r-1)
+  // to go back through the matrix
+  int row = 0;
+  for (int col = 0; col < cols - 1; ++col) {
+    if (a[col + row * cols] != 0) {
+
+      // divide row by pivot and leaving pivot as 1
+      for (int i = cols - 1; i >= static_cast<int>(col); --i)
```
Gathros: why cast `col` from int to int.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+std::vector<double> back_substitution(const std::vector<double>& a, int cols) {
+  assert(a.size() % cols == 0);
+  int rows = a.size() / cols;
+
+  // Creating the solution Vector
+  std::vector<double> soln(rows);
+
+  // initialize the final element
+  soln[rows - 1] =
```
Gathros: you don't need this.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: This technique called  IIFE stands for an immediately-invoked function expression, it was popularized in JS and adopted in C++. It has some benefits comparing with old style:
- initialize the variable at declaration
- keep local var in the lambda in the scope

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: Will fix it, I did not know about it. Do we have any style guide document, so in the future I would not make this sort of mistakes? Like snake_case and enforced curly brackets?

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
Gathros: Fair enough, but it's still hard to read. I've asked around on discord and people agree.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: I agree we should not assert on user input, but what you can offer instead?

I can see multiple options: 
- remove assert: the code will crash later of even worse return incorrect result
- log the error and crash: same as assert but more lines of code
- return the error code: for production solution it might be OK, but it requires a lot of boiler plate code, and hides the important parts of the code 
- throw an exception: for production solution it might be OK, but it requires some of boiler plate code, and hides the important parts of the code

Conclusion: for this particular case, sample code for the book about algorithms I prefer to use assert.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: ups

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
Gathros: Have an if statement there like `if (a.size() % cols != 0) {`.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: I copied from Jula code:
    # initialize the final element
    soln[rows] = A[rows, cols] / A[rows, cols-1]

Is Jula code has a bug?

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: Ups

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
Gathros: I'll tell @leios. But looking at the for loop, when `i` is rows - 1 sum will be 0. This is because the first for loop will just be jumped making that line useless since it will happen later.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: Not sure exactly about your motivation, but it is fine for me. I'll make the changes.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
Gathros: Sorry that was a bad statement I wrote. You have a fair point but I have asked around and people think it's hard to read.

antonte: Your suggestion is conflicting with C++ Core Guidelines: ES.5: Keep scopes small, ES.20: Always initialize an object. To avoid this without using lambda I would need to create named function.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
antonte: I found the bug, it is next line; should be -2 instead of -1.

Gathros: It looks better if you remove that line.

antonte: You right. I'll make a pull request for Julia code as well.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
Gathros: The lambda function makes it hard for people to understand your c++ code if they are new to the language. Also my C code wasn't long:

```C
for (size_t i = k + 1; i < rows; ++i) {
    if (fabs(a[i * cols + k]) > fabs(a[pivot * cols + k])) {
        pivot = i;
    }
}
```

antonte: Let's wait what @leios would say.

- Modern C++ uses lambda quite aggressively it was introduces in C++ 11 more than 7 years ago
- Without IIFE it would be hard to follow C++ Core Guidelines

leios: To be clear, we are comparing the following code samples:

```cpp
  for (int col = 0; col < cols - 1; ++col) {
    // Step 1: finding the maximum element for each column
    int max_index = [&]() {
      int res = row;
      int max_element = a[col + row * cols];
      for (int r = row + 1; r < rows; ++r) {
        if (max_element < std::abs(a[col + r * cols])) {
          max_element = std::abs(a[col + r * cols]);
          res = r;
        }
      }
      return res;
    }();
...
}
```

and 

```cpp
  for (int col = 0; col < cols - 1; ++col) {
    // Step 1: finding the maximum element for each column
    int max_index = col + row * cols;
    for (size_t i = col + 1; i < rows; ++i) {
        if (fabs(a[i * cols + col]) > fabs(a[max_index * cols + col])) {
            max_index = i;
        }
    }
...
}
```

In my opinion, the latter one is easier to read (unless I screwed something up and this doesn't actually do the right action). I feel I am not fully understanding the conversation. @antonte How does the second code snippet violate the core guidelines? I don't think it adds another variable to the scope and doesn't violate ES.20.

antonte: @leios you right. I'll make changes. The problem in my code is not about using IIFE, but about caching max_element.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,137 @@
...
+  int row = 0;
+  for (int col = 0; col < cols - 1; ++col) {
+    if (a[col + row * cols] != 0) {
+
+      // divide row by pivot and leaving pivot as 1
+      for (int i = cols - 1; i >= col; --i)
+        a[i + row * cols] /= a[col + row * cols];
+
+      // subtract value from above row and set values above pivot to 0
+      for (int i = 0; i < static_cast<int>(row - 1); ++i)
```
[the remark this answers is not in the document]
antonte: man... so many static_casts... :( 

```row - 1 ``` looks right, we are not touching current ```row``` we only updating all rows above.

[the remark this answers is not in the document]
Comment (leios) [reply contents/gaussian_elimination/code/c++/gaussian_elimination.cpp]: So, when running this with `row`, we get the right answer:

```
[1.000000	0.000000	0.000000	1.636364	]
[0.000000	1.000000	0.000000	-1.272727	]
[0.000000	0.000000	1.000000	1.636364	]
```

but with `row - 1`, we get:

```
[1.000000	-1.333333	0.000000	3.333333	]
[0.000000	1.000000	0.705882	-0.117647	]
[0.000000	0.000000	1.000000	1.636364	]
```

which means we are ignoring the row *just above* the row we want too.

as far as the `static_cast`s are concerned, if there's a reason to leave them in, it's no problem. I was more curious about it than anything.

File: contents/gaussian_elimination/code/c++/gaussian_elimination.cpp [hunk 10]
```diff
@@ -0,0 +1,132 @@
...
+    std::cout << "]\n";
+  }
+}
+
+int main() {
+  std::vector<double> a = { 2, 3, 4, 6,
+                            1, 2, 3, 4,
+                            3, -4, 0, 10 };
+  const int cols = 4;
+  if (a.size() % cols != 0)
```
Gathros: Why not have the if statement in the other functions replacing `assert`. People are going to take them and not the main function when they need a Gaussian Elimination.

jiegillet: I think you should get rid of all `assert`s and not even replace them with `if`s. We can just always assume that the input is correct for the sake of simplicity. 
If you still want to have some checks, then inside the `if` you can have `return 0;` or a zero matrix or something.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    // Check to make sure matrix is good!
+    if (a[col + max_index * cols] == 0) {
+      std::cout << "matrix is singular!\n";
+      continue;
+    }
+
+    // Step 2: swap row with highest value for that column to the top
+    for (int c = 0; c < cols; ++c)
```
jiegillet: You should check if `row== max_index `, if it's the case you don't need to swap.

contents/gaussian_elimination/code/c++/gaussian_elimination.cpp
```diff
@@ -0,0 +1,130 @@
...
+  int row = 0;
+  for (int col = 0; col < cols - 1; ++col) {
+    if (a[col + row * cols] != 0) {
+
+      // divide row by pivot and leaving pivot as 1
+      for (int i = cols - 1; i >= col; --i)
+        a[i + row * cols] /= a[col + row * cols];
+
+      // subtract value from above row and set values above pivot to 0
+      for (int i = 0; i < row; ++i)
```
Gathros: Add brackets to this for loop.

leios:
Hey guys, what is the status on this PR?

leios:
I think there were only a few smallscale changes requested on this PR, but we have another Gaussian Elimination PR now. If this one is stale, we'll close it and work on the next one.

Gathros:
This pull request is stale, so I will close it. You had months (since Oct 3) to reply and in that time you've been online on github.

PR #261, opened 2018-07-17
Implement Graham Scan in C++

antonte:


contents/graham_scan/code/c++/graham.cpp
```diff
@@ -0,0 +1,66 @@
...
+      std::begin(points) + 1,
+      std::end(points),
+      [&](const Point& a, const Point& b) {
+        return atan2(a.y - points[0].y, a.x - points[0].x) <
+               atan2(b.y - points[0].y, b.x - points[0].x);
+      });
+
+  // M will be the point on the hull
+  auto M = 2u;
+  for (auto i = 0u; i < N; ++i) {
```
Gathros: why are you using auto so much, it's hard to read.

antonte: Following C++ Core Guideline: ES.11: Use auto to avoid redundant repetition of type names

antonte: In both lines above the type of vars is clear: ```2u``` is ```unsigned int```, ```0u``` is ```unsigned int```. Core Guidelines states: Use auto to avoid redundancy. I want to use ```unsigned int``` and writing it instead of writing ```auto``` violates the guideline.

Reply to the comment about big vector bug: It is more work to make code work for cases if array size is bigger 4 billion elements and code does not look nice. I believe handling cases for arrays bigger 4 billion elements is special case and require special treatment.

leios: I realize this discussion happened a while ago, I am catching up now. Reading the guidelines (ES.11), there are a few notable exceptions to using `auto`:
> Exception: Avoid auto for initializer lists and *in cases where you know exactly which type you want* and where an initializer might require conversion.

(I added emphasis)

So it would seem that there is no problem using `unsigned int` or `size_t`because those are the specific types we want here. We are also not using the types incessantly and are working on code at a fundamentally different scale than the guideline was created for.

In my opinion, `unsigned int M = 2` is understandable no matter what background you have `auto M = 2u` is not.

PR #264, opened 2018-07-17
Rework C++ implementation for Huffman encoding

antonte:
- Make implementation shorter and easier to understand
- Make namespace hierarchy flat
- Combine Leaf and Branch in one class Node for the tree
- Make codebook explicit
- Use std::string for string_to_encode
- Remove noise (delete function with_new_bit() for instance)
- Use unordered map for weights instead of vector
- Instead inserting to the orphan array, sort it on every iteration (debatable)
- Remove structure for cacheing bits for character, calculate it every time
- Remove error handling
- Change the encoding interface: encode() and decode() instead huffman::encode_string() and encoded.decoded()

### DIFF
--- a/contents/huffman_encoding/code/c++/huffman.cpp
+++ b/contents/huffman_encoding/code/c++/huffman.cpp
@@ -12,9 +12,12 @@ struct node {
       : left(std::move(a_left)), right(std::move(a_right)),
         weight(left->weight + right->weight) {
     left->parent = this;
-    left->branch = branch::left;
     right->parent = this;
-    right->branch = branch::right;
+  }
+
+  bool branch() const
+  {
+    return parent->right.get() == this;
   }
 
   // leaf fields
@@ -26,8 +29,6 @@ struct node {
 
   // common fields
   node* parent = nullptr;
-  enum class branch { left, right };
-  branch branch = branch::left;
   int weight = 0;
 };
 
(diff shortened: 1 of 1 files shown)


Butt4cak3:
Could you please provide some explanation for why you think this rework was necessary and why your version is better than the current?

antonte:
Updated commit message and message in PR to emphasize why rework is needed.

antonte:
@Gustorn no problem, I'll rename Node and Codebook to lower case.

antonte:
@Gustorn changes with snake_case are pushed

contents/huffman_encoding/code/c++/huffman.cpp
```diff
@@ -1,268 +1,124 @@
...
-  std::vector<bool> encode(Iter first, Iter last) const;
+  // branch fields
+  std::unique_ptr<node> left;
+  std::unique_ptr<node> right;
 
-  template <typename Iter>
-  std::string decode(Iter first, Iter last) const;
+  // common fields
+  node* parent = nullptr;
+  enum class branch { left, right };
```
Gustorn: I don't think storing which branch we're on is necessary.

antonte: What do you mean? It used here:             
```C++
compressed_char.push_back(node->branch != node::branch::left);
```
https://github.com/antonte/algorithm-archive/blob/71c0d4295d4a892c91e1da64fdb7feaa9ea56dd3/contents/huffman_encoding/code/c%2B%2B/huffman.cpp#L83

leios:
Hey guys. What's the status on this PR? It looks like the review was completed; however, it was not yet merged into the AAA. Can I dismiss the review @Gustorn and merge?

Gathros:
Another stale pull request.

PR #267, opened 2018-07-17
Updating C implemintation of Gaussian Elimination.

Gathros:


leios:
@antonte What questions about the original implementation?

antonte:
@leios they are minor:

- it feel unsafe to use size_t since we are iterating backward, I would prefer to use int
- declaring 2d array and casting it to double*
- using 3 and 4 instead of declaring constants cols and rows

chapters/algorithms/gaussian_elimination/code/c/gaussian_elimination.c
```diff
@@ -2,49 +2,53 @@
...
     for (size_t j = 0; j < cols; ++j) {
         double tmp = a[i * cols + j];
         a[i * cols + j] = a[pivot * cols + j];
         a[pivot * cols + j] = tmp;
     }
 }
 
-void gaussian_elimination(double *a, const size_t rows, const size_t cols) {
-    size_t min_dim = (rows < cols)? rows: cols;
+void gaussian_elimination(double * a, const size_t rows, const size_t cols) {
```
Gustorn: Did you change the pointer alignment formatting on purpose? Having it attached to neither the type nor the variable is one of the less frequently used versions.

leios:
Screwed up with git a bit, but this is in the AAA: https://github.com/algorithm-archivists/algorithm-archive/commit/f110a471985d3063f97c37b53de10146fddff5fa

Thanks!

PR #249, opened 2018-07-15
Split op updates

leios:
This PR will make large-scale changes to the text of the quantum systems chapter and split-op, adding in imaginary time evolution and several important quantities to understanding quantum systems.

At this point, I will be making a pass through for revisions and also adding in a number of citations that I am missing (I cannot find too many really good ones, but I'll add what I can).

Basically, there are still some somewhat heavy revisions necessary, but I thought I would put up a PR in the case anyone saw something that I was clearly missing at this point

PR #274, opened 2018-07-19
Fix references to Julia code in split-operator method

antonte:


leios:
Hey @antonte , thanks for the PR! I think this has been fixed in the split op updates PR #249 , which will change the code a good amount. I'll go ahead and merge it for now, but the changes are coming.

PR #276, opened 2018-07-19
Minor clean up for cooley tukey in C++

antonte:
- rename c64 to complex
- make pi a constant instead of function
- fix complilation warnings about comparing signed int with unsigned int

### DIFF
--- a/contents/cooley_tukey/code/c++/fft.cpp
+++ b/contents/cooley_tukey/code/c++/fft.cpp
@@ -26,8 +26,7 @@ void dft(Iter X, Iter last) {
   std::vector<complex> tmp(N);
   for (auto i = 0; i < N; ++i) {
     for (auto j = 0; j < N; ++j) {
-      using namespace std::literals::complex_literals;
-      tmp[i] += X[j] * exp(-2.0 * M_PI * i * j / N * 1i);
+      tmp[i] += X[j] * exp(complex(0, -2.0 * M_PI * i * j / N));
     }
   }
   std::copy(std::begin(tmp), std::end(tmp), X);
--- a/contents/cooley_tukey/code/c++/fft.cpp
+++ b/contents/cooley_tukey/code/c++/fft.cpp
@@ -20,6 +20,19 @@ using std::size_t;
 using complex = std::complex<double>;
 static const double pi = 3.14159265358979323846264338327950288419716;
 
+template <typename Iter>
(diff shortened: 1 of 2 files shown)


chapters/algorithms/cooley_tukey/code/c++/fft.cpp
```diff
@@ -30,12 +27,12 @@ void cooley_tukey(Iter first, Iter last) {
   if (size >= 2) {
     // split the range, with even indices going in the first half,
     // and odd indices going in the last half.
-    auto temp = std::vector<c64>(size / 2);
-    for (size_t i = 0; i < size / 2; ++i) {
+    auto temp = std::vector<complex>(size / 2);
+    for (int i = 0; i < size / 2; ++i) {
```
Gustorn: Similarly to the other PR, this should actually be `ptrdiff_t`

antonte: Copy from the other PR:

@Gustorn I decided to use int instead of ptrdiff_t to improve the readability of the code, not to many people know about ptrdiff_t. And to overflow int on modern computers for vector of complex values we would need to have 2 billions elements in the array which will consume 8 (for double) * 2 (for real and imaginary) * 2 (billions elements) = 32 GB of memory. Kind of unlikely.

Gathros:
Since your changing the code, can you add the `dft` function. Look at the C code to understand how it works.

contents/cooley_tukey/code/c++/fft.cpp
```diff
@@ -17,10 +17,20 @@ using std::swap;
...
+static const double pi = 3.14159265358979323846264338327950288419716;
+
+template <typename Iter>
+void dft(Iter X, Iter last) {
+  const auto N = last - X;
+  std::vector<complex> tmp(N);
+  for (auto i = 0; i < N; ++i) {
+    for (auto j = 0; j < N; ++j) {
+      using namespace std::literals::complex_literals;
+      tmp[i] += X[j] * exp(-2.0 * M_PI * i * j / N * 1i);
```
Gathros: Why not do this, `tmp[i] += X[j] * exp(complex(0, -2.0 * M_PI * i * j / N));`.

antonte: ok

PR #277, opened 2018-07-19
Minor clean up for convolutions in C++

antonte:
- rename c64 to complex
- make pi a constant instead of function
- use int instead of ptrdiff_t
- use for-loop instead of for_each

### DIFF
--- a/contents/convolutions/code/c++/convolutions.cpp
+++ b/contents/convolutions/code/c++/convolutions.cpp
@@ -15,26 +15,22 @@ using std::begin;
 using std::end;
 using std::swap;
 
-using std::ptrdiff_t;
 using std::size_t;
 
-using c64 = std::complex<double>;
-template <typename T>
-constexpr T pi() {
-  return 3.14159265358979323846264338327950288419716;
-}
+using complex = std::complex<double>;
+static const double pi = 3.14159265358979323846264338327950288419716;
 
 // This section is not a part of the algorithm
 template <typename Iter>
 void fft(Iter const first, Iter const last) {
   auto const size = last - first;
   if (size >= 2) {
-    auto temp = std::vector<c64>(size / 2);
-    for (ptrdiff_t i = 0; i < size / 2; ++i) {
+    auto temp = std::vector<complex>(size / 2);
+    for (int i = 0; i < size / 2; ++i) {
       temp[i] = first[i * 2 + 1];
       first[i] = first[i * 2];
     }
-    for (ptrdiff_t i = 0; i < size / 2; ++i) {
+    for (int i = 0; i < size / 2; ++i) {
       first[i + size / 2] = temp[i];
     }
 
@@ -42,8 +38,8 @@ void fft(Iter const first, Iter const last) {
     fft(first, split);
     fft(split, last);
 
-    for (ptrdiff_t k = 0; k < size / 2; ++k) {
-      auto w = std::exp(c64(0, -2.0 * pi<double>() * k / size));
+    for (int k = 0; k < size / 2; ++k) {
+      auto w = std::exp(complex(0, -2.0 * pi * k / size));
 
       auto& bottom = first[k];
       auto& top = first[k + size / 2];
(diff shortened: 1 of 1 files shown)


chapters/algorithms/convolutions/code/c++/convolutions.cpp
```diff
@@ -15,35 +15,31 @@ using std::begin;
...
 
 // This section is not a part of the algorithm
 template <typename Iter>
 void fft(Iter const first, Iter const last) {
   auto const size = last - first;
   if (size >= 2) {
-    auto temp = std::vector<c64>(size / 2);
-    for (ptrdiff_t i = 0; i < size / 2; ++i) {
+    auto temp = std::vector<complex>(size / 2);
+    for (int i = 0; i < size / 2; ++i) {
```
Gustorn: I like the rest of the changes but this one is actually just incorrect if the vector happens to be huge. There's a good reason to use `ptrdiff_t` here

Gorzoid: Would size_t not make more sense? Or std::vector<c64>::size_type.
ptrdiff_t doesn't express the usage of the variable very well, since it's being used to index vector.

chapters/algorithms/convolutions/code/c++/convolutions.cpp
```diff
@@ -55,16 +51,17 @@ void fft(Iter const first, Iter const last) {
 
 template <typename Iter>
 void inverse_fft(Iter const first, Iter const last) {
-  std::for_each(first, last, [](auto& it) { it = std::conj(it); });
+  for (auto it = first; it != last; ++it)
+    *it = std::conj(*it);
```
Gorzoid: This seems like std::transform would be fitting here, I kinda prefer these functions to a for loop, especially since we are given first and last iterators.
Same for line 60

chapters/algorithms/convolutions/code/c++/convolutions.cpp
antonte: @Gorzoid Probably size_t is fine to use here, the down side of using size_t in this code, it will generate at some places compilation warnings for instance in the following line:
```c++
auto const size = last - first;
```
it will make size singed integer, and down the road compiler will complain about comparing singed with unsigned integers. So to fix the compilation warning I would be forced to use static_cast<size_t>, which is little bit ugly.

@Gustorn I decided to use int instead of ptrdiff_t to improve the readability of the code, not to many people know about ptrdiff_t. And to overflow int on modern computers for vector of complex values we would need to have 2 billions elements in the array which will consume 8 (for double) * 2 (for real and imaginary) * 2 (billions elements) = 32 GB of memory. Kind of unlikely.

chapters/algorithms/convolutions/code/c++/convolutions.cpp
antonte:
```c++
std::transform(first, last, [](complex& x) { x = std::conj(x); });
for (auto it = first; it != last; ++it) *it = std::conj(*it);
```
As you can see here old school for-loop is slightly shorter, also it does not require knowledge about STL algorithms, but I am fine with std::transform(), and if you insist I am OK to change it to std::transform().

chapters/algorithms/convolutions/code/c++/convolutions.cpp
Gustorn: `int` is actually just incorrect though. You can cast `size` to `size_t`: if we assume that `last` is actually after `first` then this is a safe operation and it does solve the problem. Same goes for the other PR

leios:
Hey, it looks like this PR is good to go, but it is blocked by @Gustorn. Should I just dismiss the review and merge?

chapters/algorithms/convolutions/code/c++/convolutions.cpp
Gorzoid: Sorry never responded to this, I wouldn't insist on using transform, it isn't even fully suited for this purpose since it acts on itself and not another array. we could do something like 
```c++
std::transform(first, last, first, std::conj);
```
if std::conj wasn't an overloaded function ): really wish C++ templates would allow functions to propagate without needing to specify overload.

chapters/algorithms/convolutions/code/c++/convolutions.cpp
Gorzoid: I'd say making that line
```c++
const size_t size = static_cast<size_t>(last - first); // or size_t(last - first)
```
It might be a bit ugly but I think it is appropriate since we can assume last >= first.
int is still acceptable though as it is kinda the goto data type for this kinda thing, you can pretty safely use it in most contexts. size_t is the correct approach, while int is the reader friendly approach.

leios:
I think `size_t` is preferred here, right?

Gorzoid:
size_t is the `correct` approach, I personally prefer it despite it requiring some verbose casting. antonte prefers the simplicity of int and I see his point. Merging now would finalize the use of int, we probably need somebody else to voice their thoughts on it before coming to a conclusion.

berquist:
What about `unsigned`? `int` should _never_ be used if it is physically incorrect for a variable to be negative, regardless of the size.

Gathros:
I'm closing this pull request since there is no convolutions chapter now.

PR #279, opened 2018-07-19
Quicksort

MarcoPorracin:


leios:
Hey @MarcoPorracin thanks for the PR! I see that you have code and text. It will take me some time to get to this for a review (sorry about that). Expect some somewhat heavy revisions in the text and possibly the code. The ultimate test for chapter submission is whether I can implement it in code using only the text provided.

I might send some updates to your fork too. It depends on the situation.

MarcoPorracin:
@leios Dont worry im totally aware and expecting this kind of changes in fact.
I'm glad to help and don't hesitate on sending/asking me for anything you might need.

lulucca12:
@Gustorn I think that what @MarcoPorracin did wasn't necessarily bad, it was just another style in a way. In Haskell, by the language restrictions in abstraction, it is impossible to mutate a value given by an argument (as it generates a side effect). As in Javascript, I see that in this implementation it is used a very imperative style and if keeping it I agree with you in making the code correct in space and time complexity by changing the original array passed into the function.

MarcoPorracin:
@julianschacher You don't insult me and no i did not i just google it ...

julianschacher:
@MarcoPorracin Then we can't accept the code to be part of the AAA, since we can't just use code other people wrote for this book.

MarcoPorracin:
@julianschacher why not? as long as it does not have any copyright... Anyway i can write new code myself i just didn't thought it was necessary.

julianschacher:
@MarcoPorracin The code should be written with the chapter in mind by people who want to contribute to the AAA. Just copy-pasting stuff just doesn't match the philosophy of this book.

MarcoPorracin:
@julianschacher Ok! Give me a couple of days and i ll rewrite the code myself considering the comments before. Luckily i only searched for languages im actually familiar with so this should be no problem.
Also, if anyone else want to do any if this is more than welcome.

MarcoPorracin:
@julianschacher Re thinking about this, maybe the text should be reworked first and then code the algorithms based on how the text presents them? What do you think?

julianschacher:
Ok, so I read the chapter you submitted and found quite a few more issues:
- I'm not an expert regarding Quicksort, but Wikipedia suggests, that the topic can include a lot more detail. I'm reading the following line there: "The pivot selection and partitioning steps can be done in several different ways; the choice of specific implementation schemes greatly affects the algorithm's performance. " So it might be interesting to mention the different ways this can be done. Your text doesn't mention these ways at all.
- The english in your text isn't great. I understand, that you might not be a native English speaker, but it affects the quality of the text nonetheless. 
- You didn't include any code imports in the chapter. Code imports are used in every other chapter and are a core concept used by the AAA.
- The complexity of Quicksort doesn't get addressed. 
- It's all written in a "Let's take an example" perspective. I don't know, if we want that.

Also your commit messages don't give much explanation of what the commit 

I appreciate, that you put effort into the creation of this PR, but the fact that you googled implementations and just pasted them in here as well as the overall quality and detail-level of the text suggest that you didn't put **too** much care into it. So I think we should just close this PR and you can submit a new one with a rewritten version of this Quicksort text, if you want.

MarcoPorracin:
As suggested i might be opening a new PR with a new approach.

julianschacher:
Thanks for understanding. I hope you understand the concerns I have and hope you didn't felt hurt by my rather harsh comment.

MarcoPorracin:
@julianschacher Don't worry ! We learn from our mistakes .. The only point i might not be able to fix is my english but i ll just ask for a review in order to fix that. (when the moment comes)

julianschacher:
I think we can fix and help with that, when the moment comes.

leios:
I just saw the updates here. I agree with all the points @julianschacher made. The text needs a rework to address all possible styles quicksort can be written in and there should be code provided for each of these. In the end, you should have a working code sample that runs all examples and you should take snippets from that for each example implementation in the main text. For the first attempt, just stick to one language (python, preferably).

I realize I failed at a complete description for bubble sort, bogo sort, and Gale--Shapley. Bubble sort and Gale--Shapley are on the list of algorithms to update (it's already a long list).

Quicksort is also an incredibly important algorithm, so I would suggest adding in some citations for people to read about the history of the algorithm and such... At least this one: https://dl.acm.org/citation.cfm?doid=366622.366644 (I know I am lacking in this area too, so just add whatever citations you used that were not from wikipedia). 

If I'm honest, one of the reasons I have stayed away from writing a chapter on quicksort (or any "real" sorting algorithm) is because it needs to be done right, and I am not confident enough in my writing abilities to do them right now. 

I don't mind you reworking the quicksort text, but know that it will be a somewhat long and rigorous reviewing process which can only happen when we both have time for it.

PR #280, opened 2018-07-19
Added a note in README.md for new contributors

Butt4cak3:
We're receiving an increasing number of PRs of people trying to submit entirely new chapters. Even though @leios decided that he would allow some people to write new chapters, it's not an entirely open-to-the-public process yet.

This PR adds a very visible note right at the top of the README.md file in the root of this repository. I think a note like this is long overdue. New contributors cannot know the current situation when it comes to chapter contributions. With this note, we can at least point to something when we close such PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,11 @@
+> **IMPORTANT NOTE FOR NEW CONTRIBUTORS:**
+>
+> We do **not** yet accept entirely new chapters by everyone. If you would like to start work on one, get in contact with Leios first. If you create a full chapter, including text, and submit it as a pull request it is most likely going to get rejected.
+>
+> If you want to help, it is best to write language examples for *existing* chapters. You can also try to find spelling or other mistakes in existing chapters and submit fixes for those.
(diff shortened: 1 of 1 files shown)


leios:
Yeah, this is a good idea. As a note, the following statement is already in the *How to Contribute* page: "At this point, I am trying to figure out the best way to balance community contributions and text. Right now, I feel comfortable writing the text associated with each algorithm and asking for the community to write individual implementations. In the future, I might allow other users to write algorithm chapters, but for now let's keep it simple: I'll do the writing, everyone else does the coding."

I thought that would be enough. I'll allow chapter submissions if people talk to me about it first.

Butt4cak3:
@leios The thing is that it's just not visible enough in my opinion. Otherwise we wouldn't have this number of chapter pull requests right now. But if this doesn't help, then I don't know what does.

Butt4cak3:
We can always remove it when we feel it's not necessary anymore.

PR #283, opened 2018-07-20
Added Jarvis March for C++

Gorzoid:
There's a few things I was unsure of while making this, for input I used template ForwardIter, which allows it to be used with multiple datastructures. I wasn't sure, however whether to return a vector, or use an OutputIter argument. The vector option is simpler but kind of ruins the point of user template iterators as input. Due to this algorithm not having a predetermined output size (up to the size of the points), I feel using an OutputIter might lead to buffer overflows if used incorrectly.

Seeing as this is meant to be informative rather than a standardized algorithm, I feel like I should just keep the vector return. Using a vector as the input would hardly change the code and I still kinda like using range based algorithms.


Butt4cak3:
Hey @Gorzoid and welcome and thanks for contributing! We currently have a *lot* of open C++ pull requests, so I think we're a little short on avilable C++ reviewers. Your PR might take a little longer than usual to get reviewed because of this. I hope you don't mind.

Gorzoid:
That's grand, yeah I can see there are a lot C++ PR's.

contents/jarvis_march/code/c++/jarvis_march.cpp
```diff
@@ -0,0 +1,64 @@
...
+    {
+        hull_points.push_back(*next_point_it);
+
+        const Point& p1 = hull_points.back();
+
+        // Largest internal angle
+        next_point_it = std::max_element(
+            start,
+            end,
+            [p1](const Point& p2, const Point& p3){
```
antonte: ```[&p1]``` is slightly better.

contents/jarvis_march/code/c++/jarvis_march.cpp
```diff
@@ -0,0 +1,64 @@
...
+    double x, y;
+
+    bool operator==(const Point& b) const
+    {
+        return x == b.x && y == b.y;
+    }
+};
+
+template<typename ForwardIter>
+std::vector<Point> jarvis_march(ForwardIter start, ForwardIter end)
```
antonte: Take as input ```const std::vector<Point> &```

antonte: It crashes in case of empty set. Should return an empty vector.

contents/jarvis_march/code/c++/jarvis_march.cpp
Gustorn: This is actually incorrect, `[p1]` will generate better code since the `Point` struct is fairly small (and so you can pass it in registers). Since you don't need to mutate the captured variable, I would stick with the by value version

contents/jarvis_march/code/c++/jarvis_march.cpp
Gorzoid: If I take out the do...while I need to add extra code to add the first element. Anyway the main problem is the min_element to find the left most point. If empty it will return points.end(), I can't think of any way to resolve that would result in shorter code.

PR #287, opened 2018-07-21
Minor improvement in Gaussian Elimination in Julia

antonte:
- Remove extraneous assignment


leios:
I'm 50:50 on this one. It makes it ~2% harder to read (simply because you have to think about why it works for the first element at all without throwing an error in the internal `for` loop) and adds another unnecessary assignment for `sum`. Ultimately not a big deal either way.

Obviously, I don't mind changing things if you guys think it's better though! 

As another note: would you mind updating the corresponding `.md` file with the change in line-numbers if we do merge it?

antonte:
@leios thank you for pointing to .md file. I'll do it.

antonte:
- updated the corresponding .md file

jiegillet:
It's not the world most divisive question, but I think I'm with @leios in this case. Our goal is to explain algorithms, That first line is super explicit and matches the text. 
Obviously in my own code it's way more obfuscated haha ^^

antonte:
I am fine if you reject the PR, especially if the PR makes the code inconsistent with the text in the book.

jiegillet:
Not inconsistent, your modification is fine, just, as leios said, 2% less readable for the purpose of writing a book.

Issue #257, opened 2018-07-16
Improving implementations for Verlet Integration


Butt4cak3:
## 1. Printing inside the individual functions vs. printing in main
Some of the code examples print the results directly inside the `verlet` function. Others have the functions return the results and print them in `main`.

With #243, we're already steering in the direction of printing in `main` instead of the individual functions. To keep everything in line, we should do this in every example.

## 2. Using the velocity variable in Stormer-Verlet
Currently, the chapter about Verlet Integration includes code examples for the Stormer-Verlet and Velocity-Verlet methods. Both of them have a velocity term, but it's completely unused in Stormer-Verlet. The variable is assigned a value which is never actually used. In #256, @leios and I already agreed that it would be an improvement to also print the velocity. I want to leave this up for discussion before we decide on it.

I think that if we print the velocity for Stormer-Verlet, we should also print it for Velocity-Verlet.

## Progress
If we were to change these 2 things, these are the languages examples that would need adjustment:

- [x] C (#298)
- [x] C++ (both)
- [x] Java (#290)
- [x] JavaScript (print velocity)
- [x] Julia (both)
- [x] Python (#256)
- [x] Rust (#297)

Feel free to post your opinion on the second topic and to submit PRs for the listed languages. I will keep this list updated as the code comes in.

berquist:
For point 2, it should be more clear which form of the velocity equation is being calculated, the one at `t` or the one at `t + dt`.

berquist:
According to both the book and [Wikipedia](https://en.wikipedia.org/wiki/Verlet_integration#Computing_velocities_%E2%80%93_St%C3%B6rmer%E2%80%93Verlet_method), there is the form where the velocity is calculated one step behind (`t`) with error squared in `t`, and the form where velocity is calculated at the same step (`t + dt`) with error linear in `t`. Because there are multiple choices for defining the velocity, it should be stated which one is used in the code. I suppose it's clear from "Here's what it looks like in code", but the divergence in how this is presented in the two articles was temporarily confusing.

leios:
It appears that this has been resolved already!

PR #290, opened 2018-07-22
Implementation Edit to Verlet Integration in Java

CDsigma:
- Updated the code to use the new standard (#257) 
- Put code in a class so that it can be run as is


Butt4cak3:
How about renaming *verlet.java* to *Verlet.java* while you're at it? That way one can actually run it as-is.

CDsigma:
Nice catch, just made that change.

contents/verlet_integration/code/java/Verlet.java
```diff
@@ -0,0 +1,77 @@
...
+            time += dt;
+            temp_pos = pos;
+            pos = pos*2 - prev_pos + acc * dt * dt;
+            prev_pos = temp_pos;
+        }
+
+        return time;
+    }
+  
+    static double[] stormer_verlet(double pos, double acc, double dt) {
```
Gustorn: I would argue that returning a named class here would be much more elegant

CDsigma: I appreciate the suggestion and I agree. I just made a commit to return a class instead of double[]

CDsigma:
Hi, I just wanted to check in on the state of this PR :)

julianschacher:
Hmm, apparently the PR is approved, but not merged. Is there a reason for that @Gustorn ?

Gathros:
I've checked it and it looks fine.

PR #291, opened 2018-07-22
Implementation Edit to Bogo Sort in Javascript

CDsigma:
Added main() to code so that it can be run as is

### DIFF
--- a/contents/bogo_sort/code/javascript/bogo.js
+++ b/contents/bogo_sort/code/javascript/bogo.js
@@ -24,7 +24,7 @@ function shuffle(arr) {
 }
 
 function main() {
-  let testArray = [4, 5, 123, 24, 34, -5];
+  const testArray = [4, 5, 123, 24, 34, -5];
   bogoSort(testArray);
   console.log(testArray);
 }
--- a/contents/bogo_sort/code/javascript/bogo.js
+++ b/contents/bogo_sort/code/javascript/bogo.js
@@ -12,7 +12,6 @@ function bogoSort(arr) {
   while (!isSorted(arr)) {
     shuffle(arr);
   }
-  return arr;
 }
 
 function shuffle(arr) {
@@ -25,8 +24,9 @@ function shuffle(arr) {
 }
 
 function main() {
-  var testArray = [4,5,123,24,34,-5];
-  print(bogoSort(testArray));
+  let testArray = [4, 5, 123, 24, 34, -5];
+  bogoSort(testArray);
+  console.log(testArray);
 }
 
 main();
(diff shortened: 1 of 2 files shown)


contents/bogo_sort/code/javascript/bogo.js
```diff
@@ -12,6 +12,7 @@ function bogoSort(arr) {
   while (!isSorted(arr)) {
     shuffle(arr);
   }
+  return arr;
```
Butt4cak3: The function doesn't have to return the array, because it's passed by reference anyway.

contents/bogo_sort/code/javascript/bogo.js
```diff
@@ -22,3 +23,10 @@ function shuffle(arr) {
     arr[r] = tmp;
   }
 }
+
+function main() {
+  var testArray = [4,5,123,24,34,-5];
+  print(bogoSort(testArray));
```
Butt4cak3: `print` does not exist in JavaScript. And because you don't need `return arr` above, you'd have to output it like this:

```JavaScript
bogoSort(testArray);
console.log(testArray);
```

contents/bogo_sort/code/javascript/bogo.js
```diff
@@ -25,8 +24,9 @@ function shuffle(arr) {
 }
 
 function main() {
-  var testArray = [4,5,123,24,34,-5];
-  print(bogoSort(testArray));
+  let testArray = [4, 5, 123, 24, 34, -5];
```
Gustorn: This doesn't change, it should be `const`. Same in the other PR

CDsigma: Oops, sorry about that. I made a commit to correct the code.

julianschacher:
@Butt4cak3 @Gustorn How is the state of this PR?

contents/bogo_sort/code/javascript/bogo.js
```diff
@@ -22,3 +22,11 @@ function shuffle(arr) {
     arr[r] = tmp;
   }
 }
+
+function main() {
+  const testArray = [4, 5, 123, 24, 34, -5];
```
jiegillet: I don't know JS too well, but if you declare `testArray` and `const` can you actually modify it?

contents/bogo_sort/code/javascript/bogo.js
jiegillet: Why does this not change? It's being sorted later.

contents/bogo_sort/code/javascript/bogo.js
MichalHanajik: yes, you can not reassign this. but you can add/remove/reorder elements in array like this.

PR #184, opened 2018-06-30
Py monte carlo

GuyPozner:
I've review the current pull request for the python implementation and made change to resemble the current c and julia versions. Hope that it helps.

PR #294, opened 2018-07-23
Add Python code to Monte Carlo Integration

hybrideagle:


contents/monte_carlo_integration/code/python/monte_carlo.py
```diff
@@ -0,0 +1,35 @@
+# submitted by hybrideagle
+from random import random
+from math import pi
+
+
+def inCircle(xPos, yPos):
```
Butt4cak3: Functions and variables are usually named using snake_case rather than camelCase in Python. For example,. this line should rather be:

```Python
def in_circle(x_pos, y_pos):
```

This goes for everything in this file.

Butt4cak3:
I totally forgot this, but we currently have multiple Python Monte Carlo pull requests. I can't guarantee that your code will actually be used because other people were faster than you. I'll ask @jiegillet about the situation as soon as he's back since he's handling the situation as far as I know.

jiegillet:
Hi @hybrideagle, thanks for you contribution. #184 is ahead of you. Feel free to look at my review of that code and change your code accordingly if you wish, but I will only accept your code if that other PR goes dead. That has happened 2 or 3 times before for monte carlo in Python, somehow this combination is cursed.

hybrideagle:
This is the weirdest algorithm to have this kind of churn. 
Should I close this PR for now?

jiegillet:
I think you can close it. We are close to resolution with the other PR.

PR #299, opened 2018-07-24
Update C++ implementations for Verlet algorithms

berquist:
Done according to https://github.com/algorithm-archivists/algorithm-archive/issues/257.

### DIFF
--- a/contents/verlet_integration/code/c++/verlet.cpp
+++ b/contents/verlet_integration/code/c++/verlet.cpp
@@ -1,8 +1,10 @@
 #include <iomanip>
 #include <iostream>
-#include <utility>
 
-using vpair = std::pair<double, double>;
+struct timestep {
+  double time;
+  double vel;
+};
 
 double verlet(double pos, double acc, double dt) {
 
@@ -19,7 +21,7 @@ double verlet(double pos, double acc, double dt) {
   return time;
 }
 
-vpair stormer_verlet(double pos, double acc, double dt) {
+timestep stormer_verlet(double pos, double acc, double dt) {
 
   double prev_pos = pos;
   double time = 0;
@@ -35,10 +37,10 @@ vpair stormer_verlet(double pos, double acc, double dt) {
     vel += acc * dt;
   }
 
-  return std::make_pair(time, vel);
+  return timestep { time, vel };
 }
 
-vpair velocity_verlet(double pos, double acc, double dt) {
+timestep velocity_verlet(double pos, double acc, double dt) {
 
   double time = 0;
   double vel = 0;
@@ -48,12 +50,10 @@ vpair velocity_verlet(double pos, double acc, double dt) {
     vel += acc * dt;
   }
 
-  return std::make_pair(time, vel);
+  return timestep { time, vel };
 }
 
 int main() {
-  double time, vel;
-
   std::cout << std::fixed << std::setprecision(8);
 
   // Note that depending on the simulation, you might want to have the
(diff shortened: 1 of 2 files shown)


contents/verlet_integration/code/c++/verlet.cpp
```diff
@@ -45,18 +48,41 @@ double velocity_verlet(double pos, double acc, double dt) {
     vel += acc * dt;
   }
 
-  return time;
+  return std::make_pair(time, vel);
 }
 
 int main() {
+  double time, vel;
```
Gathros: Why not just print straight to console.

contents/verlet_integration/code/c++/verlet.cpp
```diff
@@ -1,6 +1,9 @@
+#include <iomanip>
 #include <iostream>
+#include <utility>
+
+typedef std::pair<double, double> vpair;
```
antonte:
``` C++
using vpair = std::pair<double, double>;
```
is preferable way.

[T.43: Prefer using over typedef for defining](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#t43-prefer-using-over-typedef-for-defining-aliases)

contents/verlet_integration/code/c++/verlet.cpp
antonte: [ES.20: Always initialize an object](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es20-always-initialize-an-object)

contents/verlet_integration/code/c++/verlet.cpp
```diff
@@ -45,18 +48,41 @@ double velocity_verlet(double pos, double acc, double dt) {
...
+  time = time_vel_pair.first;
+  vel = time_vel_pair.second;
+  std::cout << "Time for Stormer Verlet integration is: " \
+            << time << std::endl;
+  std::cout << "Velocity for Stormer Verlet integration is: " \
+            << vel << std::endl;
 
-  // Note that depending on the simulation, you might want to have the verlet
-  // loop outside.
+  time_vel_pair = velocity_verlet(5.0, -10, 0.01);
```
antonte: you can use [```std::tie```](https://en.cppreference.com/w/cpp/utility/tuple/tie)

contents/verlet_integration/code/c++/verlet.cpp
```diff
@@ -16,8 +19,7 @@ double verlet(double pos, double acc, double dt) {
   return time;
 }
 
-// Simple function for stormer-verlet
-double stormer_verlet(double pos, double acc, double dt) {
+std::pair<double, double> stormer_verlet(double pos, double acc, double dt) {
```
antonte: use ```vpair```

contents/verlet_integration/code/c++/verlet.cpp
```diff
@@ -28,14 +30,15 @@ double stormer_verlet(double pos, double acc, double dt) {
...
+    // straightforward
     vel += acc * dt;
   }
 
-  return time;
+  return std::make_pair(time, vel);
 }
 
-double velocity_verlet(double pos, double acc, double dt) {
+std::pair<double, double> velocity_verlet(double pos, double acc, double dt) {
```
antonte: use ```vpair```

contents/verlet_integration/code/c++/verlet.cpp
Gustorn: I think I would prefer a simple struct with descriptive fields over the pair anyway. `std::pair` is alright for utility types but I dislike it for types that should have some semantic meaning.

contents/verlet_integration/code/c++/verlet.cpp
```diff
@@ -45,18 +50,35 @@ double velocity_verlet(double pos, double acc, double dt) {
...
+  // For example, if your acceleration chages as a function of time,
+  // you might need to also change the acceleration to be read into
+  // each of these functions.
+
+  double time = verlet(5.0, -10, 0.01);
+  std::cout << "Time for Verlet integration is: " \
+            << time << std::endl;
+
+  timestep timestep_sv = stormer_verlet(5.0, -10, 0.01);
+  std::cout << "Time for Stormer Verlet integration is: " \
```
Gathros: Why is there a backslash?

berquist: I prefer to use an explicit line continuation indicator even when not required.

PR #308, opened 2018-07-28
Adding split_op.c and energy.c

Gathros:


contents/split-operator_method/code/c/split_op.c
```diff
@@ -0,0 +1,210 @@
...
+#include <string.h>
+#include <math.h>
+
+#include <fftw3.h>
+
+struct params {
+    double xmax;
+    int res;
+    double dt;
+    int timesteps;
```
berquist: It makes more physical sense for this to be `unsigned`.

contents/split-operator_method/code/c/split_op.c
```diff
@@ -0,0 +1,210 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <math.h>
+
+#include <fftw3.h>
+
+struct params {
+    double xmax;
+    int res;
```
berquist: It doesn't make physical sense for the resolution to be signed.

leios:
The implementation seems fine here; however, because we are using the fftw library, would you mind adding in a comment at the top of the file with compilation instructions (or maybe even in the text above the codeblock, but within the C-language method).

PR #239, opened 2018-07-09
added monte carlo method c#

sanstorik:
Simple monte carlo method implementation in c#

PR #309, opened 2018-07-28
Add Monte Carlo Integration in C#

julianschacher:
Since [this PR](https://github.com/algorithm-archivists/algorithm-archive/pull/239) is dead, I wrote it myself.

- [x] Use`Math.Pow` instead of `point.X * point.X`.
- [x] Make the error a percentage.
- [x] Use `10,000,000` instead of `10.000.000` for the console output.
- [ ] Calculate the percent error in Main.

### DIFF
--- a/contents/monte_carlo_integration/code/csharp/Circle.cs
+++ b/contents/monte_carlo_integration/code/csharp/Circle.cs
@@ -1,0 +1,25 @@
+using System;
+
+namespace MonteCarloIntegration
+{
+    public struct Point
+    {
+        public double X { get; set; }
+        public double Y { get; set; }
+
+        public Point(double x, double y)
+        {
+            this.X = x;
+            this.Y = y;
+        }
+    }
+
+    public class Circle
+    {
+        public double Radius { get; private set; }
+
+        public Circle(double radius) => this.Radius = Math.Abs(radius);
+
+        public bool IsInMe(Point point) => point.X * point.X + point.Y * point.Y < Radius * Radius;
+    }
+}
--- a/contents/monte_carlo_integration/code/csharp/Circle.cs
+++ b/contents/monte_carlo_integration/code/csharp/Circle.cs
@@ -20,6 +20,6 @@ namespace MonteCarloIntegration
 
         public Circle(double radius) => this.Radius = Math.Abs(radius);
 
-        public bool IsInMe(Point point) => point.X * point.X + point.Y * point.Y < Radius * Radius;
+        public bool IsInMe(Point point) => Math.Pow(point.X, 2) + Math.Pow(point.Y, 2) < Math.Pow(Radius, 2);
     }
 }
--- a/contents/monte_carlo_integration/code/csharp/MonteCarlo.cs
+++ b/contents/monte_carlo_integration/code/csharp/MonteCarlo.cs
@@ -1,0 +1,37 @@
+using System;
+
+namespace MonteCarloIntegration
+{
+    public class MonteCarlo
+    {
+        public struct RunOutput
+        {
+            public double PiEstimate { get; private set; }
+            public double Error { get; private set; }
+
+            public RunOutput(double piEstimate, double error)
+            {
+                this.PiEstimate = piEstimate;
+                this.Error = error;
+            }
+        }
+
+        public RunOutput Run(int samples)
+        {
+            var circle = new Circle(1.0);
+            var count = 0;
+            var random = new Random();
+
+            for (int i = 0; i < samples; i++)
+            {
+                var point = new Point(random.NextDouble(), random.NextDouble());
+                if (circle.IsInMe(point))
+                    count++;
+            }
+
+            var piEstimate = 4.0 * count / samples;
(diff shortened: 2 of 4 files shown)


contents/monte_carlo_integration/code/csharp/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        static void Main(string[] args)
+        {
+            var monteCarlo = new MonteCarlo();
+            System.Console.WriteLine("Running with 10.000.000 samples.");
+            var output = monteCarlo.Run(10000000);
+            System.Console.WriteLine($"The estimate of pi is: {output.PiEstimate}");
+            System.Console.WriteLine($"The error is: {output.Error}");
+
+            System.Console.WriteLine("How many samples should be used?");
```
julianschacher: Should I leave this custom user input in?

contents/monte_carlo_integration/code/csharp/Circle.cs
```diff
@@ -0,0 +1,25 @@
...
+        }
+    }
+
+    public class Circle
+    {
+        public double Radius { get; private set; }
+
+        public Circle(double radius) => this.Radius = Math.Abs(radius);
+
+        public bool IsInMe(Point point) => point.X * point.X + point.Y * point.Y < Radius * Radius;
```
julianschacher: I used `point.X * point.X` now instead of using `Math.Pow`. It doesn't really matter here, what do you find cleaner?

contents/monte_carlo_integration/code/csharp/MonteCarlo.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            for (int i = 0; i < samples; i++)
+            {
+                var point = new Point(random.NextDouble(), random.NextDouble());
+                if (circle.IsInMe(point))
+                    count++;
+            }
+
+            var piEstimate = 4.0 * count / samples;
+            var error = Math.Abs(Math.PI - piEstimate);
```
jiegillet: Make the error a percentage (and add % when you print out the error) to remove ambiguity

contents/monte_carlo_integration/code/csharp/Program.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+
+namespace MonteCarloIntegration
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            var monteCarlo = new MonteCarlo();
+            System.Console.WriteLine("Running with 10.000.000 samples.");
```
jiegillet: maybe it should be `10,000,000`?

contents/monte_carlo_integration/code/csharp/Program.cs
```diff
@@ -0,0 +1,26 @@
...
+            var piEstimate = monteCarlo.Run(10000000);
+            System.Console.WriteLine($"The estimate of pi is: {piEstimate.Estimate}");
+            System.Console.WriteLine($"The error is: {piEstimate.Error}");
+
+            System.Console.WriteLine("How many samples should be used?");
```
MichalHanajik: I personally would not include this input part's to code examples. I would keep them minimal.

contents/monte_carlo_integration/code/csharp/Program.cs
julianschacher: You're right, just thought german there :D

contents/monte_carlo_integration/code/csharp/Program.cs
julianschacher: I personally like to give the user a choice here, but I understand that it makes things quite a bit more messy.

contents/monte_carlo_integration/code/csharp/Program.cs
MichalHanajik: use ``` 10 000 000``` then

julianschacher: I think 10,000,000 is fine since the example is in english.

contents/monte_carlo_integration/code/csharp/MonteCarlo.cs
```diff
@@ -0,0 +1,37 @@
...
+
+            for (int i = 0; i < samples; i++)
+            {
+                var point = new Point(random.NextDouble(), random.NextDouble());
+                if (circle.IsInMe(point))
+                    count++;
+            }
+
+            var piEstimate = 4.0 * count / samples;
+            var percentError = (piEstimate - Math.PI) / Math.PI * 100;
```
jiegillet: One last thing: it feels strange to me to use Math.PIinside of a function that estimates pi. I've usually recommended to other people to calculate the error in main.

julianschacher: Yeah, I see what you mean. I'm gonna change that.

jiegillet:
Whoa, +4 -18 for moving a single line. OOP really is verbose :)

julianschacher:
@jiegillet I didn't need that output class anymore :D

PR #312, opened 2018-07-28
Added mathjax reference on algorithm-archivists/plugin-mathjax

MichalHanajik:
This is applicable if

- [x] https://github.com/algorithm-archivists/plugin-mathjax/pull/1 is merged
- [ ] Check locally if dependency on updated plugin is correctly download in ```gitbook install```

I set version which is currently as fallback **2.6.1** but it's possible use any version, just changing **book.json** for future :)

### DIFF
--- a/book.json
+++ b/book.json
@@ -2,7 +2,7 @@
   "gitbook": "3.x.x",
   "plugins": [
     "fontsettings",
-    "mathjax",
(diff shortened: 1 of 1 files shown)


MichalHanajik:
It was just someone else roo test locally if the plugin is correctly downloading from master branch after PR in his repo.

PR #316, opened 2018-07-29
Use Travis CI to test the book build and then deploy it to the site, take 2

berquist:
Continuation of https://github.com/algorithm-archivists/algorithm-archive/pull/284.

Don't merge this yet, because things need to be added.

The last PR failed in three ways:

1. The site went down because the CNAME for custom domain redirection was not present. After checking the log ([which I saved](https://github.com/algorithm-archivists/algorithm-archive/files/2239118/409399060.txt)), if you're curious, I noticed that there are several other files present in algorithm-archivists.github.io that do _not_ come from either the main repo or building the book. These can persist because `update_site.sh` only copies files over, rather than wipe and move. This is personal preference, but I think the website repo should be entirely reproducible from the main repo via this script. You can see the difference by comparing [your website repo](https://github.com/algorithm-archivists/algorithm-archivists.github.io) with [mine](https://github.com/berquist/algorithm_archive_build). Let me know which path you'd rather take, and I'll either need to add those files to the main repo and have them copy over during deploy, or have the deploy copy the build rather than wipe the book repo.

2. The guard for when the book should be deployed was wrong. Once the initial PR was merged, every Travis `pull_request` build (see below) was going to try deploy. This is before the merge actually happens. I don't think Travis would continue to this step if the build failed earlier, but it's still wrong.

3. A bunch of files got deleted. I still don't know how this happened, because I think they should have been present in the book build. I don't think they were extra files left over from before the flattening, either.

---

For reference, and because I didn't understand it for a while, there are two kinds of Travis builds:

1. `push`:
    - Under the status dialog, this is called `Travis CI - Branch` and `continuous-integration/travis-ci/push`.
    - "This is a normal build for the <branch-name> branch. You should be able to reproduce it by checking out the branch locally."
    - What it does is test the branch as-is. By default, this will run for _everyone_ who has a fork of the main repo every time they push to their own fork.

2. `pull_request`:
    - Under the status dialog, this is called `Travis CI - Pull Request` and `continuous-integration/travis-ci/pr`.
    - "This is a pull request build. It is running a build against the merge commit, after merging <pull-request-name>. Any changes that have been made to the master branch before the build ran are also included."
    - What is does is it simulates merge specified by the PR, then runs the Travis config. This only runs after creating a pull request.

The reason each type of CI run appears twice is that the former leads to the pretty GitHub summary, and the latter leads to the actual Travis log.

The config is now set up so that only `push` commits to master will deploy to algorithm-archivists.github.io. I checked, and this works in all cases: when there's a merge commit, a rebase and FF merge, a squash and merge, or a regular commit. _Any_ time the history of master is updated, the book will be built and deployed.

---

Note that I changed the names of some environment variables you must set in Travis to be more explicit. Hopefully things are clearer now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+dist: trusty
+sudo: false
+
+language: node_js
+node_js:
+  - "4"
+
+cache:
+  directories:
+    - node_modules
+
+install:
+  - npm install gitbook-cli -g
+  - gitbook install
+
+before_script:
+  - mkdir -p "${TRAVIS_BUILD_DIR}"/build
+
+script:
+  - gitbook build . "${TRAVIS_BUILD_DIR}"/build
+  - env | sort
+  - |
+    if [[ "${TRAVIS_BRANCH}" == master && "${TRAVIS_PULL_REQUEST}" == false ]]; then
+        # Commits to master that are not pull requests, that is, only
+        # actual addition of code to master, should deploy the book to
+        # the site.
+        bash "${TRAVIS_BUILD_DIR}"/tools/deploy/update_site_travis.bash
+    fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ language: node_js
 node_js:
   - "4"
 
-cache:
-  directories:
-    - node_modules
+# cache:
+#   directories:
+#     - node_modules
 
 install:
   - npm install gitbook-cli -g
--- a/tools/deploy/update_site_travis.bash
+++ b/tools/deploy/update_site_travis.bash
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+# update_site_travis.bash: Update a GitHub repo with the book build
+# products.
+
+set -o errexit
+
+# Required environment variables:
+# - DOCS_BRANCH_NAME: name of the remote branch serving the book
+# - DOCS_REPO_NAME: name of the remote repo
+# - DOCS_REPO_OWNER: name of the remote repo's owner
+# - GH_TOKEN: [secret] Personal Access Token
+
+bold=$(tput bold)
+normal=$(tput sgr0)
+
+if [ -z ${DOCS_BRANCH_NAME+x} ]; then
+    echo "${bold}\$DOCS_BRANCH_NAME is not set!${normal}"
+    exit 1
+elif [ -z ${DOCS_REPO_NAME+x} ]; then
+    echo "${bold}\$DOCS_REPO_NAME is not set!${normal}"
+    exit 1
+elif [ -z ${DOCS_REPO_OWNER+x} ]; then
+    echo "${bold}\$DOCS_REPO_OWNER is not set!${normal}"
+    exit 1
+elif [ -z ${GH_TOKEN+x} ]; then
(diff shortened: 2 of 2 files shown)


berquist:
Now that Travis has been activated, you can see that the `pr`/`pull_request` build didn't deploy to the book, which would have broken the site...again...

leios:
Ok. I think we are ready to give this another go, but I would like for @berquist to be available when we merge this PR just in case... So maybe we should set up a time to give this a shot on discord?

berquist:
Sure. I won't be around Monday EST, but any time from Tuesday EST on is fine.

In the meantime, read the first paragraph again. This PR needs to be updated before merging, otherwise the site will break.

leios:
Mind if I add a WIP to the title to reflect this?

berquist:
Go right ahead.

Just to be clear, what changes I make next depend on your decision about
point 1.

leios:
So to be clear: you are arguing there are a lot of extra files because...
> update_site.sh only copies files over, rather than wipe and move. This is personal preference, but I think the website repo should be entirely reproducible from the main repo via this script.

I'm up for updating the update_site command. What do we need to do with the CNAME stuff?

berquist:
Uncommenting [this line](https://github.com/algorithm-archivists/algorithm-archive/pull/316/files#diff-cce3fb257dfc4f7fce163092094a083bR45) will take care of CNAME. I still need time to check the rest of the files.

However, there is a new problem: https://travis-ci.org/algorithm-archivists/algorithm-archive/builds/409608474#L427. If you go to https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions:

> Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.

> Pull requests sent from forked repositories do not have access to encrypted variables or data.

> If your build relies on encrypted variables to run, for instance to run Selenium tests with BrowserStack or Sauce Labs, your build needs to take this into account. You won’t be able to run these tests for pull requests from external contributors.

> To work around this, restrict these tests only to situations where the environment variables are available, or disable them for pull requests entirely, as shown in the following example:

I need to figure out if the AAA will be affected by this. I think so, because I am on another org's repo, but for some reason this PR ran fine? It will take some investigation. Maybe later this week is best.

leios:
Hey @berquist, quick question on the state of this PR. Obviously, no pressure to finish, but I just wanted to make sure it was still on the table.

leios:
Of course no rush, just checking in

berquist:
Can you do me a favor and try restarting Travis? I don't understand the sudden failure. https://travis-ci.org/algorithm-archivists/algorithm-archive/builds/432252148

leios:
I'll be honest, I have no idea how to do that.

berquist:
Is there a button labeled "Restart job" or something similar in the area where I have whitespace on the right?

screen shot 2018-09-23 at 7 47 07 pm

leios:
No, I see what you see

julianschacher:
@leios are you signed in to Travis?

leios:
Yeah, alright. That was it. I logged in and could restart it. Sorry.

PR #317, opened 2018-07-30
Adding Tree Traversal in Swift

CDsigma:
Added Tree Traversal in Swift


contents/tree_traversal/code/swift/tree.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    print(node.value)
+}
+
+func dfsRecursiveInOrderBinary(node: Node) {
+    if node.children?.count == 2 {
+        dfsRecursiveInOrderBinary(node: node.children![0])
+        dfsRecursiveInOrderBinary(node: node.children![1])
+        print(node.value)
```
jiegillet: For in order, `print(node.value)` has to come in between the children.

PR #321, opened 2018-08-01
Added Thomas Algorithm in Haskell

jiegillet:
Also corrects minor typo in the chapter.

That's a short one. @lulucca12 I think this one will be more to your liking.

@leios that chapter is a little weak on the explanation of the algorithm. In particular, I don't get "In particular, we need to find an optimal scale factor for each row and use that.". Also you write down the formula for the forward step, but you don't explicit the back substitution (sure, it's similar to the gaussian elimination but the result is so simple...). I know that you have many things to do. Would you be OK with me fleshing it out a bit? (In another PR this time!!)

### DIFF
--- a/contents/thomas_algorithm/code/haskell/thomas.hs
+++ b/contents/thomas_algorithm/code/haskell/thomas.hs
@@ -1,0 +1,18 @@
+import Data.List (zip4)
+import Data.Ratio
+
+thomas :: Fractional a => [a] -> [a] -> [a] -> [a] -> [a]
+thomas a b c = init . scanr back 0 . tail . scanl forward (0, 0) . zip4 a b c
+  where
+    forward (c', d') (a, b, c, d) =
+      let denominator = b - a * c'
+       in (c / denominator, (d - a * d') / denominator)
+    back (c, d) x = d - c * x
+
+main :: IO ()
+main = do
+  let a = [0, 2, 3] :: [Ratio Int]
+      b = [1, 3, 6]
+      c = [4, 5, 0]
+      d = [7, 5, 3]
+  print $ thomas a b c d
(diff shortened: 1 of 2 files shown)


leios:
@jiegillet yeah, please feel free to update the text here (in a separate PR). This one could definitely use some help.

lulucca12:
Couldn't you separate the forward and back functions? On the other languages, they are in different blocks of code. And why Data.Ratio, I do not really know what it does but it doesn't make sense for me.

jiegillet:
Data.Ratio is not part of the algorithm, it's just part of the test in `main`.  `Ratio Int` is a type that keeps track of fractional numbers with `Int`s and reduces them. For example if you compute `4/8` as a `Double`, you will get `0.5` but as `Ratio Int` it will be `1 % 2` (where the % is the fraction sign). It's really neat. This way, after the algorithm we get a result in fractions and not as real numbers. But it's not necessary, if you ommit the `:: [Ration Int]` everything will be treated as `Double`. Polymorphism for the win :)

`forward` and `back` are not separated? They are defined separately inside of`thomas`. They don't really make sense outside of that function.

lulucca12:
Cool way to represent fractions, I didn't think of that. And in other programming languages forward and back are different 'for loops' for example and here they are all bounded to the same function. But I think it makes more sense on the way that you did. I've seen that the functions that combine 
```haskell
tail . scanl
```
and
```haskell
init . scanr
```
don't exist like

```haskell
let scanl2 f xs = tail $ scanl f 0 xs
let scanr2 f xs = init $ scanr f 0 xs
```
in Data.List because they could shorten your code.

I wouldn't like defining them here, it would be unnecessary.

jiegillet:
I'm glad you like my code ^^
Could you add a review so we can merge it?

PR #322, opened 2018-08-01
Added Thomas Algorithm in Scratch

jiegillet:
Yay Scratch!

I've set the width of the svg to 1000, but I think it resizes down automatically if the screen size is smaller. Pretty neat.


jiegillet:
Yay! I'm super happy :D

And yes, Thomas algorithm is easy to implement cleanly when I don't have access to crazy abstractions :p

PR #323, opened 2018-08-02
Rename directory from "py" to "python" for the sake of consistency and complying to the guidelines.

julianschacher:


julianschacher:
@jiegillet yup it was :)

PR #326, opened 2018-08-03
Adding the Missing Rust Tree Traversal Functions

CDsigma:
Figured I could quickly add in the dfs_recursive_postorder function in Rust since it is so similar to the dfs_recursive.

The inorder binary tree function is still missing.

### DIFF
--- a/contents/tree_traversal/code/rust/tree.rs
+++ b/contents/tree_traversal/code/rust/tree.rs
@@ -14,6 +14,14 @@ fn dfs_recursive(n: &Node) {
     }
 }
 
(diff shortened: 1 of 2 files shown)


CDsigma:
I don't really know Rust. I figured I could just add in `dfs_recursive_postorder` because it is just swapping the order of `dfs_recursive`.

Since I don't really know Rust I don't feel super comfortable doing `dfs_recursive_inorder_btree`. I might in the future figure out Rust and do `dfs_recursive_inorder_btree` but I'm not sure when I'll have the time.

PR #336, opened 2018-08-05
Fix typo in Rust tree traversal

berquist:
Creating trees is also more consistent with the Julia implementation.

### DIFF
--- a/contents/tree_traversal/code/rust/tree.rs
+++ b/contents/tree_traversal/code/rust/tree.rs
@@ -16,7 +16,7 @@ fn dfs_recursive(n: &Node) {
 
 fn dfs_recursive_postorder(n: &Node) {
     for child in &n.children {
-        dfs_recursive(child);
+        dfs_recursive_postorder(child);
     }
 
     println!("{}", n.value);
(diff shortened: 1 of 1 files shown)


contents/tree_traversal/code/rust/tree.rs
```diff
@@ -69,15 +69,16 @@ fn create_tree(num_row: u64, num_child: u64) -> Node {
...
     dfs_stack(&root);
     println!("Queue BFS:");
     bfs_queue(&root);
-    println!("Recursive PostOrder DFS: ");
+    println!("Recursive post-order DFS:");
     dfs_recursive_postorder(&root);
-    println!("Recursive DFS BTree:");
-    dfs_recursive_inorder_btree(&root);
+    println!("Recursive in-order DFS BTree:");
+    let root_binary = create_tree(3, 2);
```
Gathros: Why not just have one tree.

jiegillet: If you want to test the in-order, you need a binary tree. The other tree is ternary.

Gathros: Just keep the other tree binary too.

jiegillet: The Julia code and chapter show examples with ternary, though. Not a big deal to me, but I'm not actually reviewing the code, just answering one question :)

Gathros: You're right it, should match the julia code.

PR #338, opened 2018-08-05
Adding Formatting to File Names

CDsigma:
Some programs use multiple files and have the title of the file above the code. Currently there is no formatting on these titles.

I added in '#####' before these titles which I think looks a little nicer.

This might just be a personal preference though so I understand if it isn't merged.

### DIFF
--- a/contents/bogo_sort/bogo_sort.md
+++ b/contents/bogo_sort/bogo_sort.md
@@ -49,9 +49,9 @@ We are done here!
 {% sample lang="jl" %}
 [import, lang:"julia"](code/julia/bogo.jl)
 {% sample lang="cs" %}
-BogoSort.cs
+##### BogoSort.cs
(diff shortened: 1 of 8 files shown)


leios:
I agree with this. It's a bit nicer. I'll leave it up for a bit so people can see it, though.

jiegillet:
I don't mind.

jiegillet:
Are these changes exhaustive?

leios:
I think so, but haven't checked thoroughly yet. There are not too many programming languages that require multiple files for their implementations.

jiegillet:
I'm thinking of blocking this PR until we come up with 

1. A Script that finds and replace such occurrences (bash or Python, I guess)
2. A place in the AAA where we describe the new format.

Does it sound OK? I think we need both of these before we merge this.

julianschacher:
I'm in favor of these changes. The filenames look better in bold text.

PR #342, opened 2018-08-06
Implement C++ version of Monte Carlo integration

bsamseth:
I just came across this project and wanted to chip in something. Hope this conforms to any style preferences. 👍

### DIFF
--- a/contents/monte_carlo_integration/code/c++/monte_carlo.cpp
+++ b/contents/monte_carlo_integration/code/c++/monte_carlo.cpp
@@ -22,7 +22,7 @@ inline bool in_circle(double x, double y, double r = 1) {
  */
 double monte_carlo_pi(unsigned samples) {
     static std::default_random_engine generator;
-    static std::uniform_real_distribution<double> dist(-1, 1);
+    static std::uniform_real_distribution<double> dist(0, 1);
 
     unsigned count = 0;
     for (unsigned i = 0; i < samples; ++i) {
@@ -33,7 +33,7 @@ double monte_carlo_pi(unsigned samples) {
             ++count;
     }
 
-    return 4.0 * count / (double) samples;
+    return 4.0 * count / samples;
 }
 
 int main() {
@@ -42,7 +42,7 @@ int main() {
     std::cout << "Enter samples to use: ";
     std::cin >> samples;
 
-    double pi = monte_carlo_pi(samples);
-    printf("Pi = %f\n", pi);
-    printf("Percent error is: %g %%\n", 100 * std::abs(pi - PI) / PI);
+    double pi_estimate = monte_carlo_pi(samples);
+    std::cout << "Pi = " << pi_estimate << '\n';
+    std::cout << "Percent error is: " << 100 * std::abs(pi_estimate - PI) / PI << " %\n";
 }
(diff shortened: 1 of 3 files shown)


contents/monte_carlo_integration/code/c++/monte_carlo.cpp
```diff
@@ -0,0 +1,48 @@
...
+    unsigned count = 0;
+    for (unsigned i = 0; i < samples; ++i) {
+        double x = dist(generator);
+        double y = dist(generator);
+
+        if (x*x + y*y < 1)
+            ++count;
+    }
+
+    return 4.0 * count / (double) samples;
```
berquist: This cast isn't necessary.

contents/monte_carlo_integration/code/c++/monte_carlo.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+int main() {
+    unsigned samples;
+
+    std::cout << "Enter samples to use: ";
+    std::cin >> samples;
+
+    double pi = monte_carlo_pi(samples);
```
berquist: Could you name this something else? The difference between lowercase and uppercase pi isn't enough to differentiate where they come from.

contents/monte_carlo_integration/code/c++/monte_carlo.cpp
```diff
@@ -0,0 +1,48 @@
...
+    printf("Pi = %f\n", pi);
```
berquist: Can you use `std::cout` here and the next line?

contents/monte_carlo_integration/code/c++/monte_carlo.cpp
```diff
@@ -0,0 +1,48 @@
...
+}
+
+/**
+ * Return an estimate of PI using Monte Carlo integration.
+ * @param samples number of iterations to use
+ * @return estimate of pi
+ */
+double monte_carlo_pi(unsigned samples) {
+    static std::default_random_engine generator;
+    static std::uniform_real_distribution<double> dist(-1, 1);
```
jiegillet: All the other implementation use a distribution should be between 0 and 1 since we are only looking into a corner. It doesn't fundamentally change anything since you only ever consider `x*x` anyway, but it would be more consistent with the chapter.

contents/monte_carlo_integration/code/c++/monte_carlo.cpp
```diff
@@ -0,0 +1,48 @@
...
+double monte_carlo_pi(unsigned samples) {
+    static std::default_random_engine generator;
+    static std::uniform_real_distribution<double> dist(0, 1);
+
+    unsigned count = 0;
+    for (unsigned i = 0; i < samples; ++i) {
+        double x = dist(generator);
+        double y = dist(generator);
+
+        if (x*x + y*y < 1)
```
jiegillet: Wait, I just realized you are not using the `in_circle ` function. That defeats the purpose of defining it :)

bsamseth: Your right! My bad, fixed this now!

jiegillet: Super quick!

PR #345, opened 2018-08-07
Implement Java version of the stable marriage problem.

bsamseth:


contents/stable_marriage_problem/code/java/stable-marriage.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.Collections;
+
+class StableMarriage {
+
+    /*
+     * Use the stable marriage algorithm to find stable pairs from the
+     * lists of men and women.
+     */
+    public static void findStableMarriages(List<Woman> women, List<Man> men) {
+        boolean keep_going;
```
Gustorn: Going by the Java style guide this should be `keepGoing`

contents/stable_marriage_problem/code/java/stable-marriage.java
```diff
@@ -0,0 +1,146 @@
...
+                    man.propose();
+
+            // The women pick their favorite suitor.
+            for (Woman woman : women)
+                woman.chooseMate();
+
+            // Repeat the process if someone is left without a partner.
+            keep_going = false;
+            for (Person person : leastCommonGender) {
+                if (keep_going |= person.isLonely()) {
```
Gustorn: I really don't like this. First, it uses the less common variant of the OR operator (non short-circuiting). Second, it checks the result of an assignemnt which is probably even worse. You could do this instead:

```java
keepGoing = leastCommonGender.stream().anyMatch(Person::isLonely)
```

Or alternatively, get rid of `keepGoing` altogether:

```java
if (!leastCommonGender.anyMatch(Person::isLonely)) {
    break;
}
```

contents/stable_marriage_problem/code/java/stable-marriage.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+        } while (keep_going);
+
+        women.forEach(w -> System.out.println(w + " married to " + w.getMate()));
+    }
+
+    public static void main(String[] args) {
+        int nPairs = 5;
+        List<Woman> women = new LinkedList<>();
```
Gustorn: You should really use `ArrayList` in both of these cases. `LinkedList` is just a bad data structure in general

contents/stable_marriage_problem/code/java/stable-marriage.java
```diff
@@ -0,0 +1,146 @@
...
+
+        findStableMarriages(women, men);
+    }
+
+}
+
+class Person {
+    private final String name;
+    protected Person mate;
+    protected List<Person> preferedMates;
```
Gustorn: prefered -> preferred

contents/stable_marriage_problem/code/java/stable-marriage.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    public Person getMate() {
+        return mate;
+    }
+
+    public void receiveOptions(List<? extends Person> mates) {
+        // Preferences are subjective.
+        preferedMates = new LinkedList<>(mates);
```
Gustorn: Similarly, use an ArrayList here

contents/stable_marriage_problem/code/java/stable-marriage.java
```diff
@@ -0,0 +1,146 @@
...
+        return preferedMates;
+    }
+
+    public String toString() {
+        return getClass().getName() + "(" + name + ")";
+    }
+}
+
+class Woman extends Person {
+    private List<Man> suitors = new LinkedList<>();
```
Gustorn: Use `ArrayList`

contents/stable_marriage_problem/code/java/stable-marriage.java
bsamseth: I very much agree, that is much cleaner!

contents/stable_marriage_problem/code/java/stable-marriage.java
bsamseth: The reasoning was that removal from the front of the list would be efficient. But I agree, `ArrayList` is probably at least just as good.

bsamseth:
@Gustorn Have you had time to look over the changes? Just realized this has been sitting untouched for some weeks now.

Butt4cak3:
I reviewed the changes you made and found everything like @Gustorn requested it. I also went ahead and checked that everything displays properly on the website. I'll go ahead and merge this.

PR #347, opened 2018-08-09
stack and queue in c

c252:
I implemented a stack and queue in c, James told me it was okay even though the section isn't open for code contribution.


leios:
I'll be updating this section this weekend (hopefully), so we'll hold off on merging this until that is done.

binarydinosaur252:
sounds good.

binarydinosaur252:
I will put in some error handling for the size. One problem is that I don't think you can have dynamically sized arrays in c. I think my queue implementation could also be more efficient though.

(should I close this PR, commit the changes and then make a new one)

Gathros:
No, you can just commit to the same branch and the PR updates. Also you use `malloc` and resize it with `realloc` to make a dynamically sized array in c.

Gustorn:
Of course you can have dynamically sized arrays in C, it wouldn't be a very useful language otherwise :) You just need to do it by hand with `malloc`

PR #348, opened 2018-08-09
Added Lisp implementation for bubble sort

Trashtalk217:
Added my name twice to CONTRIBUTORS.md but I have no clue how to fix it.

### DIFF
--- a/contents/bubble_sort/code/lisp/bubble_sort.lisp
+++ b/contents/bubble_sort/code/lisp/bubble_sort.lisp
@@ -1,17 +1,30 @@
 ;;;; Bubble sort implementation
 
-;;; Swaps two elements in a list (complexity: O(n))
-(defun swap (lst low high)
-  (let ((list-tail (nthcdr low lst)))
-  (rotatef (car list-tail) (elt list-tail (- high low)))
-  lst))
+(defun bubble-up (list)
+  (if
+    (< (length list) 2) 
+    list  
+    (if 
+      (> (first list) (second list))
+      (cons 
+        (second list) 
+        (bubble-up 
+          (cons 
+            (first list) 
+            (rest (rest list)))))
+            
+      (cons 
+        (first list)
+        (bubble-up
+          (cons
+            (second list)
+            (rest (rest list))))))))
 
-(defun bubble-sort (lst)
-  (dotimes (m (- (length lst) 1) lst)           ;loop
-    (dotimes (n (- (length lst) (+ 1 m)) lst)   ;loop   
-      (if (> (nth n lst) (nth (+ n 1) lst))     ;if
-      (swap lst n (+ n 1))                      ;then
-      lst))))                                   ;else
+(defun bubble-sort (list)
+  (if 
+    (equal list (bubble-up list))
+    list
+    (bubble-sort (bubble-up list))))
 
 ;; The built-in sort: (sort (list 5 4 3 2 1) #'<)
 (print 
--- a/contents/bubble_sort/code/lisp/bubble_sort.lisp
+++ b/contents/bubble_sort/code/lisp/bubble_sort.lisp
@@ -11,20 +11,20 @@
         (bubble-up 
           (cons 
             (first list) 
-            (rest (rest list)))))
-            
+            (rest (rest list)))))           
       (cons 
         (first list)
         (bubble-up
-          (cons
-            (second list)
-            (rest (rest list))))))))
+          (rest list))))))
 
 (defun bubble-sort (list)
   (if 
-    (equal list (bubble-up list))
+    (< (length list) 2)
     list
-    (bubble-sort (bubble-up list))))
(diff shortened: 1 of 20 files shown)


contents/bubble_sort/code/lisp/bubble_sort.lisp
```diff
@@ -0,0 +1,18 @@
...
+
+(defun bubble_sort (lst)
+    (dotimes (m (- (length lst) 1) lst)                     ;loop
+        (dotimes (n (- (length lst) (+ 1 m)) lst)           ;loop   
+            (if (> (nth n lst) (nth (+ n 1) lst))           ;if
+            (swap lst n (+ n 1))                            ;then
+            lst))))                                         ;else
+
+;;Use of an array insead of a list would be faster
+;;The built-in sort is also quicker (sort (list 5 4 3 2 1))
```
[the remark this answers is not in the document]
MichalHanajik: Is *insead* correct word?

contents/bubble_sort/code/lisp/bubble_sort.lisp
```diff
@@ -0,0 +1,30 @@
...
+      (swap lst n (+ n 1))                      ;then
+      lst))))                                   ;else
+
+;; Use of an array instead of a list would be faster
+(print 
+  (bubble-sort (list 5 4 3 2 1)))
+(print
+  (bubble-sort (list 1 2 3 3 2 1)))
+
+;; The built-in sort is quicker, see the usage below
```
jiegillet: This comparison part below is not useful, you can get rid of it

contents/bubble_sort/code/lisp/bubble_sort.lisp
```diff
@@ -0,0 +1,30 @@
...
+  lst))
+
+(defun bubble-sort (lst)
+  (dotimes (m (- (length lst) 1) lst)           ;loop
+    (dotimes (n (- (length lst) (+ 1 m)) lst)   ;loop   
+      (if (> (nth n lst) (nth (+ n 1) lst))     ;if
+      (swap lst n (+ n 1))                      ;then
+      lst))))                                   ;else
+
+;; Use of an array instead of a list would be faster
```
jiegillet: Why don't you use an array then?

contents/bubble_sort/code/lisp/bubble_sort.lisp
```diff
@@ -0,0 +1,30 @@
+;;;; Bubble sort implementation
+
+;;; Swaps two elements in a list (complexity: O(n))
```
jiegillet: For the moment you're using `nth` and `swap` to check and swap the values, which are both O(n). It's not like we are aiming for performance here, but it seems like you could do make everything a little tighter by navigating through the list and making changes as they come up, and loop over that process only once. I don't know Lisp, so I don't know if it's hard to do, though.

contents/bubble_sort/code/lisp/bubble_sort.lisp
Trashtalk217: A list looks a bit prettier and - turns out - an array doesn't work with `nth` only works with lists. I probably need to look at that code or that comment... :(

jiegillet: Other solution: remove the comment altogether and don't open the can of worms :)

contents/bubble_sort/code/lisp/bubble_sort.lisp
Trashtalk217: I'm sorry, but I don't think I can fix this. Other sorts are way easier to implement, and my knowledge of lisp is limited. I could just copy it from elsewhere and it would be more efficient. But should I do that?

jiegillet: No, no, don't copy code. And I know how awkward bubble sort is for functional programming, I did it in Haskell.
I was thinking go a function like this (in pseudocode):
```
function bubble-up (list):
  if length(list) < 2 : return list
  else:
    if first(list)>second(list): second(list) + bubble-up (first(list) + rest of the list)
    else: first(list) + bubble-up (second(list) + rest of the list)
```
Then you run `bubble-up` `n` times and _voila_. Is that difficult to implement?

contents/bubble_sort/code/lisp/bubble_sort.lisp
```diff
@@ -1,17 +1,30 @@
...
+        (second list) 
+        (bubble-up 
+          (cons 
+            (first list) 
+            (rest (rest list)))))
+            
+      (cons 
+        (first list)
+        (bubble-up
+          (cons
```
jiegillet: I think `(cons (second list) (rest (rest list))  )` is the same as `rest list`

Trashtalk217: Good point!

PR #350, opened 2018-08-09
Bubble sort: run MATLAB implementation and Markdown fixes

berquist:


berquist:
I don't understand. Racket is a Scheme, and there is no Racket in highlight.js (your version or the new one), only Scheme.

jiegillet:
"Racket is a Scheme" was a bit hard to understand for us since no one knew Scheme or Racket. 
I don't know why Racket isn't in highlight.js, I don't think we've modified that file before. If you want to give it a try, please do!

berquist:
Ok, in case you don't know, this can be confusing. In the 70s, Lisp evolved into essentially two different major forms: Common Lisp and Scheme (Clojure is probably the third but new, and Emacs Lisp the antiquated fourth). Because the language specifications are relatively small (compared to say C++), there are different dialects and implementations of each. Saying "Racket is a Scheme" means Racket implements some part of the Scheme language standard, plus a bunch of non-standard extra functionality that differentiates Racket from Chicken and Chez. There can also be non-dialect implementation differences, like how one would expect Intel compilers to produce faster code on Intel CPUs than GNU compilers, etc.

However, all dialects of Scheme and Common Lisp will share a common set of core language keywords (`define` in Scheme, `defun` in CL/Elisp). _That_ is what the different syntax highlighting covers, and why you shouldn't expect anything other than a very specialized editor or editor mode to differentiate deeper than (Common) Lisp, Emacs Lisp, Clojure, or Scheme.

jiegillet:
Thank you, that was very helpful.
Now knowing that, that makes me think there should be a way to define "alias" languages that redirect all Schemes to Scheme. Unfortunately every time I looked into those highlight source code, my head got dizzy :p

PR #354, opened 2018-08-13
Jarvis March Implementation in Java

crafter312:
It works, but I could use some tips on how to improve it!


contents/jarvis_march/code/java/JarvisMarch.java
```diff
@@ -0,0 +1,67 @@
...
+    else {
+      double bs = Math.sqrt(Math.pow(a[0]-b[0],2) + Math.pow(a[1]-b[1],2));
+      double cs = Math.sqrt(Math.pow(c[0]-b[0],2) + Math.pow(c[1]-b[1],2));
+      double as = Math.sqrt(Math.pow(a[0]-c[0],2) + Math.pow(a[1]-c[1],2));
+      double theta = Math.acos(((bs*bs)+(cs*cs)-(as*as))/(2*bs*cs));
+      return theta;
+    }
+  }
+  
+  public static ArrayList<Integer[]> jarvisMarch(ArrayList<Integer[]> arr) {
```
Trashtalk217: I kinda feel like this should be encapsulated in a private method named `getLeftestPoint`, but you really only need it once so it's not like that will safe you much space. Discussion is open on this one...

contents/jarvis_march/code/java/JarvisMarch.java
```diff
@@ -0,0 +1,67 @@
...
+      if(pointOnHull[0]==hull.get(0)[0]&&pointOnHull[1]==hull.get(0)[1])
+        break;
+      else {
+        hull.add(pointOnHull);
+      }
+    }
+    return hull;
+  }
+  
+  public static void main(String[] args) {
```
Trashtalk217: Isn't there a way to do this all at once and define the ArrayList without the for-loop?

contents/jarvis_march/code/java/JarvisMarch.java
```diff
@@ -0,0 +1,67 @@
...
+    for(Integer[] p: gift) {
+      System.out.println(Arrays.toString(p));
+    }
+    ArrayList<Integer[]> wrapping = jarvisMarch(gift);
+    System.out.println("Wrapping:");
+    for(Integer[] p: wrapping) {
+      System.out.println(Arrays.toString(p));
+    }
+  }
+
```
Trashtalk217: I don't know why but a lot of styleguides hate it when you forget an enter at the end of a file.

contents/jarvis_march/code/java/JarvisMarch.java
```diff
@@ -0,0 +1,67 @@
+import java.util.*;
+
+class JarvisMarch {
+
```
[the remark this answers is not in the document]
crafter312: Ok, that makes sense. I’ve never used Vector2D, so i’ll read about it.

leios:
So it looks like this one was mostly reviewed by @Trashtalk217 , what is the current state of this PR @crafter312 ?

Trashtalk217:
@crafter312 
I've looked into the library issue and there are a couple of things I found:
The javax library is maintained by Oracle and those are also the guys maintaining Java itself.
However Vector2D and the whole of javax is not part of the core, meaning it requires some extra setup for a theoretical project where someone has to download and setup javax. Another option could be making a small Vector2D class for yourself like this:

````    
Class Vector2D {
        float x, y

        public float angle(Vector2D) {
        //etc
        }
````
I'm kinda on the fence on which to choose, which one do you prefer?

**Comment by CDsigma** [comment]
I support the idea of using a nested class. I think it would be the best option to keep the code approachable and concise

File: contents/jarvis_march/code/java/JarvisMarch.java (hunk 1 above)
Comment (crafter312) [reply contents/jarvis_march/code/java/JarvisMarch.java]: I think I'll try that because it will make the code a bit more understandable.

File: contents/jarvis_march/code/java/JarvisMarch.java [hunk 5]
```diff
@@ -0,0 +1,100 @@
...
+    }
+    public Point(Integer[] p) {
+      x = p[0]; y = p[1];
+    }
+    
+    public double getX() { return x; }
+    public double getY() { return y; }
+    
+    public boolean equals(Point p) {
+      if(p.getX()==x&&p.getY()==y)
```
PudottaPommin: We have recently wrote [Code style guide](https://github.com/algorithm-archivists/algorithm-archive/wiki/Code-style-guide). In general is usage of spaces. So all places like `if(p.getX()==x&&p.getY()==y)` should be `if (p.getX() == x && p.getY() == y)` and also I have to ask you to add braces to this and all conditions.

contents/jarvis_march/code/java/JarvisMarch.java
```diff
@@ -0,0 +1,100 @@
...
+        return true;
+      else
+        return false;
+    }
+    public double magnitude() {
+      return Math.sqrt(Math.pow(x, 2) + Math.pow(y, 2));
+    }
+    public void print() {
```
PudottaPommin: Could you remove this print? In AAA is anti-patter to have printing/outputting values outside of main method (few exceptions).

contents/jarvis_march/code/java/JarvisMarch.java
```diff
@@ -0,0 +1,100 @@
...
+    System.out.println("Gift:");
+    for(Integer[] p: lst2) {
+      System.out.println(Arrays.toString(p));
+    }
+    
+    //find and print the array of points in the hull
+    ArrayList<Point> hull = jarvisMarch(gift);
+    System.out.println("Wrapping:");
+    for(Point p: hull) {
+      p.print();
```
PudottaPommin: This can be replaced with content of `print`

contents/jarvis_march/code/java/JarvisMarch.java
```diff
@@ -0,0 +1,100 @@
...
+  
+  public static void main(String[] args) {
+    
+    //test array setup
+    Integer[][] lst = new Integer[][]{{-5, 2}, {5, 7}, {-6, -12}, {-14, -14}, {9, 9},
+                                      {-1, -1}, {-10, 11}, {-6, 15}, {-6, -8}, {15, -9}, 
+                                      {7, -7}, {-2, -9}, {6, -5}, {0, 14}, {2, 8}};
+    ArrayList<Integer[]> lst2 = new ArrayList<Integer[]>(Arrays.asList(lst));
+    ArrayList<Point> gift = new ArrayList<Point>();
+    for(Integer[] i: lst2) {
```
PudottaPommin: This whole part would be much more readable if you define just `new ArrayList<Point>()` and add all points directly into that, instead of this transformation. 
When you change it to creating points, you can remove constructor `Point(Integer[] p) `.

contents/jarvis_march/code/java/JarvisMarch.java
crafter312: I fixed my usage of spaces. Also, the editor I use for Java was configured to use two spaces for indentation, so I changed that to four.

PudottaPommin:
Great job and thank you for improving AAA! :)

PR #355, opened 2018-08-18
implemented bogo sort in nim

c252:
I implemented bogo sort in nim, I also fixed a small formatting error I made in my last commit under book.json (the nim lang section was off a space).


jiegillet:
The formatting in book.json is still weird. The code looks fine to me though

PR #356, opened 2018-08-18
implemented monte carlo integration in nim

c252:
monte carlo integration in nim.

### DIFF
--- a/contents/monte_carlo_integration/code/nim/monte_carlo.nim
+++ b/contents/monte_carlo_integration/code/nim/monte_carlo.nim
@@ -1,0 +1,27 @@
+import random
+import math
+
+randomize()
+
+proc in_circle(x,y,radius: float): bool =
+    return x * x + y * y < radius * radius
+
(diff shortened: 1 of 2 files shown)


contents/monte_carlo_integration/code/nim/monte_carlo.nim
```diff
@@ -0,0 +1,27 @@
...
+randomize()
+
+proc in_circle(x,y,radius: float): bool =
+    return x * x + y * y < radius * radius
+
+proc monte_carlo(samples: int): float =
+    const radius: float = 1
+    var count: int = 0
+
+    for i in 0 .. < samples:
```
jiegillet: This notation is very clean, I like Nim :)

c252: Yeah a cool thing about nim is that it is incredibly flexible, you can compile it to c, c++, or javascript. Because of this you can use any existing libraries for c or c++ or javascript.

PR #359, opened 2018-08-22
implemented bubble sort in nim

c252:
Implemented the bubble sort algorithm in nim.

### DIFF
--- a/contents/bubble_sort/bubble_sort.md
+++ b/contents/bubble_sort/bubble_sort.md
@@ -47,7 +47,7 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 {% sample lang="lisp" %}
 [import:3-28, lang:"lisp"](code/lisp/bubble_sort.lisp)
 {% sample lang="nim" %}
-[import:6-10, lang:"nim"](code/nim/bubble_sort.nim)
+[import:5-9, lang:"nim"](code/nim/bubble_sort.nim)
 {% endmethod %}
 
 ... And that's it for the simplest bubble sort method.
--- a/contents/bubble_sort/bubble_sort.md
+++ b/contents/bubble_sort/bubble_sort.md
@@ -46,6 +46,8 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 [import:3-15, lang:"php"](code/php/bubble_sort.php)
 {% sample lang="lisp" %}
 [import:3-28, lang:"lisp"](code/lisp/bubble_sort.lisp)
+{% sample lang="nim" %}
+[import:6-10, lang:"nim"](code/nim/bubble_sort.nim)
 {% endmethod %}
 
 ... And that's it for the simplest bubble sort method.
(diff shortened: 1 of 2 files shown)


contents/bubble_sort/code/nim/bubble_sort.nim
```diff
@@ -0,0 +1,14 @@
...
+    for n in 0 .. len(a)-1:
+        echo a[n]
+
+proc bubble_sort(a: var openArray[int]) =
+    for i in 0 .. < len(a):
+        for j in 0 .. < (len(a) - 1):
+            if a[j+1] > a[j]:
+                swap(a[j],a[j+1])
+
+var x: array[10,int] =  [32,32,64,16,128,8,256,4,512,2]
```
jiegillet: Print the unsorted array first, maybe add some message: "Unsorted: " "Sorted: "

contents/bubble_sort/code/nim/bubble_sort.nim
```diff
@@ -0,0 +1,14 @@
+
```
jiegillet: Starting with an empty line bothers me more than it probably should

contents/bubble_sort/code/nim/bubble_sort.nim
```diff
@@ -0,0 +1,14 @@
...
+proc print_array(a: openArray[int]) =
+    for n in 0 .. len(a)-1:
```
jiegillet: You use `.. < len(a)`below. Either is fine, but consistency is better.

c252: I only need to do < len(a) for the j loop because otherwise it will be out of bounds. If I do that with the other two, it won't go through the whole array. Thanks for spotting that.

jiegillet: That's not exactly what I meant.
There are to identical ways to loop for an array, either `0 .. len(a)-1` or `0 .. < len(a)`. I just thought it would look better to stick to only one notation, now you are using `<` twice and `-1` once. I think using either `<` or `-1` everywhere is better.
Unrelated issue, you are now running the loop for `i` `len(a)+1` times, where I think you only really need `len(a)-1`. The loop for `j` is good as it is now.

c252:
I think I fixed it now.

jiegillet:
Four more things (in increasing order of importance):
- You don't need the parentheses in `0 .. < (len(a) - 1)`
- You haven't addressed my request for adding  "Unsorted: " "Sorted: "
- You are still doing one pass to many on your loop for `i` (If you sort a array of two values, you only need one pass to sort it)
- You are sorting the array in descending order, the norm for sorting algorithms is sorting in ascending order

contents/bubble_sort/bubble_sort.md
```diff
@@ -46,6 +46,8 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 [import:3-15, lang:"php"](code/php/bubble_sort.php)
 {% sample lang="lisp" %}
 [import:3-28, lang:"lisp"](code/lisp/bubble_sort.lisp)
+{% sample lang="nim" %}
+[import:6-10, lang:"nim"](code/nim/bubble_sort.nim)
```
jiegillet: Change `6-10` to `5-9` since you removed that empty line

c252:
Oh I see what you mean :). Thanks for being so helpful Jie, I really
appreciate it.

PR #360, opened 2018-08-24
Implement bogo in ruby

jasmaa:


contents/bogo_sort/code/ruby/bogo.rb
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env ruby
+
+def is_sorted(a)
+  for i in 0...a.length-1
```
jiegillet: It should be `a.length-2` right?

contents/bogo_sort/code/ruby/bogo.rb
```diff
@@ -0,0 +1,24 @@
...
+def bogo_sort(a)
+  while !is_sorted(a)
+    a.shuffle!
+  end
+end
+
+def main()
+	a = [1, 1, 0, 3, 7]
+	bogo_sort(a)
+	print a
```
jiegillet: Print something like "Unsorted: " then the array, then "Sorted:"

contents/bogo_sort/code/ruby/bogo.rb
```diff
@@ -0,0 +1,24 @@
...
+end
+
+main()
```
jiegillet: I think the file should end with a newline

contents/bogo_sort/code/ruby/bogo.rb
jasmaa: The range is exclusive and it runs up to the penultimate element so it should be fine I think?...

jiegillet: Ah, sorry `...` is exclusive, you're right then!

PR #363, opened 2018-09-04
Monte Carlo integration Fortran 90

depate:
Hi all!

The code was compiled with `gfortran` (gcc-8.2.0) with the following command:

`gfortran -Og -Wall -o montecarlo monte_carlo_pi.f90` .

// The warning or error is concerning the implicit call of the `in_circle` function.

// Thanks in advance for any suggestion or idea solving this issue.

Compiling the code on an other machine with `gfortran -Og -Wall -Werror -o montecarlo monte_carlo.f90` does not raise the error.

Best regards,

Patrik


leios:
Thanks a bunch and welcome to the AAA community!

A few notes:
1. Please remove the binary from the PR
2. I don't think the `.gitignore` is necessary here either
3. Please put your name in the `CONTRIBUTORS.md` file in the main directory
4. This is the first time we have fortran90 in the AAA, so please put the language in the `book.json` file (following the example of the other languages). Note here that the "lang" value will be the filename extension (here `.f90`) and the "name" will be the language name (Here `Fortran90`).
5. Please also update the `monte_carlo_integration.md` file with the fortran entry, using the tag you created in the `book.json` file and following the example from the other languages. Basically, you just need to grab the line numbers for the in-text snippets and then put them in the appropriate spot in the `method`. It will probably look like this:
```
{% sample lang="f90" %}
[import:3-10, lang:"fortran"](code/f90/monte_carlo.f90)
```

Sorry there is so much bookkeeping to do! I'll take a look at the actual code soon. I have a lot of PR's to go through, but this one is Fortran, so it's on my list too.

Thanks again!

contents/monte_carlo_integration/code/fortran/monte_carlo.f90
```diff
@@ -0,0 +1,45 @@
+FUNCTION in_circle(pos_x, pos_y, r) 
+IMPLICIT NONE
```
leios: I realize that fortran77 doesn't usually have indenting for functions, but I thought everything about F90 did? I don't know if this is standard or not, but it might help with readability

contents/monte_carlo_integration/code/fortran/monte_carlo.f90
```diff
@@ -0,0 +1,45 @@
...
+REAL(16), INTENT(IN) :: pos_x, pos_y, r
+LOGICAL :: in_circle
+in_circle = (pos_x ** 2 + pos_y ** 2) < r ** 2
+END FUNCTION in_circle 
+
+PROGRAM monte_carlo
+IMPLICIT NONE
```
leios: Again, a bit of spacing would help readability.

contents/monte_carlo_integration/code/fortran/monte_carlo.f90
```diff
@@ -0,0 +1,45 @@
...
+       
+    IF (in_circle(pos_x, pos_y, r) .EQV. .TRUE.) THEN 
+        pi_count = pi_count + 1d0
+    END IF
+END DO
+
+pi_est   = 4d0 * pi_count / n
+pi_error = 100d0 * (abs(pi_est - pi)/pi)
+
+WRITE(*,*) 'The pi estimate is: ', pi_est
```
leios: Could we do some formatting here to truncate the number of significant digits output for this result?

leios:
Thanks again! Sorry it took me so long to get to this PR. I'll tackle the other Fortran PR's after this one. It's ultimately good, but maybe add some formatting to the output and some spacing to the functions. The spacing is your call. I am not sure what "standard fortran" looks like nowadays.

contents/monte_carlo_integration/code/fortran/monte_carlo.f90
```diff
@@ -0,0 +1,50 @@
...
+    IMPLICIT NONE
+    REAL(16), INTENT(IN) :: pos_x, pos_y, r
+    LOGICAL              :: in_circle
+
+    in_circle = (pos_x ** 2 + pos_y ** 2) < r ** 2
+
+END FUNCTION in_circle 
+
+PROGRAM monte_carlo
+IMPLICIT NONE
```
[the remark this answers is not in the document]
depate: Will be done. I did this in the following contributions as well.

contents/monte_carlo_integration/code/fortran/monte_carlo.f90
```diff
@@ -0,0 +1,51 @@
+FUNCTION in_circle(pos_x, pos_y, r)
+    IMPLICIT NONE
+    REAL(16), INTENT(IN) :: pos_x, pos_y, r
+    LOGICAL              :: in_circle
+
+    in_circle = (pos_x ** 2 + pos_y ** 2) < r ** 2
+
+END FUNCTION in_circle 
+
+PROGRAM monte_carlo
```
leios: to be clear: When I said "spacing" before, I was actually asking to index this block. Sorry for the confusion.

depate: Ok, got it. This feels a bit odd because it's unfamiliar to see Fortran Code this way, but I'll get along with it.

contents/monte_carlo_integration/code/fortran/monte_carlo.f90
depate: ``` I used to use 6 spaces because that was what Fortran77 required ```
Ah, that's interesting. I was already wondering, why my programming helper addon in VIM always put 6 spaces in.

contents/monte_carlo_integration/code/fortran/monte_carlo.f90
depate: As I understand the reference one can use indenting optionally. But in this case consistent use of 2, 3 or 4 spaces should be followed.

The reference manual only uses indentation for code control structures like DO blocks etc. But I also saw references, e.g. from companies, that take the indentation further for all structures e.g. FUNCTION and SUBROUTINE procedures. 

So we are pretty much free to set our own convention. I would suggest to stick to 'indent all the things' scheme. You, and I guess, many other programmers or reviewers are used to indented source codes. But should any other contributor insist on the more historical 'tightly block and pack all the things', I suggest we may also accept this. Because the reference is only strict about consistency. 

It may be nice to look into the history of the different FORTRAN references and their style in the not-yet-done section of the book.

PR #364, opened 2018-09-04
Verlet integration in Fortran90

depate:
Hi all!

The fortran code was compiled with `gfortran` in `gcc 8.2.0` with:
`gfortran -Wall -Og -Werror -o verlet verlet.f90`

and raises no errors.

Best regards,

Patrik


leios:
Hey, thanks for the PR! As a quick note, changes to your Monte Carlo PR are also in this PR, which means that if we merge this one, we will also be merging the Monte Carlo stuff. I don't mind treating this as one large PR and reviewing both submissions at the same time; however, this is already on a separate feature branch for you, so you can just remove the files from this PR and we can review them separately.

depate:
I made some unknown mistake somewhere. 

@julianschacher and me fixed the PR. The monecarlo files were faulty, e.g. monte_carlo.f90 empty.

contents/verlet_integration/code/fortran/verlet.f90
```diff
@@ -0,0 +1,108 @@
...
+            next_pos = pos * 2d0 - prev_pos + acc * dt ** 2
+            prev_pos = pos
+            pos      = next_pos
+        ELSE
+            EXIT
+        END IF
+    END DO
+END SUBROUTINE verlet
+
+SUBROUTINE stromer_verlet(pos, acc, dt, time, vel)
```
leios: typo here. It should be `stormer_verlet()`

PR #365, opened 2018-09-04
Add bogo sort in Golang

christopherm99:
This was compiled the `go1.10.3` compiler on linux, and raises no errors.


leios:
Hey, sorry it's taking so long to get to this PR. My go knolwedge isn't great and we don't have too many reviewers who use the language. Thanks for the submission!

contents/bogo_sort/code/go/main.go
```diff
@@ -0,0 +1,39 @@
...
+package main
+
+import (
+	"fmt"
+	"math/rand"
+	"time"
+)
+
+func shuffle(a []int) []int {
+	rand.Seed(time.Now().UnixNano())
```
Liikt: You should rarely (I dare to say never) seed your prng more than once. I would move that line into the main.

contents/bogo_sort/code/go/main.go
```diff
@@ -0,0 +1,39 @@
...
+		if a[i+1] < a[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func bogo_sort(a *[]int) {
+	for !is_sorted(*a) {
+		*a = shuffle(*a)
```
Liikt: If you already work with pointers why not make the shuffe inplace. So basically change `*a = shuffle(*a)` to `shuffle(a)`. 
I would say either remove all pointers and let every function return it's value or change line 15 to ` (*a)[i], (*a)[j] = (*a)[j], (*a)[i]` and take in a `*[]int` as a parameter in shuffle.

contents/bogo_sort/code/go/main.go
```diff
@@ -0,0 +1,39 @@
...
+func is_sorted(a []int) bool {
+	for i := 0; i < len(a)-1; i++ {
+		if a[i+1] < a[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func bogo_sort(a *[]int) {
```
Liikt: Oh and as a general style rule golang uses camelcase and not underlines. Meaning the function should be called `bogoSort` and not `bogo_sort`

christopherm99:
Sorry, this must have slipped my mind. I have submitted a commit which should resolve the requested changes.

PR #366, opened 2018-09-06
Euclidean julia edit

depate:
I recognised that the code examples in the page markdown were in an odd order.


Gathros:
Remove the Fortran file.

PR #367, opened 2018-09-06
Adding Euclidean Algorithms in X86-64 Assembly

Gathros:


leios:
Hey @Gathros I know you are aware of this, but we are currently lacking the ability to adequately review this code because no one really knows assembly here. Sorry about that. This goes for #368 as well.

Liikt:
Looks good to me now

PR #368, opened 2018-09-06
Bubble sort in X86 Assembly

Gathros:


contents/bubble_sort/code/x86asm/bubble_sort.asm
```diff
@@ -0,0 +1,147 @@
...
+    mov ebp, esp
+    push esi
+    push edi
+    push ebx
+    sub esp, 4
+    mov DWORD [ebp - 16], 0
+    jmp .for_check
+.for_loop:
+    mov eax, DWORD [ebp - 16]
+    lea edx, [0 + eax * 4]
```
Liikt: I would remove the `0 +` since it is not really needed. This goes for all the following `0 +` as well.

contents/bubble_sort/code/x86asm/bubble_sort.asm
```diff
@@ -0,0 +1,147 @@
...
+    jb .for_loop
+    push 10
+    call putchar
+    add esp, 4
+    pop ebx
+    pop edi
+    pop esi
+    leave
+    ret
+
```
Liikt: Just as a style thing i would remove one of the newlines here.

Liikt:
I probably prefer @Gustorn solution a bit more since it is cleaner and is x86-64 and thus has a sane calling convention. Also the bubblesort is indeed more straight forward. 

Another thing I just realized while reading Gastorns solution that comments would've been really nice. Especially because not all people can read asm that easily.

Gustorn:
I prefer the `.rodata` solution because it's conceptually more similar to things you're used to from other programming languages (it also mirrors the C solution). But I can see the argument for either version.

Liikt:
I totally agree to that the `.rodata` solution mirrors C better. You just tend to see the `.rwdata` in classes that teach assembly as it is generally easier to understand. Though that is more a minute thing and I'm happy with either way.

Gathros:
Should we be doing X86-64 assembly instead x86 assembly?

Gustorn:
My vote would be yes, that's what most people would see nowadays anyway (during debugging, etc.)

Liikt:
I would also say that we should use x86-64 over x86 for basically the same reasons.

PR #375, opened 2018-09-12
Remove credit for ubsan (and in general) from two C++ files.

julianschacher:
After I talked to ubsan and told her about the C++ style decisions, she requested me to remove her name from the C++ files, that contain her name.
So this PR does exactly that.

It also removes the credit in general, if you (@leios ) don't want that, feel free to tell me, then I re-add it again.


leios:
This is fine, I just wish she would talk to me about this.

PR #377, opened 2018-09-13
Tiny Tiny Edit to Thomas Algorithm in Python

CDsigma:
- Added some spaces


contents/thomas_algorithm/code/python/thomas.py
```diff
@@ -8,17 +8,17 @@ def thomas(a, b, c, d):
     d[0] = d[0] / b[0]
 
     n = len(d) #number of equations to solve
```
berquist: There should be a space between `#` and `number`.

contents/thomas_algorithm/code/python/thomas.py
```diff
@@ -8,17 +8,17 @@ def thomas(a, b, c, d):
...
-    for i in range(1,n):
+    for i in range(1, n):
         #scale factor for c and d
```
berquist: There should be a space between `#` and `scale`.

contents/thomas_algorithm/code/python/thomas.py
CDsigma: Agreed, I just made a commit to add a space there :)

CDsigma:
Thanks for the suggestions :)

1. I'm not exactly sure I understand what 'line ending whitespace' is. For clarification is that like line 18, where there is a tab but then no code which follows it?
2. I agree, I made a commit to address this.
3. I agree, that it would be a good idea to get rid of authorship comments throughout the project. I also think that decision isn't entirely made yet and is outside the scope of this PR. My understanding is that there isn't a clear consensus around removing authorship comments so I don't think I should remove any yet.

julianschacher:
@CDsigma We plan on removing all authorship in all code examples, since git does this job perfectly and better than any code comment ever could. But I agree, it does not have to be done in this PR.
I plan to make a PR removing all the author comments in following days anyway.

berquist:
>  For clarification is that like line 18, where there is a tab but then no code which follows it?

Yes. If you use an editor or IDE that supports [EditorConfig](https://editorconfig.org/), if should nuke that stuff automatically; that's how I noticed it as seemingly blank lines changing in my diff.

julianschacher:
For using the `.editorconfig` check out this short text here: https://github.com/algorithm-archivists/algorithm-archive/wiki/How-to-Contribute#step-13---editorconfig

CDsigma:
Thanks for the links @julianschacher and @berquist :)

Also I just made a commit to get rid of that whitespace 👍

contents/thomas_algorithm/code/python/thomas.py
```diff
@@ -2,24 +2,24 @@
...
+        d[i] = (d[i] - a[i] * d[i-1]) * scale
+
 
-    
     # do the back substitution
-    for i in range(n-2,-1,-1):
-        d[i] = d[i] - c[i]*d[i+1]
-    
+    for i in range(n-2, -1, -1):
+        d[i] = d[i] - c[i] * d[i+1]
```
berquist: Can you change this to `d[i] -= c[i] ...`?

contents/thomas_algorithm/code/python/thomas.py
```diff
@@ -2,24 +2,24 @@
...
-    for i in range(1,n):
-        #scale factor for c and d
-        scale = 1 / (b[i] - c[i-1]*a[i])
-        
+    n = len(d) # number of equations to solve
+    for i in range(1, n):
+        # scale factor for c and d
+        scale = 1 / (b[i] - c[i-1] * a[i])
+
         c[i] = c[i] * scale
```
berquist: `c[i] *= scale`?

CDsigma:
Thanks for the input! I have implemented the changes

PR #379, opened 2018-09-14
Update Julia bogosort

berquist:
- bug that caused sorting order to be reversed
- spaces after commas
- Julia v1 has shuffle! in the Random package. I don't know how long the transitional period to 1.0 is, but is it a problem if this doesn't work with < 1.0?

### DIFF
--- a/contents/bogo_sort/bogo_sort.md
+++ b/contents/bogo_sort/bogo_sort.md
@@ -14,7 +14,7 @@ In code, it looks something like this:
 
 {% method %}
 {% sample lang="jl" %}
-[import:10-14, lang:"julia"](code/julia/bogo.jl)
+[import:12-16, lang:"julia"](code/julia/bogo.jl)
 {% sample lang="cs" %}
 [import:9-15, lang:"csharp"](code/csharp/BogoSort.cs)
 {% sample lang="clj" %}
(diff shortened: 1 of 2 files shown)


leios:
I can take a look at the shuffle function soon. Thanks for the heads-up!

Butt4cak3:
> Julia v1 has shuffle! in the Random package. I don't know how long the transitional period to 1.0 is, but is it a problem if this doesn't work with < 1.0?

We're not targeting a production environment here. It's more important that the code examples are relevant and up-to-date, so I'd say go with it if it's possible in the newest version or else the code in the book will be outdated earlier than necessary.

leios:
I have been going through and updating the julia code for V1.0. Since you already have this PR up, do you mind just adding a `using Random` at the start of the file?

berquist:
I was asking if it would be a problem that code samples would not longer work with < v1, because of course Arch updated very quickly, but really no one should be using < v1 and the Julia team would be the first to say that.

> Since you already have this PR up, do you mind just adding a using Random at the start of the file?

https://github.com/algorithm-archivists/algorithm-archive/pull/379/files#diff-748545546903a491988fbfcecff0171fR1

leios:
oops. That was my bad. I missed that line somehow. Let me check build this locally, then it's good to go.

Issue #324, opened 2018-08-02
Monte Carlo Percent Error


julianschacher:
The percent error can be displayed in two different ways:
- Way 1: Having a positive or negative percent error depending on the estimate we have.
- Way 2: Always having a positive value by just taking the absolute.

The two ways would be calculated like the following:
- Way 1: percent error = `[experimental value - theoretical value] / theoretical value x 100%`
- Way 2: percent error = `(AbsoluteOf([experimental value - theoretical value])) / theoretical value x 100%`

We now have the following state for the Monte Carlo implementations of the AAA:
- Some implementations are using Way 2 of the percent error and just provide an always positive number. The following languages are using way 2: Clojure, D, Go, Haskell, Python, R
- And some other languages are using Way 1, but making a mistake by doing so, they aer using the following calculation: `[theoretical value - experimental value] / theoretical value x 100%` instead of `[experimental value - theoretical value] / theoretical value x 100%` The following languages are using way 1 incorrectly: C, Java, JavaScript, Julia, Rust, Swift

So the question is now: Should we use way 1 or way 2 in the AAA?

I vote for way 1 because it provides some nice additional information, i.e.:
Having a `theoretical value` of 3 would result in a negative percentage and having a `theoretical value` of 3.5 would result in a positive percentage. With way 2 it would be a positive percentage in both cases.

# Progress Tracker - Way 2

Since we decided for way 2, the following to do list, shows which implementations need to be updated:
- [x] Julia // Done: #325 
- [x] Swift // Done: #328 
- [x] Java // Done: #371
- [x] JavaScript // Done: #372 
- [x] C // Done: #374
- [x] Rust // #381

julianschacher:
-

julianschacher:
Alright, since we're doing way 2, I'll turn this issue into a progress tracker for the languages, that need an update.

julianschacher:
All the code examples are updated. This issue gets closed. (Woo! :D )

PR #381, opened 2018-09-14
Fixing percent error in monte_carlo.rs

Gathros:
#324

### DIFF
--- a/contents/monte_carlo_integration/code/rust/monte_carlo.rs
+++ b/contents/monte_carlo_integration/code/rust/monte_carlo.rs
@@ -26,5 +26,5 @@ fn monte_carlo(n: i64) -> f64 {
 fn main() {
     let pi_estimate = monte_carlo(10000000);
 
-    println!("Percent error is {:.3}%", (100.0 * (PI - pi_estimate) / PI));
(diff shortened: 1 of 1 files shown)


contents/monte_carlo_integration/code/rust/monte_carlo.rs
```diff
@@ -26,5 +26,5 @@ fn monte_carlo(n: i64) -> f64 {
 fn main() {
     let pi_estimate = monte_carlo(10000000);
 
-    println!("Percent error is {:.3}%", (100.0 * (PI - pi_estimate) / PI));
+    println!("Percent error is {:.3}%", (100.0 * (PI - pi_estimate).abs() / PI));
```
julianschacher: It works!
In the other PRs we also changed the order of `(PI - pi_estimate)` to `(pi_estimate - PI)`, so that the theoretical value gets subtracted from the experimental one.
It doesn't change the outcome, it's just nicer, so can you maybe change `(100.0 * (PI - pi_estimate).abs() / PI)` to `(100.0 * (pi_estimate - PI).abs() / PI)`?

julianschacher:
Looks good. Thanks for the changes!

PR #383, opened 2018-09-15
Adding Thomas Algorithm in Swift

CDsigma:
Added Thomas Algorithm in Swift


jiegillet:
The algorithm looks fine, the only thing that bothers me is that you're changing the input vectors. I would prefer spitting out the solutions without modifying the vectors. Was that a deliberate choice?

CDsigma:
After thinking about it I agree with you. Having the function output the answer without modifying the input makes more sense. I just made a commit to address this.

PR #384, opened 2018-09-17
Update javascript code for tree traversal

Mukundan314:
Changes:
  - Create `Node` Class
  - Lint Code with [StandardJS](https://standardjs.com/)

### DIFF
--- a/contents/tree_traversal/code/javascript/tree.js
+++ b/contents/tree_traversal/code/javascript/tree.js
@@ -1,64 +1,61 @@
 class Node {
-  constructor(id, children=[]) {
-    this.id = id;
-    this.children = children;
+  constructor (id, children = []) {
+    this.id = id
+    this.children = children
   }
 }
 
-function createTree(numRow, numChild) {
-  var root = new Node(numRow);
+function createTree (numRow, numChild) {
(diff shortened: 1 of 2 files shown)


Butt4cak3:
Hey, thanks for the effort, but I don't think we'll merge this one. Let me give you a few reasons why.

**Code formatting**: We don't use "Standard Style" in this project. That's all.

**Code style**: You seemingly arbitrarily mixed ES5 code with ES2015 features. You used `var` instead of `let`, but you left `const` in and used `class`. An environment that understands `const` or `class` will always understand `let` and the latter is *always* the better option. There is no need to use `var` anywhere. In fact, the only reason to use `var` these days is backwards compatibility with older browsers and that's not a priority in this project at all.

You also replaced a higher-order function with a for-loop for no apparent reason (`createTree`).

**Node class**: I'm sure that @Gustorn deliberately didn't use a class for nodes and I agree. There is no need to create a class for objects with no methods in JavaScript. It makes sense when you come from other languages, but that doesn't mean it's a good idea in JS.

I'll leave this PR open for a bit to give you the chance to voice your own opinion on the matter and to let others do the same.

PR #387, opened 2018-09-20
Adding Forward Euler in Java

CDsigma:
Added Forward Euler in Java


leios:
Hey @CDsigma I think you know, but we are missing some Java expertise, so I don't know who will review this. I'll do so if no one steps up.

contents/forward_euler_method/code/java/Euler.java
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
+    return isApprox;
+}
+
+  public static void main(String[] args) {
+    double timeStep = 0.01;
+    int n = 100;
+    double threshold = 0.01;
```
Yordrar: You can make these variables as memebers of the class so that all methods can access them without needing to pass them as arguments

Butt4cak3: I disagree. These should be passed as arguments. If the methods were non-static, then fine, but in this case passing them on is the way to go.

contents/forward_euler_method/code/java/Euler.java
```diff
@@ -0,0 +1,52 @@
...
+    
+    double[] result = solveEuler(timeStep, n);
+    boolean isApprox = checkResult(result, threshold, timeStep);
+
+    if(isApprox) {
+        System.out.print("All values within threshold");
```
Butt4cak3: Please use `System.out.println` here or otherwise add a newline to the output. I assume you're using Windows, so everything works as expected, but when I run this on Linux, my shell does this:

```
user@hostname:aaa> java ForwardEuler
All values within thresholduser@hostname:aaa>
```

instead of this:

```
user@hostname:aaa> java ForwardEuler
All values within threshold
user@hostname:aaa>
```

contents/forward_euler_method/code/java/Euler.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+  
+  public static boolean checkResult(double[] result, double threshold, double timeStep) {
+    boolean isApprox = true;
+
+    for(int i = 0; i < result.length; i++) {
+        double solution = Math.pow(Math.E, -3 * i * timeStep);
+      
+        if(Math.abs(result[i] - solution) > threshold) {
+            isApprox = false;
```
Butt4cak3: You're missing a `println` before inside this block. If you look at the reference implementation (Julia), you'll see this:

```Julia
if (abs(euler_result[i] - solution) > threshold)
    println(euler_result[i], solution)
    is_approx = false
end
```

contents/forward_euler_method/code/java/Euler.java
```diff
@@ -0,0 +1,52 @@
...
+    newArray[newArray.length - 1] = newItem;
+
+    return newArray;
+}
+  
+  public static double[] solveEuler(double timeStep, int n) {
+    double[] result = {1};
+
+    for(int i = 0; i < n; i++) {
+      result = append(result, result[i] - 3 * result[i] * timeStep);
```
Butt4cak3: You're thinking *way* too complicated here. You could just make `result` an array with a length of `n` (`double[] result = new double[n]`), add the initial value of `1` (`result[0] = 1`) and then fill the array like this:

```Java
result[i] = result[i - 1] - 3 * result[i - 1] * timeStep;
```

The loop would of course have to start from `1` to skip the first element.

This way, you could drop the `append` method above. In addition to that, if I'd need an `append` method for an array, I'd use an `ArrayList` instead.

contents/forward_euler_method/code/java/Euler.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    return result;
+  }
+  
+  public static boolean checkResult(double[] result, double threshold, double timeStep) {
+    boolean isApprox = true;
+
+    for(int i = 0; i < result.length; i++) {
+        double solution = Math.pow(Math.E, -3 * i * timeStep);
```
Butt4cak3: There`s a method `Math.exp(double a)` which does exactly this.

```Java
double solution = Math.exp(-3 * i * timeStep);
```

contents/forward_euler_method/forward_euler_method.md
```diff
@@ -121,6 +121,8 @@ Full code for the visualization follows:
 [import, lang:"matlab"](code/matlab/euler.m)
 {% sample lang="swift" %}
 [import, lang:"swift"](code/swift/euler.swift)
+{% sample lang="java" %}
+[import, lang:"java"](code/java/Euler.java)
```
Butt4cak3: Don't forget to update the file name in case you rename the source file.

Butt4cak3:
This PR has been stale for over 10 days now and you haven't shown up in Discord either. I'm assuming you're on a vacation or you're preoccupied by something else. Either way, we have another PR for Forward Euler in Java in #518, so I'm going to go ahead and work with that one. I'm sure you'll understand.

PR #389, opened 2018-09-23
Update all julia code to V1.0

leios:
This is a quick update to all the julia code so that it runs in Julia V1.0


berquist:
Everything else works and looks good.

leios:
I think the haskell graham scan is actually the best. It generates a list of random numbers within a circle and then checks the hull of that to see how far off it is from an actual circle.

I don't mind adding more points, though.

PR #392, opened 2018-09-26
Prevent Travis from deploying the book, even when it fails.

berquist:
https://docs.travis-ci.com/user/customizing-the-build/#customizing-the-build-step

Placing the deployment in `after_success` works around the fact that all
commands under `script` run even when one or more of them return a non-zero
exit code.


leios:
that was quick! thanks!

Issue #394, opened 2018-09-28
Missing aThat's oklgorithm implementations #hacktoberfest


leios:

leios:
Here is a list of all the partially complete language implementations:

Convolutions:
- [] Haskell

Cooley-Tukey:
- [] Haskell
- [] Scratch

Gaussian Elimination
- [] Rust

Tree Traversal
- [] JavaScript
- [] Haskell

Verlet Integration
- [] Matlab
- [] LabVIEW
- [] Scratch
- [] Haskell

Of course, if there are any language implementations that look funny, Feel free to update those too!

leios:
These issues are still present, even after hacktoberfest, so I will leave this open as a general issue. As a note, I might continually update this issue as more inconsistencies arise.

leios:
@jiegillet We are specifically missing some scratch code here.

I don't remember who did rust, but I do remember who did LabVIEW and matlab and I'm pretty sure those will not be updated soon... Not sure what to do with those for now.

leios:
I just updated the previous list to include information about the FFT chapter (undergoing revision). Basically, all of the implementations besides julia do not have an appropriate test against an actual FFT library.

As a note, you guys should maybe wait a week on this until after the revisions for this chapter ate complete.

varunkmr19:
Can I Implement Stack and Queue Code in Cpp?

leios:
We have no stack or queue code, but sure

PR #395, opened 2018-09-29
Haskell: added DFS Stack to Tree Traversal

jiegillet:
Added DFS Stack.
I also ran hindent which reformatted the file, so I had to change all import lines.

For the record, it's not _really_ using stacks, just as BFS is not _really_ using queues and the other DFS  implementations are definitely better. I'm just adding it for completeness.


leios:
this fixes issue #394, right?

jiegillet:
One of them, yes

PR #398, opened 2018-09-29
Haskell: added Stormer Verlet and Velocity Verlet

jiegillet:
Added all methods to Verlet algorithms.
Yay #hipsterhacktoberfest #394 

@leios I think your Julia code for calculating the velocity in Stormer Verlet is not correct. You use `v += a * dt` instead of `v = (pos - prev_pos) / dt`, which actually yields better results but it's not the method you explain in the text.


jiegillet:
Sorry for closing and re-opening I just want to check if this will add the PR on hacktoberfest. I won't do it for my other PRs.

PR #403, opened 2018-10-01
add huffman encoding in go

Liikt:
compiled with `go version go1.11 linux/amd64` and without errors


contents/huffman_encoding/code/golang/huffman.go
```diff
@@ -0,0 +1,127 @@
...
+	if node == nil {
+		return
+	}
+
+	if node.left == nil && node.right == nil {
+		(*cb)[node.char] = string(code)
+	}
+
+	// 0x30 == "0" and 0x31 == "1"
+	code = append(code, 0x30)
```
Gustorn: Why not just use `'0'` and `'1'`?

Liikt: i forgot that '1' is a rune/byte

PR #407, opened 2018-10-01
Euclidean algorithm in brainfuck

Ariana1729:
So I basically wrote Euclidean algorithm in brainfuck

Currently, the input is ascii and output is ascii, maybe I can parse a numerical input and print numerically? Not sure which is better


leios:
I think it would be nice to include the explanation into the main markdown file for the chapter in the language switching method. That way, when BF is chosen, it comes with an appropriate explanation.

jiegillet:
For the record, this is awesome.

jiegillet:
Oh yeah, you haven't included the code in the chapter. Also you need to add brainfuck to `book.json`

Ariana1729:
Ok so I've added a subtraction algorithm that was way longer than expected and an explanation

PR #408, opened 2018-10-01
Implement conv in python

jasmaa:
Note: Discrete convolution implemented differently because I couldn't get the previous implementations to get the same answer as the fft ones.

### DIFF
--- a/contents/convolutions/code/python/conv.py
+++ b/contents/convolutions/code/python/conv.py
@@ -1,0 +1,46 @@
+import numpy as np
+from scipy.fftpack import fft, ifft
+
+def conv(signal1, signal2):
+    """Discrete convolution by definition"""
+    
+    out = []
+    n = range(min(len(signal1), len(signal2)))
+
+    for i in n:
+        s = 0
+        for j in n:
+            s += signal1[j] * signal2[i - j]
+        out.append(s)
+
+    return out
+
+def conv_fft(signal1, signal2):
+    """Convolution using fft and convolutional theorem"""
+    
+    fft_s1 = fft(signal1)
+    fft_s2 = fft(signal2)
+    out = []
+
+    for i in range(min(len(signal1), len(signal2))):
+        out.append(fft_s1[i] * fft_s2[i])
+
+    return ifft(out)
+
+
+def main():
+    # Example convolution with sin and cos
+    x = np.linspace(0, 1, 5)
+    s1 = np.sin(x)
+    s2 = np.cos(x)
+
+    print("Discrete Convolution")
+    print(conv(s1, s2))
+
+    print("FFT Convolution")
+    print(conv_fft(s1, s2))
+
+
+if __name__ == "__main__":
+    main()
+
--- a/contents/convolutions/convolutions.md
+++ b/contents/convolutions/convolutions.md
@@ -44,6 +44,8 @@ In code, this looks something like:
 [import:5-18, lang:"c_cpp"](code/c/convolutions.c)
 {% sample lang="cpp"%}
 [import:68-88, lang:"c_cpp"](code/c++/convolutions.cpp)
+{% sample lang="python"%}
+[import:4-16, lang:"python"](code/python/conv.py)
 {% endmethod %}
 
 Note that in this case, the output array will be the size of `f[n]` and `g[n]` put together.
@@ -93,6 +95,8 @@ Where the `.*` operator is an element-wise multiplication.
 [import:20-30, lang:"c_cpp"](code/c/convolutions.c)
 {% sample lang="cpp"%}
 [import:90-105, lang:"c_cpp"](code/c++/convolutions.cpp)
+{% sample lang="python"%}
+[import:18-28, lang:"python"](code/python/conv.py)
 {% endmethod %}
 
 This method also has the added advantage that it will *always output an array of the size of your signal*; however, if your signals are not of equal size, we need to pad the smaller signal with zeros.


jiegillet:
That's normal, the FFT method does not work the same way.

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
jiegillet: You're only using numpy to make examples, I don't think it's worth it...

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,46 @@
...
+from scipy.fftpack import fft, ifft
+
+def conv(signal1, signal2):
+    """Discrete convolution by definition"""
+    
+    out = []
+    n = min(len(signal1), len(signal2))
```
jiegillet: You should end up with `len(signal1) + len(signal2) - 1` points, not `min(len(signal1), len(signal2))`

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,46 @@
...
+
+    for i in range(n):
+        s = 0
+        for j in range(n):
+            s += signal1[j] * signal2[i - j]
```
jiegillet: You're not overlapping all the points. Look at the Julia code for reference.
Don't try to mimic the FFT method, they have different assumptions.

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,46 @@
...
+    return out
+
+def conv_fft(signal1, signal2):
+    """Convolution using fft and convolutional theorem"""
+    
+    fft_s1 = fft(signal1)
+    fft_s2 = fft(signal2)
+    out = []
+
+    for i in range(min(len(signal1), len(signal2))):
```
jiegillet: I'm not sure if there is an official method in the case the signals have different lengths. I think instead of cutting the longer signal, I would add zeros to the shorter one until they have the same length. That's the usual method when using FFTs.

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,46 @@
...
+        out.append(fft_s1[i] * fft_s2[i])
+
+    return ifft(out)
+
+
+def main():
+    # Example convolution with sin and cos
+    x = np.linspace(0, 1, 5)
+    s1 = np.sin(x)
+    s2 = np.cos(x)
```
jiegillet: I think you can get away with something like `s1 = [ math.sin(x) for x in range(5)]` if we want to skip numpy.

contents/convolutions/code/python/conv.py
jasmaa: Thanks for the feedback!
I'm not that familiar with convolutions so please bear with me:

The discrete convolution from the definition is as listed:

f*g[n] = sum[-inf,inf][ f[m]*g[n-m] ]

In the code, I am going for each element of f*g using the outer for loop.

Taking the [-inf,inf] to effectively be the full length of the signal arrays,
for each element in f*g, I sum the values f[m]*g[n-m] where n and m are represented
by i in the outer for loop and j in the inner for loop respectively.
This j should go the length of the signal as m is in the range, [-inf,inf].

The code follows the discrete convolution definition to my knowledge and,
additionally, also appears to give the same answer as the fft version,
so I am unsure as to why it is not valid.

I will experiment with the Julia code when I can get a chance but a cursory
look at the code shows the length of the outputted signal to be the sum of
the two inputted signals. Since the outer for loop runs through this combined
length, isn't there a chance of an array-out-of-bounds when indexing the smaller
signal2 with i-j when i is big and j is small?

FFT also appears to output a signal of equal length to the input signal, so
I am unclear if there is ever a case where the fft version of the convolution
could give back the same output signal described by the Julia implementation.

I will try to find more information since there is likely a misunderstanding on
my part. Meanwhile, it would be helpful if some of these questions can be cleared up.

contents/convolutions/code/python/conv.py
jasmaa: This was based off of what I could gather from the C implementation on AAA, but I'll try to change it.

contents/convolutions/code/python/conv.py
jiegillet: The trick is to simulate a [-inf,inf] range by adding an infinite number of zeroes to the left and right of both signals, convolute them and clip the zeros at the end. For example, I want to convolve a=[1,2,3,4] and b=[8,9]. So if I flip b and slide it across a, I have:
```
0 0 1 2 3 4 0 0
9 8 0 0 0 0 0 0  => sum of product = 0

0 0 1 2 3 4 0 0
0 9 8 0 0 0 0 0  => sum of product = 0*9 + 1*8 = 8

0 0 1 2 3 4 0 0
0 0 9 8 0 0 0 0  => sum of product = 1*9 + 2*8 = 25

0 0 1 2 3 4 0 0
0 0 0 9 8 0 0 0  => sum of product = 2*9 + 3*8 = 42

0 0 1 2 3 4 0 0
0 0 0 0 9 8 0 0  => sum of product = 3*9 + 4*8 = 59

0 0 1 2 3 4 0 0
0 0 0 0 0 9 8 0  => sum of product = 4*9 + 0*8 = 36

0 0 1 2 3 4 0 0
0 0 0 0 0 0 9 8  => sum of product = 0
```
so the output should be [8, 25, 42, 59, 36], which is of length 5 = len(a)+len(b)-1.
This is the best way to mimic the continuous case. 

The FFT method reproduces the same result *only if* the signals you want to convolve are infinite and periodic. If you use the FFT method on finite signals it will assume that both signals repeat left and right forever, so if I did the same process above, we would have something like:

```
. . 1 2 3 4 . .
9 8 9 8 9 8 9 8  => sum of product = 1*9 + 2*8 + 3*9 + 4*8 = A

. . 1 2 3 4 . .
8 9 8 9 8 9 8 9  => sum of product = 1*8 + 2*9 + 3*8 + 4*9 = B

. . 1 2 3 4 . .
9 8 9 8 9 8 9 8  => sum of product = 1*9 + 2*8 + 3*9 + 4*8 = A

. . 1 2 3 4 . .
8 9 8 9 8 9 8 9  => sum of product = 1*8 + 2*9 + 3*8 + 4*9 = B
```
and the output would be [B, A] (or maybe [A,B]?) of length 2 = min(len(a), len(b)). 

But it gets a bit more messy when you actually use the FFT when you have signals of different lengths, you need to either cut the longest signal (which you have done, and is actually a valid choice) or pad the shorter signal with zeros (that would be my preference, but it's not more valid). 

Depending on your data you might chose different methods. The FFT is a faster algorithm but comes with different assumptions on the data.

contents/convolutions/code/python/conv.py
jiegillet: According to Wikipedia, both ways of doing it are valid, the results will a bit look different. As long as you're making an informed decision it's all good.

contents/convolutions/code/python/conv.py
jasmaa: Understood. Thanks for the explanation!

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,72 @@
+import math
+from scipy.fftpack import fft, ifft
+
+def conv(signal1, signal2):
+    """Discrete convolution by definition"""
+
+    n = len(signal1) + len(signal2)
+    signal1 = signal1.copy()
```
jiegillet: You don't need to modify the signals if your loops are done right. Lines 8 to 17 can be deleted without any effect on the algorithm.

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,72 @@
...
+    for i in range(max_len - len(signal1)):
+        signal1.append(0)
+    for i in range(max_len - len(signal2)):
+        signal2.append(0)
+        
+    # build output signal
+    out = []
+    
+    for i in range(n):
+        s = complex(0)
```
jiegillet: Why complex? just `0` is fine, if the data is complex it will get updated anyway.

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,72 @@
...
+        
+        for j in range(i):
+            if j < len(signal1) and i - j - 1  < len(signal2):
+                s += signal1[j] * signal2[i - j - 1]
+                
+        out.append(s)
+
+    return out[1:]
```
jiegillet: `return out`.

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,72 @@
+import math
+from scipy.fftpack import fft, ifft
+
+def conv(signal1, signal2):
+    """Discrete convolution by definition"""
+
+    n = len(signal1) + len(signal2)
```
jiegillet: Define `n = len(signal1) + len(signal2) -1` that's the most natural value

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,72 @@
...
+    for i in range(max_len - len(signal2)):
+        signal2.append(0)
+        
+    # build output signal
+    out = []
+    
+    for i in range(n):
+        s = complex(0)
+        
+        for j in range(i):
```
jiegillet: change to `range(i+1)` this way all the `i-1` become `i` and it's clearer.

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,72 @@
...
+            if j < len(signal1) and i - j - 1  < len(signal2):
```
jiegillet: `i - j - 1` => `i - j `

contents/convolutions/code/python/conv.py
```diff
@@ -0,0 +1,72 @@
...
+                s += signal1[j] * signal2[i - j - 1]
```
jiegillet: `i - j - 1` => `i - j `

PR #409, opened 2018-10-01
Added/fixed the C and C++ langs on the split operator method

rajdakin:
This pull request fixes the split operator method in C (as every code blocks were one line off), and adds the C++ implementation of that split operator method.

### DIFF
--- a/contents/split-operator_method/code/c++/split_op.cpp
+++ b/contents/split-operator_method/code/c++/split_op.cpp
@@ -1,0 +1,249 @@
+#include <complex>
+#include <iostream>
+#include <cstring>
+
+// Using fftw3 library.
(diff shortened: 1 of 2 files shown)


Gathros:
This should be two different pull requests not one.

leios:
This is actually just C++ code, so we should just modify the title to be exclusively C++, not C/C++

Still, we appreciate the submission!

Gathros:
@leios no, there is also changes to the markdown that change the way c code is displayed.

leios:
Oh, I see. Yeah, the line number changes should be a separate PR

rajdakin:
Creating two other pull requests, one for each change.

PR #411, opened 2018-10-01
Euler fortran

depate:
Hey all!

Another Fortran implementation done. 
Compiled without erros with `gcc-Version 8.2.1 20180831 (GCC)` using `gfortran -Og -Wall -Werror -std=f95 -fimplicit-none euler.f90 -o euler`.

Something odd with my git again. The merge history is shown. Any idea how to resolve this even though it does not collide with AAA's master branch?

Regards

depate


berquist:
This is how I would do it, assuming that you have `upstream` set to this repo:
```bash
git fetch --all
git checkout euler_fortran
git rebase -i upstream/master
```
This allows you to do an interactive rebase of your branch on top of the current master, so you can change anything in your history. The option I would take is to rewrite the commit message of the first commit to make sense, then "fixup" all other later commits into that one. It squashes everything but discards commit messages.

Compare this to when we hit "Squash and merge" on GitHub, which will pack everything into a single commit just fine, but keep all the nonsense commit messages.

leios:
I think this is a good PR. Quick question about the `CONTAINS` statement. Can that be used to clean up the interfaces in previous implementations?

PR #417, opened 2018-10-02
Bubble sort in scala implementation

kenpower:


leios:
I don't now scala so well, so we'll leave this up to see if someone else wants to do the review. Thanks for the submission!

If no one else does it, I'll do the review. I just might ask some dumb questions.

book.json
```diff
@@ -139,7 +139,11 @@
         {
           "lang": "f90",
           "name": "Fortran90"
-        }
+        },
+        {
+          "lang": "scala",
+          "name": "Scala"
+         }
```
jiegillet: Tiny nitpick here, the closing `}` is not aligned with the opening one

contents/bubble_sort/code/scala/bubble_sort.scala
```diff
@@ -0,0 +1,22 @@
+object BubbleSort {
+
+  def bubbleUpLargestToTheEnd(list: List[Int]): List[Int] =
+    list match {
+      case first :: second :: tail if first > second => second :: bubbleUpLargestToTheEnd(first :: tail)
+      case first :: second :: tail => first :: bubbleUpLargestToTheEnd(second :: tail)
+      case _ => list
+    }
+
+  def bubbleSort(list: List[Int]): List[Int] =
```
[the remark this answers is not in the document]
jiegillet: You don't need a loop if you want to keep the FP style, I prefer consistent FP style myself.
Your current version is much easier to read. My only nitpick is that your variables names (especially `bubbleUpLargestToTheEnd`) are super long and lines 5 and 6 are too long. Can you reformat to make them <= 80 characters? Unless it's OK to have long lines in Scala.

contents/bubble_sort/code/scala/bubble_sort.scala
```diff
@@ -0,0 +1,22 @@
...
+    bubbleUpLargestToTheEnd(list) match {
+      case unsorted :+ largest => bubbleSort(unsorted) :+ largest
```
jiegillet: Just for curiosity, is the `:+` operator O(n) or O(1)? I ask because in Haskell it would be O(n), which is not ideal.

kenpower: Thanks for your feedback @jiegillet. 

Not sure but I suspect it is O(n) given how the `List` is implemented. I've played around with getting rid of `:+` in the pattern matcher but all my attempts are either less readable/elegant or idiomatic. 

If I were concerned about performance, I would bring back the tail recursive approach from my version. Also we could implement using `Array`, but then the solution starts to look like an imperative approach.

contents/bubble_sort/code/scala/bubble_sort.scala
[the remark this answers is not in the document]
kenpower: Not a nitpick if the line can't fit on the screen!

contents/bubble_sort/code/scala/bubble_sort.scala
jiegillet: I suppose we are really not concerned with performance here, so the more readable, the better.

jiegillet:
I think it's ready to merge! One last thing, you can add your name to `CONTRIBUTORS.md`.

jiegillet:
Yay! Well done :)

PR #420, opened 2018-10-02
Added C++ lang implementation of the split operator

rajdakin:
This pull request is one of the two new pull requests requested in #409 and is made to integrate the C++ lang with the split operator method.


contents/split-operator_method/split-operator_method.md
```diff
@@ -119,8 +119,8 @@ Afterwards, we turn them into operators:
 {% sample lang="jl" %}
 [import:36-62, lang:"julia"](code/julia/split_op.jl)
 {% sample lang="c" %}
-[import:22-28, lang:"c_cpp"](code/c/split_op.c)
-[import:74-95, lang:"c_cpp"](code/c/split_op.c)
+{% sample lang="cpp" %}
```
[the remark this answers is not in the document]
rajdakin: This was simply due to the fact that I didn't pulled #419. This is now fixed.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,249 @@
+#include <complex>
+#include <iostream>
+#include <cstring>
+
+// Using fftw3 library.
+#include <fftw3.h>
+
+class Params {
```
leios: I don't think we have a need for heavily OOP for this case. I think a simple `struct` with everything public is fine.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,249 @@
...
+public:
+    Params(double _xmax, unsigned int _res, double _dt, unsigned int _timesteps, bool im) {
+        xmax = _xmax;
+        res = _res;
+        dt = _dt;
+        timesteps = _timesteps;
+        dx = 2.0 * xmax / res;
+        x = new double[res];
```
leios: In this case, I think a `std::vector` might be more clear than the double values. I know fftw wants arrays, but you can just do `vector.data()`. This is up for debate, though.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,249 @@
...
+    double dt;
+    unsigned int timesteps;
+    double dx;
+    double *x;
+    double dk;
+    double *k;
+    bool im_time;
+};
+
+class Operators {
```
leios: Same criticism as before, I think this is unnecessarily OOP

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,249 @@
...
+        pe = new std::complex<double>[size];
+        ke = new std::complex<double>[size];
+        wfc = new std::complex<double>[size];
+        
+        for (size_t i = 0; i < size; ++i) {
+            v[i] = 0.5 * pow(par.getXs()[i] - voffset, 2);
+            wfc[i] = exp(-pow(par.getXs()[i] - wfcoffset, 2) / 2.0);
+            
+            if (par.isImTime()) {
+                ke[i] = exp(-0.5 * par.getDt() * pow(par.getKs()[i], 2));
```
leios: Here is an example of code that would be much easier to read without OOP

contents/split-operator_method/code/c++/split_op.cpp
berquist: To differentiate between C, `std::vector` should be used.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,249 @@
...
+        dk = M_PI / xmax;
+        k = new double[res];
+        im_time = im;
+        
+        for (size_t i = 0; i < res; ++i) {
+            x[i] = xmax / res - xmax + i * (2.0 * xmax / res);
+            if (i < res / 2) {
+                k[i] = i * M_PI / xmax;
+            } else {
+                k[i] = ((double)i - res) * M_PI / xmax;
```
berquist: The C-style cast should be replaced with `std::static_cast<double>(i)`.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,249 @@
...
+            sum *= par.getDx();
+            
+            for (size_t j = 0; j < opr.getSize(); ++j) {
+                opr.getWfc()[j] /= sqrt(sum);
+            }
+        }
+        
+        // Writing data into a file in the format of:
+        // index, density, real potential.
+        char filename[256];
```
berquist: Many people are against it, but I don't like using C I/O functions in C++, and would replace these lines with `std::ostream`.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: I changed classes to structures.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: I converted pointers to vectors.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: Same as for Params.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: That has been changed.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: I changed that.
By the way, casts are not members of the `std` namespace.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: I tried to use `std::ostream`, but I'm not really satisfied with the result...
Maybe there is an other way?

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,183 @@
...
+        sprintf(filename, "output%lu.dat", i);
+        std::ofstream fstream;
+        fstream.open(filename, std::fstream::out);
+        
+        char buffer[1023];
+        if (!fstream.fail()) {
+            for (int i = 0; i < opr.size; ++i) {
+                snprintf(buffer, 1023, "%d\t%f\t%f\n", i, density[i], real(opr.v[i]));
+                fstream.write(buffer, strlen(buffer));
```
nic-hartley: ...Why? Just do:

```cpp
fstream << i << '\t' << density[i] << '\t' << real(opr.v[i]) << '\n';
```

No need for unchecked C IO and shorter. Maybe less efficient, but more readable, which is the goal here.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,183 @@
...
+    return energy_final * par.dx;
+}
+
+int main() {
+    Params par = Params(5.0, 256, 0.05, 100, true);
+    Operators opr = Operators(par, 0.0, -1.0);
+    
+    split_op(par, opr);
+    
+    printf("The energy is %f\n", calculate_energy(par, opr));
```
nic-hartley: In this case it's... fine, I guess. But I'd personally prefer the C++ way:

```cpp
std::cout << "The energy is " << calculate_energy(par, opr) << '\n';
```

rajdakin: This was fixed.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: Right.
Didn't know about the `std::stringstream`s, that's practical. Thanks for the information!
This was fixed.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,183 @@
...
+    double dk;
+    std::vector<double> k;
+    bool im_time;
+};
+
+struct Operators {
+public:
+    Operators(Params &par, double voffset,
+              double wfcoffset) {
+        size = par.res;
```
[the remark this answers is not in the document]
rajdakin: I tried using a constructor to reserve space, that's why I didn't reserved the space.
Thanks for the info!
This is fixed now.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,183 @@
...
+        
+        // Writing data into a file in the format of:
+        // index, density, real potential.
+        char filename[256];
+        sprintf(filename, "output%lu.dat", i);
+        std::ofstream fstream;
+        fstream.open(filename, std::fstream::out);
+        
+        char buffer[1023];
+        if (!fstream.fail()) {
```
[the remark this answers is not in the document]
rajdakin: I've seen in a website that the `std::ofstream::operator bool` is implemented in `C++11`...
Please confirm?

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,183 @@
...
+                opr.wfc[j] /= sqrt(sum);
+            }
+        }
+        
+        // Writing data into a file in the format of:
+        // index, density, real potential.
+        char filename[256];
+        sprintf(filename, "output%lu.dat", i);
+        std::ofstream fstream;
+        fstream.open(filename, std::fstream::out);
```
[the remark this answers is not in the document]
rajdakin: I wasn't sure about it. This was fixed.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: Yeah, using `std::stringstream`s is a much better way to do it... Thanks.

rajdakin:
@nic-hartley Where do I implicitly use `std::` instead of explicitly use it?

berquist:
I am getting compilation errors:
```
$ g++ -lfftw3 split_op.cpp
split_op.cpp: In constructor ‘Operators::Operators(Params&, double, double)’:
split_op.cpp:61:84: error: missing template arguments before ‘(’ token
                 ke.emplace_back(exp(-0.5 * par.dt * pow(par.k[i], 2) * std::complex(0.0, 1.0)));
                                                                                    ^
split_op.cpp:62:72: error: missing template arguments before ‘(’ token
                 pe.emplace_back(exp(-0.5 * par.dt * v[i] * std::complex(0.0, 1.0)));
                                                                        ^
split_op.cpp: In function ‘double calculate_energy(Params, Operators)’:
split_op.cpp:150:58: error: missing template arguments before ‘(’ token
     std::vector<std::complex<double>> wfc_r = std::vector(opr.wfc);
                                                          ^
split_op.cpp:151:58: error: missing template arguments before ‘(’ token
     std::vector<std::complex<double>> wfc_k = std::vector(opr.wfc);
                                                          ^
split_op.cpp:163:50: error: missing template arguments before ‘(’ token
         energy_k[i] = wfc_k[i] * pow(std::complex(par.k[i], 0.0), 2);
                                                  ^
```

rajdakin:
The problem was not due to the fact that I accessed `wfc_c` before initialization (as it was initialized), but rather to the fact that I assigned values to an address that was not allocated (`wfc_c.reserve(`...`);` fixes this), and there were also more of these problems.
I also needed to use `push_back` as it didn't want to assign values the way I tried to.

leios:
It looks like this problem has been fixed? I just compiled and it worked just fine.

leios:
Would it be possible to put the energy calculation in the quantum systems chapter as well? It's a small modification to what you currently have.

nic-hartley:
> The problem was not due to the fact that I accessed `wfc_c` before initialization (as it was initialized), but rather to the fact that I assigned values to an address that was not allocated (`wfc_c.reserve(`...`);` fixes this), and there were also more of these problems.
> I also needed to use `push_back` as it didn't want to assign values the way I tried to.

Uh, no. `.reserve` doesn't fix it; it's still UB to access beyond the bounds of the vector. You want to `resize`, or better yet, just call the ctor with the desired size. All `reserve` does is keep you from having to reallocate multiple times while `push`ing `back`.

rajdakin:
@berquist I fixed the problems, it works now.
@leios OK, I'm going to put the code there too.
@nic-hartley What do you mean by UB? I tried assign the vector's value using vector`[i] = `...`;` and it "worked" (it only crashed later when doing `fftw_execute` as the vectors were empty).

nic-hartley:
@rajdakin You can also assign and read outside the bounds of a normal array without crashing; that's how Heartbleed worked. It's still UB. All `reserve` does is increase the _capacity_, which is the number of elements the `vector` can hold without needing to reallocate; you should still `push_back` into a reserved vector. Or you can call the constructor, set its initial size, and assign in. Either one will work.

leios:
Hey, what is the status on this PR? Once this is merged, we can attack #476 . I would like to point out that the code seems to output that the energy is 0 right now, which shouldn't be correct. I can look into this more, so let me know if you want me to do this.

As a note: I am not sure what the best solution to the Undefined Behaviour (UB) is, but I can also take a swing at that, if desired.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+double calculate_energy(Params par, Operators opr) {
+    std::vector<std::complex<double>> wfc_r(opr.wfc);
+    std::vector<std::complex<double>> wfc_k(opr.wfc);
+    std::vector<std::complex<double>> wfc_c(opr.size);
+    fft(wfc_k, opr.size, false);
+    
+    for (size_t i = 0; i < opr.size; ++i) {
+        wfc_c.push_back(conj(wfc_r[i]));
```
berquist: Everywhere in `calculate_energy` that you call `push_back` will not do what you expect, because of the preallocation of every `std::vector` here. Switch to indexing just like the C implementation.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,191 @@
...
+    std::vector<std::complex<double>> wfc;
+};
+
+void fft(std::vector<std::complex<double>> x, int n, bool inverse) {
+    std::complex<double> y[n];
+    memset(y, 0, sizeof(y));
+    fftw_plan p;
+    
+    p = fftw_plan_dft_1d(n, reinterpret_cast<fftw_complex*>(x.data()), reinterpret_cast<fftw_complex*>(y),
+                         (inverse ? FFTW_BACKWARD : FFTW_FORWARD), FFTW_ESTIMATE);
```
berquist: The first line is too long. Put each cast on a separate line.

berquist:
There's also an incredible amount of whitespace in the file that needs to be removed.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,191 @@
...
+                
+                fstream.write(data_stream.str().c_str(), data_stream.str().length());
+            }
+        }
+        
+        fstream.close();
+    }
+}
+
+double calculate_energy(Params par, Operators opr) {
```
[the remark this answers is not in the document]
rajdakin: This is fixed.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+    
+    size_t size;
+    std::vector<std::complex<double>> v;
+    std::vector<std::complex<double>> pe;
+    std::vector<std::complex<double>> ke;
+    std::vector<std::complex<double>> wfc;
+};
+
+void fft(std::vector<std::complex<double>> x, int n, bool inverse) {
```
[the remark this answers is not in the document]
rajdakin: This is fixed.

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: What does it changes? `push_back` increases the vector's size by one, and not its capacity, so why would `reserve` (that only change the vector's capacity) change anything?

contents/split-operator_method/code/c++/split_op.cpp
rajdakin: This is fixed.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,191 @@
...
+                pe.emplace_back(exp(-0.5 * par.dt * v[i]));
+            } else {
+                ke.emplace_back(exp(-0.5 * par.dt * pow(par.k[i], 2) * std::complex<double>(0.0, 1.0)));
+                pe.emplace_back(exp(-0.5 * par.dt * v[i] * std::complex<double>(0.0, 1.0)));
+            }
+        }
+    }
+    
+    size_t size;
+    std::vector<std::complex<double>> v;
```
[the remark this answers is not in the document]
rajdakin: This is fixed.

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+struct Params {
+    Params(double _xmax, unsigned int _res, double _dt, unsigned int _timesteps, bool im) {
+        xmax = _xmax;
+        res = _res;
+        dt = _dt;
+        timesteps = _timesteps;
+        dx = 2.0 * xmax / res;
+        x.reserve(res);
+        dk = M_PI / xmax;
```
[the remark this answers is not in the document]
rajdakin: Right. I use Linux, and it is apparently automatically defined, but this is platform-dependant, that's why. This is now fixed.

rajdakin:
@berquist I also removed all the unused whitespaces.

contents/split-operator_method/code/c++/split_op.cpp
berquist: When you do `std::vector<double> myvector(23)`, that is allocating space for 23 doubles into the vector (what they are is UB, but it doesn't matter). When you call `myvector.push_back(30.2)`, that number is accessed at `myvector[23]`, _not_ `myvector[0]`. If you are working with a vector that already has allocated space, you must set the elements directly or use an STL iterator function.

http://www.cplusplus.com/reference/vector/vector/vector/

contents/split-operator_method/code/c++/split_op.cpp
```diff
@@ -0,0 +1,201 @@
...
+        k.reserve(res);
+        im_time = im;
+
+        for (size_t i = 0; i < res; ++i) {
+            x.push_back(xmax / res - xmax + i * (2.0 * xmax / res));
+            if (i < res / 2) {
+                k.push_back(i * M_PI / xmax);
+            } else {
+                k.push_back((static_cast<double>(i) - res) * M_PI / xmax);
```
berquist: I would prefer you be consistent with your use of `push_back` vs. `emplace_back`. There shouldn't be much practical difference here, so change all of those to `emplace_back`.

rajdakin: Right, I thought I never used `emplace_back`. By the way, I think that `push_back` is here more recommanded, so I used `push_back`s instead of `emplace_back`s as `push_back` push the value to the end whereas `emplace_back` creates a new value, which we don't necessarily want here.

leios:
It looks like this review is mostly done. After this, I'll double check that the energy calculation is consistent in #476 and attack that next.

contents/split-operator_method/split-operator_method.md
```diff
@@ -140,6 +144,8 @@ The final step is to do the iteration, itself.
 [import:65-112, lang:"julia"](code/julia/split_op.jl)
 {% sample lang="c" %}
 [import:98-148, lang:"c_cpp"](code/c/split_op.c)
+{% sample lang="cpp" %}
+[import:82-190, lang:"c_cpp"](code/c++/split_op.cpp)
```
berquist: I'd like this to only be the function that does the iteration, `split_op`, in order to be consistent with the other implementations:
```diff
diff --git a/contents/split-operator_method/split-operator_method.md b/contents/split-operator_method/split-operator_method.md
index 1d50c47..ff74f99 100644
--- a/contents/split-operator_method/split-operator_method.md
+++ b/contents/split-operator_method/split-operator_method.md
@@ -145,7 +145,7 @@ The final step is to do the iteration, itself.
 {% sample lang="c" %}
 [import:98-148, lang:"c_cpp"](code/c/split_op.c)
 {% sample lang="cpp" %}
-[import:82-190, lang:"c_cpp"](code/c++/split_op.cpp)
+[import:100-157, lang:"c_cpp"](code/c++/split_op.cpp)
 {% sample lang="py" %}
 [import:57-95, lang:"python"](code/python/split_op.py)
 {% sample lang="hs" %}
```

rajdakin: This is fixed.

leios:
Hey @berquist sorry to bug you, but the changes also seem to be made. Is this one good to go?

PR #427, opened 2018-10-03
add Scala implementation for Euclidean

kenpower:


contents/euclidean_algorithm/code/scala/euclidean.scala
```diff
@@ -0,0 +1,20 @@
...
+    b match{
+      case 0 => a
+      case _ => euclid_mod(b, a % b)
+    }
+
+  def main(args: Array[String]): Unit =
+    println(euclid(135749,27482))
+    println(euclid_mod(135749,27482))
+
+}
```
jiegillet: End the file on a newline

contents/euclidean_algorithm/code/scala/euclidean.scala
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+  def euclid_mod(a: Int, b: Int):Int =
+    b match{
+      case 0 => a
+      case _ => euclid_mod(b, a % b)
+    }
+
+  def main(args: Array[String]): Unit =
+    println(euclid(135749,27482))
```
jiegillet: Could you write these numbers as their prime decomposition? Just so its obvious what the answer should be.

contents/euclidean_algorithm/code/scala/euclidean.scala
```diff
@@ -0,0 +1,20 @@
+object Euclid {
+
+  def euclid(a: Int, b: Int):Int =
+    (a,b) match{
```
jiegillet: Most other implementations take the absolute value of the input to avoid problems with negative numbers. Maybe you could have a wrapper that takes care of that, and also the case when one number is zero while we're at it.

kenpower: Thanks, forgot about that.

contents/euclidean_algorithm/code/scala/euclidean.scala
```diff
@@ -0,0 +1,22 @@
+object Euclid {
+
+  def euclid(a: Int, b: Int): Int =
+    (Math.abs(a), Math.abs(b)) match {
```
jiegillet: This doesn't actually change the values of `a` and `b` so if they are negative, you might still end up with an infinite loop.

contents/euclidean_algorithm/code/scala/euclidean.scala
```diff
@@ -0,0 +1,22 @@
...
+      case (0, _) | (_, 0) => 0
+      case (x, y) if x < y => euclid(x, y - x)
+      case (x, y) if x > y => euclid(x - y, y)
+      case _ => a
+    }
+
+  def euclid_mod(a: Int, b: Int): Int =
```
jiegillet: You should also have a wrapper here that checks for `b=0` and call  `euclid_mod(abs(a),abs(b))`.

contents/euclidean_algorithm/code/scala/euclidean.scala
```diff
@@ -0,0 +1,22 @@
+object Euclid {
+
+  def euclid(a: Int, b: Int): Int =
```
jiegillet: Could you rename it to `euclid_sub`?

contents/euclidean_algorithm/code/scala/euclidean.scala
kenpower: It doesn't change a & b (they're immutable), but it does change the value passed to x & y, so x & are always positive.

kenpower: But if you think a wrapper is better, I can do it.

contents/euclidean_algorithm/code/scala/euclidean.scala
kenpower: OK. You're concerned about having `Euclid` and `euclid` in same namespace?

contents/euclidean_algorithm/code/scala/euclidean.scala
jiegillet: You're correct, my bad!

contents/euclidean_algorithm/code/scala/euclidean.scala
jiegillet: Not really, it's just more readable when you have 2 methods based on subtraction and modulo

PR #429, opened 2018-10-03
Implementation Thomas algorithm in PHP

PudottaPommin:


PudottaPommin:
@Butt4cak3 good to go

Butt4cak3:
As I said, I don't know how this algorithm works or what it does, so someone else will have to review it.

PR #431, opened 2018-10-03
Added intergration of Monte Carlo for PHP

PudottaPommin:


contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
...
+  }
+
+  return 4 * $in_circle_count / $samples;
+}
+
+$piEstimate = monte_carlo(100000000);
+$percentError = abs($piEstimate - pi()) / pi() * 100;
+echo sprintf('The estimate of PI is: %s', $piEstimate);
+echo PHP_EOL;
+echo sprintf('The percent error is: %s', $percentError);
```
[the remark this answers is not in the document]
MichalHanajik: ?> is discouraged and not needed since PHP 5.3 ( I think that version, not sure if 2/3/4 ).
I haven't left newline in any of previous PHP files, so this is with same style and format

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
...
+declare(strict_types=1);
+
+function in_circle(float $positionX, float $positionY, float $radius = 1): bool
+{
+  return pow($positionX, 2) + pow($positionY, 2) < pow(abs($radius), 2);
+}
+
+function random_zero_to_one(): float
+{
+  return mt_rand(0, 100) / 100;
```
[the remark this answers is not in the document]
MichalHanajik: Yes, there is. We want number in interval <0,1> so I'm getting int in range <0,100> and dividing 100 to get required number

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
...
+    if (in_circle(random_zero_to_one(), random_zero_to_one(), $radius))
+      $in_circle_count++;
+  }
+
+  return 4 * $in_circle_count / $samples;
+}
+
+$piEstimate = monte_carlo(100000000);
+$percentError = abs($piEstimate - pi()) / pi() * 100;
+echo sprintf('The estimate of PI is: %s', $piEstimate);
```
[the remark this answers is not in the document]
MichalHanajik: 1. it has, but there is not much difference and sprintf gives easier way to extract variable
2. %f would generate X.XXXXX format when %s givex X.XX (depending on number of numbers after dot)
3. PHP_EOL is constant depending on environment. from SO "PHP_EOL represents the endline character for the current system", so it's much prefered to use it

MichalHanajik:
@berquist I know it can be set, but there is not any setting for PHP in editorconfig.
I personally very much dislike 4 spaces indent (or 1 tab wide 4 spaces). It just takes too much space in a bit deeper levels/conditions...

About the newline. What's the benefit of that in PHP file?

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
...
+{
+  return mt_rand(0, 100) / 100;
+}
+
+function monte_carlo(int $samples, float $radius = 1): float
+{
+  $in_circle_count = 0;
+
+  for ($i = 0; $i < $samples; $i++) {
+    if (in_circle(random_zero_to_one(), random_zero_to_one(), $radius))
```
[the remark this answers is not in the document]
MichalHanajik: Haven't seen any of the other implementations use radius to multiply X or Y. So what's different in this?

berquist:
I was sort of prompting you to add the correct settings to the .editorconfig file, because I don't know what they should be, and you probably do :)

See here for why adding a newline may be considered important: https://stackoverflow.com/q/5813311 (again, EditorConfig will do this automatically when the plugin is on).

MichalHanajik:
1st merge #442 , then I modify code style :)

contents/monte_carlo_integration/code/php/monte_carlo.php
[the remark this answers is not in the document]
Gorzoid: That results in only 101 posible values 0.000, 0.001, 0.002, ..., 0.999, 1.0, running over 101^2 times would then be meaningless, instead you should take advantage of the full range of floating point precision by dividing by mt_getrandmax()

contents/monte_carlo_integration/code/php/monte_carlo.php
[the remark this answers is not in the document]
Gorzoid: Other implementations chose for radius to be a constant 1.0, you put radius as a function parameter which default value 1.0
Either you can make radius constant 1.0 or implement it so it correctly calculates pi when the radius parameter is not equal to 1.0
Take a look at the python implementation, they generate random numbers from 0 to radius. The easiest way to do this is just multiply your existing random function by $radius.

contents/monte_carlo_integration/code/php/monte_carlo.php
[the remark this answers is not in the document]
MichalHanajik: I see, it's changed, thanks for explanation :)

contents/monte_carlo_integration/code/php/monte_carlo.php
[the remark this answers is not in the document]
Gorzoid: Fair enough, I don't understand what you mean by "sprintf gives easier way to extract variable" but your other arguments make sense, I was under the impression PHP would \n to the correct line ending like C/C++.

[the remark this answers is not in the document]
MichalHanajik: it's basically from my experience, where you don't print to console, but you need to work with strings, build them, etc. You can store sprintf to variable and work with it later. printf would print to output. 

yes, it would but with PHP_EOL. otherwise if you specify \n it uses that, which might be problem on some platforms.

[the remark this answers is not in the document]
Butt4cak3: > You can store sprintf to variable and work with it later.
You do not do that, though, so there's no reason to use it. I have to agree with @Gorzoid here. `printf()` makes a little more sense.

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
+<?php
+declare(strict_types=1);
+
+function in_circle(float $positionX, float $positionY, float $radius = 1): bool
```
Butt4cak3: I'm not sure I like snake_case in PHP here. I'm aware of the fact that you already submitted other algorithms with the same code style, but I wasn't here for those.

Is there any style guide you're sticking to or is this just complete personal preference? The PHP code I usually deal with uses camelCase for functions/methods and variables. I'm looking at big projects like Symfony and Laravel here. I believe the AAA should always stick to the most used style for a language, if possible.

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    return pow($positionX, 2) + pow($positionY, 2) < pow($radius, 2);
```
Butt4cak3: Adding to what I said in the above comment, you used camelCase here (`$positionX` instead of `$position_x`) and in other places. I recommend we stay consistent, whether it's camel- or snake-case, but I must say I'd prefer to go for camel-case.

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
+<?php
+declare(strict_types=1);
```
Butt4cak3: I don't think we need this here. I know what it does and I know that it's usually a good idea if you use type hinting in PHP 7, but we had a similar discussion with Python already. Simplicity and readability is very important for the AAA and introducing clutter that is not important to the algorithms them is undesirable.

On the other hand, we have to find a good balance between simplicity and best-practice and this could definitely be considered best practice. I'm kind of torn apart here and I'd like to hear your (and other's) opinion on this.

MichalHanajik: Here's some [explanaition](https://stackoverflow.com/questions/48723637/what-does-strict-types-do-in-php). Basically it enforces correct types to be used when they are set. It's better for type safety , otherwise PHP just tries to convert passed value and eventually fails.

contents/monte_carlo_integration/code/php/monte_carlo.php
MichalHanajik: It's easy. Functions which are not defined inside of class are snake_case, because if you look, pretty much every php function is snake_case. Functions inside classes should be per PSR/Code style you set for the project.

If I define class and functions inside, I agree with PSR-2 or Symfony style, but if it's global function like in this case, PHP style of naming is preferred, via my point of view, so it's consistent with PHP functions.

contents/monte_carlo_integration/code/php/monte_carlo.php
MichalHanajik: Regarding naming parameters there is no definitive answer. I never found any exact order or style in any of PSR standards. They look at classes, methods but never found anything about variables. 

But I agree that `$position_x` can be better in this case. camel_case are better for local function calls and passing arguments and pascalCase for class properties.

contents/monte_carlo_integration/code/php/monte_carlo.php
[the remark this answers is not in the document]
MichalHanajik: I give you this one, I'm gonna change it. But still I'm staying, that sprintf looks better 🤣

contents/monte_carlo_integration/code/php/monte_carlo.php
Butt4cak3: As I said, I know what it does and why it's good. I just want to know whether you have a good reason for why we should use it in the archive, or in other words: Does it help with explaining or understanding the algorithm in question? Because that's what we're about. We're not writing code for production.

contents/monte_carlo_integration/code/php/monte_carlo.php
Butt4cak3: Okay. I figured this would be your answer and I understand where you're coming from. PHP is pretty inconsistent when it comes to function naming, but most function names are indeed snake case.

contents/monte_carlo_integration/code/php/monte_carlo.php
Butt4cak3: The PSRs have a few examples in which they use camel case for variable names and code of most projects I know use camel case as well. My suggestion would be to go with camel case for all variable names.

MichalHanajik: If you check [closures](https://www.php-fig.org/psr/psr-2/#6-closures) it's mixed... but yes, I can agree on camelcase for everything except class properties

contents/monte_carlo_integration/code/php/monte_carlo.php
MichalHanajik: I agree, that it's not production code. Question is, in which way that one declare makes code less readable? You can argue that almost anything improves readability. Same the other way.

I'm definitely for keeping that if we are gonna use PHP 7. I consider this line basically like imports in any other language. Does improve readability?

Gorzoid: As someone who doesn't use PHP that often and never knew of this feature, I think it should be kept, rather than making it hard to read it only brought to light a technique to improve your code.

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,30 @@
...
+
+function in_circle(float $positionX, float $positionY, float $radius = 1): bool
+{
+  return pow($positionX, 2) + pow($positionY, 2) < pow(abs($radius), 2);
```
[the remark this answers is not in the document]
Gorzoid: This still hasn't been resolved btw.

contents/monte_carlo_integration/code/php/monte_carlo.php
Butt4cak3: Alright. Let's keep it. It's not that important.

contents/monte_carlo_integration/code/php/monte_carlo.php
Butt4cak3: I would agree on camel case for everything except for function and variable names. I'd keep class properties camel case to stay with the theme of camel case for everything that's part of a class instance. Variables are never accessible from the outside anyway, so name them however you please. But please, stay consistent. Either use camel case or snake case for your variables, not both.

MichalHanajik:
@Butt4cak3 so my proposition:
- PSR-2 standard
- global functions (set outside of classes) will be camel_case
- variables will be camel_case, but class properties will be pascalCase

if yes, I write it to the editorconfig PR and modify other PR based on our resulted styled.

Butt4cak3:
Agreed.

PudottaPommin:
as we talked, this should be ready for final review @Butt4cak3

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,31 @@
...
+{
+    return mt_rand() / mt_getrandmax();
+}
+
+function monte_carlo(int $samples, float $radius = 1): float
+{
+    $inCircleCount = 0;
+
+    for ($i = 0; $i < $samples; $i++) {
+        if (in_circle(random_zero_to_one() * $radius, random_zero_to_one() * $radius, $radius))
```
Butt4cak3: Taken from PSR-2:

> The body of each structure MUST be enclosed by braces. This standardizes how
the structures look, and reduces the likelihood of introducing errors as new
lines get added to the body.

contents/monte_carlo_integration/code/php/monte_carlo.php
```diff
@@ -0,0 +1,31 @@
+<?php
+declare(strict_types=1);
+
+function in_circle(float $positionX, float $positionY, float $radius = 1): bool
+{
+    return pow($positionX 2) + pow($positionY, 2) < pow($radius, 2);
```
Butt4cak3: Missing comma between `$positionX` and `2`.

PR #434, opened 2018-10-03
Add Monte Carlo implementation in Factor

nic-hartley:
@jiegillet You might be interested in this, too ;)

By the way -- I got a little annoyed of writing the same Factor config code into book.json. If you merge the other Factor issues first it should be fine.


nic-hartley:
retrigger

nic-hartley:
@jiegillet I think you need to read the code more carefully, if that's really what you think. Let me draw your attention to one part specifically:

```factor
xr random yr random
```

That randomly samples a value from the X range and a value from the Y range. It's not just iterating over the entire grid; to prove it, try running this on your machine:

```factor
-1000000000 1000000000 [ >bignum ] bi@ [a,b] dup
100
[ ! in-circle check
  2 ^ swap 2 ^ +   ! get the distance from the center
  1000000000 2 ^ < ! see if it's less than the radius
]
monte-carlo
```

You'll notice it runs nearly instantly, despite supposedly needing to traverse the _entire_ quintillion points. It would take a billion gigahertz CPUs to do that in a full second if there was a one-operation way to do a Monte Carlo sampling, so that's a really impressive machine you have there!

Well, that, or it's randomly sampling 100 points from the field. But I dunno, either seems likely.

I'll close by saying two things:

1. You can run this integration equally well with `10 >bignum 50 ^ dup -1 * swap [a,b] dup` as your input ranges, which makes this _more_ precise than single-precision floats. In fact, you can make it as precise as you want.
2. There's no built-in way to generate floats from a range in Factor. No, not even `1.0 -1.0 [a,b] random`; that'll only ever return `1.0` or `-1.0`. Not even `0`.

jiegillet:
I stand corrected, you code doesn't do what I thought it did.
But it's still quite complicated. Looking at the documentation I found a function that generate a float from a range: `0 1 uniform-random-float`. I think you can simplify the code a lot using that.

nic-hartley:
@jiegillet We're supposed to write idiomatic code, right? The single biggest reason to use a concatenative language, and one of the main tenets of Factor's philosophy, is that you can and should refactor code into chunks as small as you can. So, for example, a function to find pi with the Monte Carlo approximation should _call_ a generic Monte Carlo function. I'm happy to do that; I'm not happy to mix in pi calculations with a generic area ratio approximation algorithm. For example:

```
: monte-carlo-pi ( -- pi )
  -1 1 [a,b] dup 1000 [ [ 2 ^ ] bi@ 1 < ] monte-carlo ;
```

(give or take some syntax/vocab errors)

And, honestly, that's what I think the rest should look like, too -- the page is still just about Monte Carlo, with the pi as an instructive example. The code should be like that, too. But that's a discussion for another time :)

contents/monte_carlo_integration/code/factor/monte_carlo.factor
```diff
@@ -0,0 +1,26 @@
...
+! Use the monte-carlo approximation to calculate pi
+: monte-carlo-pi ( n -- pi-approx )
+  [ ! in-circle check
+    [ 2 ^ ] bi@ + ! get the distance from the center
+    1 <           ! see if it's less than the radius
+  ]
+  monte-carlo 4 * >float
+;
+
+USING: formatting calendar ;
```
jiegillet: We don't really need timing functions, its not part of the algorithm. Also the complexity is clearly linear so there shouldn't be any surprise.

contents/monte_carlo_integration/code/factor/monte_carlo.factor
```diff
@@ -0,0 +1,26 @@
...
+! Add timing, so we can see how long things take.
+:: time-quot ( quot -- quot' )
+  [ now quot dip now swap time- duration>milliseconds "Took %dms\n" printf ]
+; inline
+
+! Actually calculate pi to 5 levels of precision.
+5 <iota> [
+  10 swap ^ dup "%d iterations\n" printf monte-carlo-pi "pi: %f\n" printf
```
jiegillet: Just do a single run with 10^6 or something like that.
Could you also calculate the error with respect to pi? Look at other implementations for the details.

contents/monte_carlo_integration/code/factor/monte_carlo.factor
```diff
@@ -0,0 +1,26 @@
...
+  monte-carlo 4 * >float
+;
+
+USING: formatting calendar ;
+! Add timing, so we can see how long things take.
+:: time-quot ( quot -- quot' )
+  [ now quot dip now swap time- duration>milliseconds "Took %dms\n" printf ]
+; inline
+
+! Actually calculate pi to 5 levels of precision.
```
jiegillet: estimating pi with 10^5 samples does not mean 5 levels of precision.

contents/monte_carlo_integration/code/factor/monte_carlo.factor
nic-hartley: Whoops! That's a holdover from some debugging stuff, I forgot to tear it out. (I ran it with ~10000000 iterations to see how fast each one was, and see how precise I could get)

contents/monte_carlo_integration/code/factor/monte_carlo.factor
nic-hartley: No, but we're calculating pi to 5 distinct levels of precision, which is what the comment means.

nic-hartley:
@jiegillet "A generic Monte Carlo method" doesn't mean "A method which can perfectly handle all possible Monte Carlo approximations", it means "a method not specifically tied to a single solution when the prose explicitly states that you can substitute out one function for something else and do something different".

contents/monte_carlo_integration/code/factor/monte_carlo.factor
nic-hartley: whoops, forgot about the second half of this, adding it now

contents/monte_carlo_integration/code/factor/monte_carlo.factor
```diff
@@ -0,0 +1,21 @@
...
+    [ 2 ^ ] bi@ + ! get the distance from the center
+    1 <           ! see if it's less than the radius
+  ]
+  monte-carlo 4 * >float
+;
+
+USING: math.constants ;
+10000000 monte-carlo-pi ! Approximate pi
+dup .                   ! Print the approximation
+pi - pi / >float abs .  ! And the error margin
```
jiegillet: Multiply by 100 to get percents

nic-hartley: Shoot, missed this comment. I'll add it at work today. Sorry about that.

jiegillet:
```suggestion
pi - pi / 100 * >float abs .  ! And the error margin
```

PR #438, opened 2018-10-03
Add gaussian elimination in Java

PaddyKe:


contents/gaussian_elimination/code/java/GaussianElimination.java
```diff
@@ -0,0 +1,111 @@
...
+		a[i] = a[pivot];
+		a[pivot] = temp;
+	}
+
+	public static void main(String[] args) {
+		double[][] a = {
+				{3, 2, -4, 3},
+				{2, 3, 3, 15},
+				{5, -3, 1, 14}
+		};
```
Yordrar: All methods access the array a, you can make it a member of the class so that all methods can access it directly without passing it as parameter

nic-hartley: @Yordrar I disagree. Wrapping things up in an object would just obfuscate what's actually being done, and making the methods access a field would make the code less reusable.

Yordrar: If we don't give good examples of encapsulation and good OOP practices then why are we supporting java in the first place? we can just copy the code from C or Python and change it slightly. Besides, if you want to use this code in an already existing project, you would have to put it as a member anyway (if you don't want to make badly designed code ofc)

nic-hartley: @Yordrar You're right. We should give *good* examples of encapsulation. This is not one of them. Wrapping up an array in a class just so we can call instance methods on `this` instead of passing it directly into static methods is not good design.

Yordrar: No, i didn't mean wrapping up the array in a separate class, i meant putting it as a member of the class that has the main and all the methods (the GaussianElimination class) for direct access without "this" or calling anything

PudottaPommin: @Yordrar 
There are few things which implementations should have.
Most important is to implement them without mistakes in algorithm (logic).
After that should be readability and simpleness. Keep it easy to understand.
Good programming habits based on language, some are OOP, some not. 

In this case. 
What would be benefit of the matrix as member of class? 
Does it help to understand algorithm? 

Tell me from your point, how person, which wants to try Java example can use different input matrix. Does make more sense to modify member of class instead just changing variable which you send into method?

These are my questions and concerns. I'm open to different views and standpoints so don't take this as an personal attack or something wrong.

contents/gaussian_elimination/code/java/GaussianElimination.java
```diff
@@ -0,0 +1,111 @@
...
+	
+	static void swapRow(double[][] a, int i, int pivot) {
+		double[] temp = a[i];
+		a[i] = a[pivot];
+		a[pivot] = temp;
+	}
+
+	public static void main(String[] args) {
+		double[][] a = {
+				{3, 2, -4, 3},
```
Butt4cak3: These are indented twice. Once you changed the indentation to 4 spaces, make sure that they're only indented once like this:
```Java
double[][] a = {
    {3, 2, -4, 3},
    {2, 3, 3, 15},
    {5, -3, 1, 14}
};
```

contents/gaussian_elimination/code/java/GaussianElimination.java
```diff
@@ -0,0 +1,111 @@
...
+				{2, 3, 3, 15},
+				{5, -3, 1, 14}
+		};
+		
+		gaussianElimination(a);
+		System.out.println("Gaussian eleimination:");
```
Butt4cak3: Typo (eleimination)

contents/gaussian_elimination/code/java/GaussianElimination.java
PaddyKe: I created an issue to discuss this problem a little further. #489

Yordrar: @PudottaPommin To be fair, no, it doesn't matter a lot having it as a variable or as a member of the class in terms of readability, i just want to have code in this project to be of the maximum quality possible (in my opinion at least, seeing that in this case some of you disagree with me) but if you guys think that it is fine as it is, so be it :)

Butt4cak3:
Very nice job! Only one review cycle - that doesn't happen too often. :D

contents/gaussian_elimination/code/java/GaussianElimination.java
nic-hartley: @Yordrar Then please don't advocate for using global variables to pass data to parameters. You want high-quality code, right? That's not how you get it.

PR #440, opened 2018-10-04
Markdown formatting updates

berquist:
- Use a normal list in CONTRIBUTING, rather than HTML
- Update license copyright year
- Unfill text that doesn't change presentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,5 @@
 # Contributing
 
-A contribution guide on how to contribute to the Arcane Algorithm Archive (AAA) can be found on this Wiki page:
-https://github.com/algorithm-archivists/algorithm-archive/wiki/How-to-Contribute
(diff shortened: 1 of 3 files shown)


leios:
These are all good changes. I cannot remember the purpose of the `<br>` tags, though...

jiegillet:
Without the `<br>` all names were on the same line.

leios:
Ah. Alright. I prematurely merged here then. Been a while since that happened.

So, we can just put a newline between each one instead of the `<br>` right?

jiegillet:
No no, it should be good now, the PR replaced the `<br>` with `-` which should create a markdown list. It's an improvement!

leios:
Ah, I should have checked this in remarkable before merging to be sure. Didn't think about it. Nice.

jiegillet:
You can check in GitHub directly, within the PR, just click on view. Once it's merged in, you can just click on the file.

PR #441, opened 2018-10-04
made change

varunyo:
a better and easy C change

### DIFF
--- a/contents/bubble_sort/code/c/bubble_sort.c
+++ b/contents/bubble_sort/code/c/bubble_sort.c
@@ -1,35 +1,36 @@
+/* Bubble sort code */
+ 
 #include <stdio.h>
-
-void print_range(int *array, size_t n) {
-    for (size_t i = 0; i < n; ++i) {
-        printf("%d ", array[i]);
+ 
+int main()
+{
+  int array[100], n, c, d, swap;
+ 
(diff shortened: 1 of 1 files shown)


leios:
I appreciate the change; however, I fail to see how this is better or easier. Could you elaborate?

It seems to me like the new version you have submitted removed the `bubble_sort()` and `print_range() functions and instead placed everything entirely in `main()`. In this case, keeping these functions seems to be more readable.

In addition, because you are asking for input from the user, it is more difficult to tell or test the code input and output without running the code. This is a book and we focus on readability over interactivity.

Gathros:
I agree with @leios but I would also like to add something important; the point of the bubble sort function is for people to read just the C implementation of bubble sort and to take when they need it for their work.

julianschacher:
Yeah the bubble sort function is existent, so that you can just read the actual code without any Console Print or other clutter. It is also just good practice in general.

Since this PR introduces the mentioned issues, I close it.

varunyo:
I am a beginner my friend so for me it was a bit easier and better.and
thanx for ur support .
btw where're you from?

Gathros:
@varunyo If you want to talk to people more casually, please, join the [discord](https://discord.gg/aPxwjrZ).

PR #444, opened 2018-10-04
Add Monte Carlo integration in Racket

PaddyKe:
If you prefer an (kind of) iterative version of this algorithm, I can add this as well but the recursive algorithm is faster.


contents/monte_carlo_integration/code/racket/monte_carlo.rkt
```diff
@@ -0,0 +1,21 @@
...
+                      count
+                      (monte_carlo* (sub1 n) 
+                                    (if (in_circle (random) (random)) 
+                                        (add1 count)
+                                        count
+                                    )
+                      )
+                  )
+               )) (monte_carlo* n 0)
+        ) n) 4)
```
nic-hartley: You don't need to multiply by four. In fact, you shouldn't. It makes this algorithm wrong.

contents/monte_carlo_integration/code/racket/monte_carlo.rkt
```diff
@@ -0,0 +1,21 @@
...
+  )
+
+(monte_carlo 100000)
```
nic-hartley: for the sake of quick running, this should probably be `(print (monte_carlo 1000))`

contents/monte_carlo_integration/code/racket/monte_carlo.rkt
PaddyKe: I think I have to keep the multiplication by four because the algorithm uses only one quarter of the unit circle to approximate pi. If I remove it, the algorithm returns pi/4.

nic-hartley: ...no, it doesn't. I'm not sure why you think that, but there's literally no way I can explain this without sounding condescending, so:

$A_c$ = area of circle.  
$A_s$ = area of square.  
$$\frac{\frac{A_c}{4}}{\frac{A_s}{4} = \frac{\frac{1}{4} A_c}{\frac{1}{4} A_s} = \frac{A_c}{A_s}$$

nic-hartley: ...no, it doesn't. I'm not sure why you think that, but there's literally no way I can explain this without sounding condescending, so:

$A_c$ = area of circle.  
$A_s$ = area of square.  
$$\frac{\frac{A_c}{4}}{\frac{A_s}{4} = \frac{\frac{1}{4} A_c}{\frac{1}{4} A_s} = \frac{A_c}{A_s}$$

jiegillet: @PaddyKe is right, you need the 4. The factor 4 comes from the fact that the side length of the square is 2 r. A_c = pi r^2 and A_s = 4 r^2 => the ratio is pi/4

nic-hartley: @jiegillet Let's rewrite the article to sound less like we're doing Monte Carlo approximations and more like we're calculating pi with them, then. Seriously -- look at the prose in the article. It's very clearly not going all the way to calculating pi, just using the setup to demonstrate Monte Carlo. If you want to print `pi`, that can equally easily be done after the actual Monte Carlo approximation is done, because it's not part of the approximation.

contents/monte_carlo_integration/code/racket/monte_carlo.rkt
```diff
@@ -15,7 +15,7 @@
                       )
                   )
                )) (monte_carlo* n 0)
-        ) n) 4)
+        ) n)
   )
 
-(display (monte_carlo 1000))
+(display (* 4 (monte_carlo 1000)))
```
PaddyKe: Maybe this is a good compromise between getting pi as result and doing plain Monte Carlo integration?

nic-hartley: Yup! That works for me.

My only concern is that (as it is now) your Monte Carlo algorithm itself is perfect. While the *4 is _necessary_ to get pi out of it, pi, here, is only a side thing -- the main attraction is the Monte Carlo. I want that to be as clear as possible, and the \*4 distracts from that.

PaddyKe: I understand your concerns. I added a comment explaining the multiplication by four.

contents/monte_carlo_integration/code/racket/monte_carlo.rkt
```diff
@@ -0,0 +1,26 @@
+#lang racket
+(define (in_circle x y)
+  (< (+ (sqr x) (sqr y)) 1)
```
jiegillet: Calculating squares is much more efficient than square roots

jiegillet: Oh no! disregard, I though these were `sqrt`

PR #445, opened 2018-10-04
Tree traversal in Smalltalk

Neverik:


contents/bogo_sort/code/smalltalk/bogosort.st
```diff
@@ -0,0 +1,9 @@
+"Add this to the SequenceableCollection: "
```
nic-hartley: Is this tree traversal, bubblesort, or bogosort? Please do one implementation per PR.

PR #446, opened 2018-10-04
Lisp jarvis march

Trashtalk217:
It works, but that is mostly all that can be said about it. I am nowhere near happy with it, but I want people to review it and give some tips. That's also the reason I haven't editted the .md files for Jarvis March. I'll be working on it way more, but for now you can enjoy the spaghetti.


Trashtalk217:
A list of things I'm still not happy with:
    The way I currently calculate angle is nice and compleet, but I'd like to cut (some) corners.
    The `return-from` in line 106 works, but I should just write the loop properly (loop until and then 
    return) instead of jumping out of it like this.
    Could maybe make the pointstruct more convenient (read-only, shorter way to read values, etc)
    Both `leftmost-point` and `next-point-on-hull` are doing something really similar, but that could be 
    abstracted, I'm also getting tired of tail-recursion, but this might have to do.
    The `p-eql` is poorly named and maybe I could somehow integrate it into the point struct, or another 
    comparison function can be used (there are 4 in common lisp!)

The reason why I removed the WIP label, is because I want additional input, besides my own. I expect that there'll be more people looking to review, when it doesn't have this label.

contents/jarvis_march/code/lisp/jarvis-march.lisp
```diff
@@ -0,0 +1,123 @@
...
+          (thetaC 
+            (atan2 
+              (- (point-y p3) (point-y p2)) 
+              (- (point-x p3) (point-x p2))))
+          (theta (- thetaC thetaA)))
+        (cond
+          ((< theta 0) (+ theta (* 2 pi)))
+          (t theta)))))
+
+(defun leftmost-point (gift)
```
c252: You defined left-most point twice

contents/jarvis_march/code/lisp/jarvis-march.lisp
```diff
@@ -0,0 +1,123 @@
...
+                  (second hull)
+                  (first hull)
+                  gift)
+                hull)))))))
+
+(defvar test-gift
+  (map 
+    'list
+    (lambda (e) (apply #'make-point e))
+    '((0 0) (2 1) (0 2)
```
c252: maybe make these values the same as other implementations.

c252: Sorry my review wasn't that helpful.

Trashtalk217:
I've tried, but I don't know how to shorten the functions involved in calculating the angle. My linear algebra is a bit rusty and the way others do it looks (and maybe is) quite different. If anyone can elaborate that would be very much appreciated.

contents/jarvis_march/code/lisp/jarvis-march.lisp
Trashtalk217: Fixed

Trashtalk217:
I've fixed all the things I was unhappy with. All I need now is a review!

contents/jarvis_march/code/lisp/jarvis-march.lisp
```diff
@@ -0,0 +1,63 @@
...
+    start
+    (make-point (point-x start) (- (point-y start) 1))
+    gift))
+
+(defun jarvis-march (gift)
+  "finds the convex hull of any random distribution of points"
+  ;deals with the edge cases
+  (if (< (length gift) 3)
+    gift
+    (loop
```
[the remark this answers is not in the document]
Trashtalk217: I honestly didn't understand your implementation @jiegillet, And since I have the loop macro available it seems fitting to use it here.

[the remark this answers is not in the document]
jiegillet: Fair enough :)

contents/jarvis_march/code/lisp/jarvis-march.lisp
```diff
@@ -0,0 +1,63 @@
...
+
+(defun next-point-on-hull (p1 p2 gift)
+  "Finds the next point on the convex hull of a gift"
+  (if (null gift)
+      p2
+      (if (is-left-p p1 p2 (first gift))
+          (next-point-on-hull p1 (first gift) (rest gift))
+          (next-point-on-hull p1 p2 (rest gift)))))
+
+(defun leftmost-point (gift)
```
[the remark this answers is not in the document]
jiegillet: Oh yeah, much better ^^

contents/jarvis_march/code/lisp/jarvis-march.lisp
```diff
@@ -0,0 +1,59 @@
...
+  "Returns the lefmost point of a gift"
+  (reduce 
+    (lambda (p1 p2)
+      (if (< (point-x p1) (point-x p2)) p1 p2))
+    gift))
+
+(defun second-point-on-hull (start gift)
```
jiegillet: I think I realized you don't need `second-point-on-hull`. Alls it does is call `next-point-on-hull` you might as well do that in your loop.
You can start your hull with the `(- (point-y start) 1)` and `start` points. And then remove the fake point later.

Trashtalk217: Are you sure this is a good idea, because the the first element of the hull, as defined in the hull list, will then become the starting point, so the check after that `until (equalp (first hull) start)` will pass and the loop will end, I cannot find an elegant way around that, do you have an idea?

Trashtalk217:
Honestly don't know if this is an improvement, but it's a couple lines less once more.

Trashtalk217:
I wouldn't have expected anything else from a haskell programmer, @jiegillet. But, jokes aside, I mostly agree with you. There's something very satisfying about getting your code under 100 lines and if you're not cramping everything together it more often then not makes it more readable.

I, however, do not understand what you mean in you're request. Could you please be a little more clear?

jiegillet:
Something like:
```lisp
    (loop 
      with start = (leftmost-point gift)
      with hull = (list start (make-point (point-x start) (- (point-y start) 1)))
      do (setq hull
             (cons 
               (next-point-on-hull (first hull) (second hull) gift)
               hull))
      while (/= (first hull) start) 
```
(although I'm pretty sure my syntax is wrong at more than one place)

Trashtalk217:
I actually can't believe it worked, but it did.

contents/jarvis_march/jarvis_march.md
```diff
@@ -40,6 +40,10 @@ Since this algorithm, there have been many other algorithms that have advanced t
 [import, lang:"python"](code/python/jarvisMarch.py)
 {% sample lang="cpp" %}
 [import, lang:"c_cpp"](code/c++/jarvis_march.cpp)
+{% sample lang="lisp" %}
+[import, lang:"lisp"](code/lisp/jarvis-march.lisp)
+{% sample lang="lisp" %}
```
jiegillet: You imported it twice!

Trashtalk217:
It's finally done I think, thanks for reviewing @jiegillet

PR #447, opened 2018-10-04
Add Euclidean algorithm in Ruby

nic-hartley:
Noticed a little hole in our ~~defenses~~ coverage, and I like Ruby, so have another non-Factor PR.


jiegillet:
One more thing, could you write the test numbers in a factorized form to maximize readability? Ex: 48 => 4 * 12

PR #452, opened 2018-10-05
Added bogo sort for smalltalk

Neverik:


contents/bogo_sort/code/smalltalk/bogosort.st
```diff
@@ -0,0 +1,9 @@
+"Add this to the SequenceableCollection: "
+bogoSort
+    "A simple bogosort."
+    [ self isSorted ] whileFalse: [ 
+        self shuffle.
+    ]
```
nic-hartley: This could _probably_ all be on one line, but there's no real point. The only thing I'd really suggest is adding a `^self.` to the end, but even that's not really necessary.

PR #453, opened 2018-10-05
Tree traversal in smalltalk

Neverik:


contents/tree_traversal/code/smalltalk/tree_traversal.st
```diff
@@ -0,0 +1,101 @@
...
+        	Transcript show: data.
+			Transcript cr.
+			^self.
+		 ].
+	children size = 0 ifTrue: [ 
+        	Transcript show: data.
+			Transcript cr.
+			^self.
+		 ].
+	Transcript show: 'This is not a binary tree!'.
```
nic-hartley: This block of `if`s can be simplified a bit:

```smalltalk
children size > 2 ifTrue: [
  Transcript show: 'This is not a binary tree!'; cr.
  ^self.
]
children size = 2 ifTrue: [
  (children at: 1) dfsInOrderBinaryTree.
]
Transcript show: data; cr.
children size >= 1 ifTrue: [
  (children at: 0) dfsInOrderBinaryTree.
]
^self.

File: contents/tree_traversal/code/smalltalk/tree_traversal.st [hunk 2]
```diff
@@ -0,0 +1,101 @@
...
+
+Node>>data: newData
+	"Data setter"
+	data := newData.
+
+Node>>dfsRecursive
+	"Recursive depth first search."
+	data isNil ifFalse: [ 
+			Transcript show: data.
+			Transcript cr.
```
nic-hartley: You do this a lot:

```smalltalk
Transcript show: something.
Transcript cr.
```

That can, and should, be simplified to this:

```smalltalk
Transcript show: something; cr.
```

I'm not gonna leave a comment everywhere you do it, because that'd be a lot.

contents/tree_traversal/code/smalltalk/tree_traversal.st
```diff
@@ -0,0 +1,101 @@
...
+Node>>bfs
+	"A breadth-first tree search using queues."
+	| queue current |
+	queue := LinkedList with: self.
+  	[ queue size > 0 ] whileTrue: [
+    	current := queue first.
+		queue removeFirst.
+    	Transcript show: current.
+		Transcript cr.
+    	current children collect: [ :child |
```
nic-hartley: Messy indenting here, too; again, please switch to all spaces.

contents/tree_traversal/code/smalltalk/tree_traversal.st
```diff
@@ -0,0 +1,101 @@
...
+			 queue addLast: child
+		].
```
nic-hartley: I think that period is unnecessary.

contents/tree_traversal/code/smalltalk/tree_traversal.st
```diff
@@ -0,0 +1,101 @@
...
+  	]
+
+| test |
+test := Node new: 1 children: { Node new: 2.
```
nic-hartley: A more complex tree structure might be good here. For example:

```smalltalk
test := Node new: 1 children: {
  Node new: 2 children: {
    Node new: 3.
  }
  Node new: 4 children: {
    Node new: 5.
    Node new: 6 children: {
      Node new: 7.
    }
  }
}
```

contents/tree_traversal/code/smalltalk/tree_traversal.st
```diff
@@ -0,0 +1,101 @@
...
+
+Node>>data
+	"Data getter"
+	^ data
+
+Node>>data: newData
+	"Data setter"
+	data := newData.
+
+Node>>dfsRecursive
```
nic-hartley: Since this is supposed to be a **d**epth-**f**irst ***s**earch*, maybe we this should search for something?

```smalltalk
Node>>dfsRecursive: value
  "Perform a depth-first search for the value"
  data = value ifTrue: [
    ^true.
  ]
  children collect: [ :child |
    (child dfsRecursive: value) ifTrue: [ ^true. ]
  ]
  ^false.
```

The same applies to the other dfs/bfs methods; then you can have _separate_ preorder, inorder, and postorder traversal methods.

contents/tree_traversal/code/smalltalk/tree_traversal.st
```diff
@@ -0,0 +1,101 @@
...
+	"Recursive depth first search."
+	data isNil ifFalse: [ 
+			Transcript show: data.
+			Transcript cr.
+		 ].
+	children collect: [ :each | each dfsRecursive].
```
nic-hartley: What exactly is an `each`, and why does `children` contain those, instead of... children? (I'm trying to tell you to rename `each` to `child`, because the parameter isn't representing an "each", it's representing one of the several children)

contents/tree_traversal/code/smalltalk/tree_traversal.st
```diff
@@ -0,0 +1,101 @@
...
+
+Node>>dfsRecursivePostorder
```
nic-hartley: Why not take a [visitor](https://en.wikipedia.org/wiki/Visitor_pattern) and return an array instead of just printing all the elements?

```smalltalk
Node>>dfsPostorder: ary into: visitor
  children inject: ary into: [ :child | child dfsPostorder: visitor ]
  data isNil ifFalse: [ ary add: visitor value: data. ]  
```

contents/tree_traversal/code/smalltalk/tree_traversal.st
Neverik: Ok, sure thing!

contents/tree_traversal/code/smalltalk/tree_traversal.st
Neverik: I did this taking reference from the Python example.

Neverik: Probably printing will be more clear (I guess).

nic-hartley: @Neverik Fair enough! This isn't a necessary change, just an idea.

Neverik:
Ok, I fixed a typo and shortened it.

contents/tree_traversal/code/smalltalk/tree_traversal.st
Neverik: fair enough

jiegillet:
@nic-hartley would you mind reviewing the changes?

nic-hartley:
@Neverik Make sure to fix the merge conflicts, and this can probably be merged then

PR #456, opened 2018-10-05
Monte carlo C fix

Gorzoid:
MonteCarlo in C was not implemented as described by the book. In the book we are told to use only the first quadrant to calculate pi. The C code produces a random range from -1 to +1. While the method stills works I think we should keep code to follow the algorithm as described by the book.

I also removed the incorrect formula where radius != 0.0 and replaced it with code to product random variables from 0 to radius.

Another smaller change was changing the return type to double. The function should only calculate pi and return it, there's no need for it to also print the estimation and how close it was, that can be left up to the main function.


Gathros:
Thanks, looks good.

PR #457, opened 2018-10-05
Monte Carlo in Lua

Gorzoid:
Another short PR, spent enough time reviewing C and PHP implementations so decided I'd write the lua version. Doesn't provide variable radius since it's not really necessary for calculating pi.


Gorzoid:
Rebased my branch so I could pull bubble_sort and bogo_sort files. Also changed huffman but did that on the huffman pr

PR #458, opened 2018-10-06
Monte carlo implementation in scala

kenpower:


contents/monte_carlo_integration/monte_carlo_integration.md
```diff
@@ -71,6 +71,8 @@ each point is tested to see whether it's in the circle or not:
 [import:1-4, lang:"ruby"](code/ruby/monte_carlo.rb)
 {% sample lang="f90" %}
 [import:1-8, lang:"fortran"](code/fortran/monte_carlo.f90)
+{% sample lang="scala" %}
+[import:3-14, lang:"scala"](code/scala/monte_carlo.scala)
```
jiegillet: Oh noooo! `import:3-14` would have been perfect for this algorithm, unfortunately here you only need to import line 3 :(

jiegillet:
Also, you should really use `git rebase -i upstream master` before pushing your branch because your commit history is very hard to follow. Another trick I've had to clean up my master branch use was `git reset --hard upstream/master`

contents/monte_carlo_integration/monte_carlo_integration.md
jiegillet: That, or maybe `import:3-3` I'm not sure. You can try both and then compile the book for yourself and see.

berquist:
As long as it isn't merged, it's not too late. Once you enter interactive rebase, "fixup" all except the first commit to nuke their commit messages ("squash" doesn't do this), save/commit, force push, then you're good.

jiegillet:
Also we are always squishing the commits on GitHub before merging PRs so it doesn't really matter for the end result, it's just easier to follow for the reviewers. And now it's back to one commit so it isn't necessarily easy to read either :)

PR #460, opened 2018-10-06
Stable marriage implementation in Scala

kenpower:


leios:
Hey @kenpower, it looks like your Monte Carlo code is still in this PR. Any way you can get rid of those files?

contents/stable_marriage_problem/code/scala/stable_marriage.scala
```diff
@@ -0,0 +1,50 @@
+import scala.collection.mutable._
+
+object StableMarriage {
+
+  case class Man(name: String, var preferences: List[Woman] = List()){
+    def propose():Unit = preferences match {
+      case woman :: remainingPreferences => {
+        if (woman.prefers(this)) {
+          bachelors ++= woman.fiance
```
jiegillet: I don't like that `bachelors` is a global variable, it should belong to `StableMarriage` somehow.

contents/stable_marriage_problem/code/scala/stable_marriage.scala
```diff
@@ -0,0 +1,50 @@
...
+  c.preferences = List(x, z, y)
+  x.preferences = List(b, a, c)
+  y.preferences = List(c, a, b)
+  z.preferences = List(a, c, b)
+
+  var bachelors = ListBuffer(a, b, c)
+
+  def main(args: Array[String]): Unit = {
+
+    bachelors foreach (b => b.propose())
```
jiegillet: This is the real matching happens, it shouldn't be in `main` it should be in `StableMarriage` somewhere.

contents/stable_marriage_problem/code/scala/stable_marriage.scala
kenpower: You are 100% correct.

contents/stable_marriage_problem/code/scala/stable_marriage.scala
```diff
@@ -0,0 +1,64 @@
...
+    def prefers(man: Man): Boolean =
+      fiance match {
+        case Some(otherMan) => preferences.indexOf(man) < preferences.indexOf(otherMan)
+        case _ => true //always prefer any man over nobody
+      }
+  }
+
+  def findStableMatches(men: Man*): Unit = {
+    bachelors = men.to[ListBuffer]
+    bachelors foreach (b => b.propose())
```
jiegillet: Does this loop adapt dynamically as you add values to `bachelors`?

kenpower: yes... a bit awkward I know. `bachelors` is acting as a LIFO queue. Maybe better in a `while()`?

kenpower: I forgot about the `mutable.queue`, I should use that also.

jiegillet: Yeah, a while loop with a mutable queue would be the most readable solution.

PR #466, opened 2018-10-06
change formatting in euclid alg in nim

c252:
Just going through and fixing the formatting errors in the nim programs. I change the indentation and change temp from a var to a let, since the it didn't need to mutable


berquist:
Sorry for the confusion!

PR #471, opened 2018-10-07
Add Tree Traversal in Emojicode.

julianschacher:


contents/tree_traversal/code/emojicode/tree_traversal.emojic
```diff
@@ -0,0 +1,120 @@
...
+  🖍🆕 id 🔢
+  🖍🆕 children 🍨🐚🌲🍆
+
+  🆕 depth_count 🔢 children_count 🔢 🍇
+    1 ➡️ 🖍id
+    🍨🍆 ➡️ 🖍children
+
+    ↪️ ❎ depth_count ◀️🙌 1❗️ 🍇
+      🔂 i 🆕⏩⏩ 0 children_count❗️ 🍇
+        🐻 children 🆕🌲⭐️ 🤜id ✖️ 10 ➕ i ➕ 1🤛 🤜depth_count ➖ 1🤛 🤜children_count🤛❗️❗️
```
thbwd: No need to wrap `🤜children_count🤛` into `🤜🤛`actually.

contents/tree_traversal/code/emojicode/tree_traversal.emojic
```diff
@@ -0,0 +1,120 @@
...
+      🍉
+    🍉
+  🍉
+
+  🔐 🆕 ⭐️ given_id 🔢 depth_count 🔢 children_count 🔢 🍇
+    given_id ➡️ 🖍id
+    🍨🍆 ➡️ 🖍children
+
+    ↪️ ❎ depth_count ◀️🙌 1❗️ 🍇
```
thbwd: This is shared logic. I’d recommend you extract it into a method.

contents/tree_traversal/code/emojicode/tree_traversal.emojic
```diff
@@ -0,0 +1,120 @@
...
+  💭 Depth-First Search Recursive pre-order
+  ❗️ 🌀 🍇
+    😀 🔡 id 10❗️❗️
+
+    🔂 child children 🍇
+      🌀 child❗️
+    🍉
+  🍉
+
+  💭 Depth-First Search Recursive post-order
```
thbwd: Shouldn’t these be documentation comments? https://www.emojicode.org/docs/reference/documentation.html

contents/tree_traversal/code/emojicode/tree_traversal.emojic
julianschacher: I did it to provide better readability. But you're right, in this case I don't need to wrap it.

julianschacher:
@thbwd Thank you very much for the reviews of the Emojicode PRs! 🙂👍

contents/tree_traversal/code/emojicode/tree_traversal.emojic
```diff
@@ -0,0 +1,122 @@
...
+    😀 🔡 id 10❗️❗️
+  🍉
+
+  📗
+    Depth-First Search Recursive Inorder Binary
+    This assumes only 2 children.
+  📗
+  ❗️ 🍭 🍇
+    ↪️ 🐔 children❗️ ▶️ 2 🍇
+      ↩️↩️
```
julianschacher: @thbwd I have one more question tho. While looking through the documentation, I found that I can also return errors.
I think it would be more appropriate to return an error here, if there are more than two children.

How do I use an error type in this method? I think, the documentation doesn't describe, how to provide an error, when you have a method, that returns nothing.

contents/tree_traversal/code/emojicode/tree_traversal.emojic
Butt4cak3: I'm so sorry, but I just have to point out the irony of doing anything in Emojicode for better readability.

julianschacher: If you read the Emojicode documentation and code in it a bit, you're able to read it fluently pretty fast. At this point you can also improve readability 😄

contents/tree_traversal/code/emojicode/tree_traversal.emojic
thbwd: This is not possible by design. An error is either the expected value or an enum.

If you do not have a return type, there would not be an a non-error type. Therefore we have *either an enum indicating an error or no value at all*, which is clearly what optionals are designed for.

This is also how the file package handles it:

https://www.emojicode.org/docs/packages/files/1f4c4.html#cm📻❗%EF%B8%8F

julianschacher: ❗️ 🍭 🍇

julianschacher: Ok, would the following be a correct implemetation?

Declare an enum to hold an error value:

```
🦃 ⏹ 🍇
  🔘 ⏫

  ❗️ 🔡 ➡️ 🔡 🍇
    ↪️ 🐕 🙌 🆕⏹⏫❗️ 🍇
      ↩️ 🔤The given tree is not binary!🔤
    🍉
    ↩️ 🔤🔤
  🍉
🍉
```

Then rewrite the Depth-First Search Recursive Inorder Binary method to return an optional holding the enum and adjust the returns according to that:

```
  ❗️ 🍭 ➡️ 🍬⏹ 🍇
    ↪️ 🐔 children❗️ ▶️ 2 🍇
      ↩️ 🆕⏹⏫❗️
    🍉

    ↪️ 🐔 children❗️ ▶️ 0 🍇
      🍭🐽 children 0❗️❗️
      😀 🔡 id 10❗️❗️
      🍭🐽 children 1❗️❗️
    🍉
    🙅 🍇
      😀 🔡 id 10❗️❗️
    🍉
    ↩️ 🤷‍♀️
  ```

and then call the method like this:

```
  ️↪️ 🍭tree❗️ ➡️ return 🍇
    😀 🔡return❗❗️️
  🍉
  ```

thbwd: Seems reasonable. Although — looking at this — a proper error handling mechanism for cases like wouldn’t be a bad idea after all. Feel free to open an issue over at the Emojicode repo if you have ideas and time.

julianschacher: Alright, I used your suggested way of error handling for methods without a return value and opened an issue over at the emojicode repo: https://github.com/emojicode/emojicode/issues/112 😀

julianschacher:
@thbwd Are you happy with the code after the latest changes?

julianschacher:
@thbwd Awesome; thank you very much for the review!

PR #472, opened 2018-10-07
Implement bubble sort in Hy

berquist:
Since someone took the first Kotlin implementation from me, Hy is a Lisp that transpiles to Python: http://docs.hylang.org/en/stable/index.html

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -55,6 +55,11 @@ indent_size = 2
 indent_style = space
 indent_size = 2
 
+# Hy
+[*.hy]
+indent_style = space
+indent_size = 2
+
 # Java
 [*.java]
 indent_style = space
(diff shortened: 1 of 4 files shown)


berquist:
So I compared using my `cons` implementation (which is not efficient) with using list concatenation. Here are timings on 1000 random integers.

With `cons`:
```
4.05 s ± 27 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
With list concatenation:
```
3.43 s ± 26.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Compare this to the current Python implementation:
```
30.9 ms ± 549 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
The Racket code took 33262 ms total / 1000 = 33.2 ms/step. Going to mark this WIP until I can figure out how to make this more idiomatic/less slow.

berquist:
@Amaras You forgot this one :)

PR #476, opened 2018-10-07
C++ quantum systems' energy calculations implementation

rajdakin:
As requested in #420, I've implemented the energy calculations in the quantum system chapter.


leios:
Thanks! We'll work on this one once everyone's happy with #420

PR #482, opened 2018-10-07
Verlet integration in ruby

tarunvelli:


jiegillet:
This is not yet a full review, just a note that you misspelled Verlet in the file name.

tarunvelli:
@jiegillet thanks for pointing it out, have updated the PR

PR #483, opened 2018-10-07
Add Monte Carlo Integration in Emojicode.

julianschacher:


leios:
Thanks again @thbwd for the review. We really appreciate it!

PR #491, opened 2018-10-09
add thomas algorithm in javascript

Yordrar:


contents/thomas_algorithm/code/javascript/thomas.js
```diff
@@ -0,0 +1,31 @@
...
+
+console.log("The system,");
+console.log("[1.0  4.0  0.0][x] = [7.0]");
+console.log("[2.0  3.0  5.0][y] = [5.0]");
+console.log("[0.0  3.0  6.0][z] = [3.0]");
+console.log("has the solution:\n");
+
+thomas(a, b, c, x);
+
+for (let i = 0; i < 3; i++)
```
Gustorn: You could use `for (const value of x) {` here. Also, you could just print the array as a whole: `console.log("has the solution: ", x)`

Yordrar: done, went for the console.log("solution: ", x) as it is simpler

Butt4cak3:
I'll take a look at it tomorrow. I didn't have access to my PC for most of today.

Yordrar:
@Butt4cak3 , I corrected it just in case, if we then decide that pure functions shouldn't be compulsory and the code is more readable without a pure function I'll change it again, no prob

Butt4cak3:
Just to clarify: I'm still waiting for that discussion to resolve. I didn't forget your PR. I just don't want us to have to do the work twice.

berquist:
We are coming up on one calendar year since the last comment. Is there anything left to be done?

Butt4cak3:
I'll stop trying to fool myself and others. I'm just not active enough in this community anymore. Any PRs that are currently assigned to me should be handled by someone else. I don't even know what these things are about anymore.

leios:
This has been repurposed into #683

PR #494, opened 2018-10-11
Thomas algorithm in javascript

Ariana1729:
Should I try to like format the original equation based on abcd or just leave it as it is?


Ariana1729:
oh oops someone alr did it

PR #495, opened 2018-10-11
Fixing Julia for Thomas algorithm

leios:
Yeah. That last version of Julia was complete and utter garbage. This fixes that and is modelled after the C code.

### DIFF
--- a/contents/thomas_algorithm/code/julia/thomas.jl
+++ b/contents/thomas_algorithm/code/julia/thomas.jl
@@ -1,28 +1,26 @@
 function thomas(a::Vector{Float64}, b::Vector{Float64}, c::Vector{Float64},
                 d::Vector{Float64}, n::Int64)
 
-    soln = Vector{Float64}(undef, n)
+    x = d
+    c_prime = c
 
     # Setting initial elements
-    c[1] = c[1] / b[1]
-    d[1] = d[1] / b[1]
+    c_prime[1] = c_prime[1] / b[1]
+    x[1] = x[1] / b[1]
 
     for i = 2:n
-        # Scale factor is for c and d
-        scale = 1.0 / (b[i] - c[i-1]*a[i])
-        c[i] = c[i] * scale
-        d[i] = (d[i] - a[i] * d[i-1]) * scale
+        # Scale factor is for c_prime and x
+        scale = 1.0 / (b[i] - c_prime[i-1]*a[i])
+        c_prime[i] = c_prime[i] * scale
+        x[i] = (x[i] - a[i] * x[i-1]) * scale
     end
-
-    # Set the last solution for back-substitution
-    soln[n] = d[n]
 
     # Back-substitution
     for i = n-1:-1:1
-        soln[i] = d[i] - c[i] * soln[i+1]
+        x[i] = x[i] - c_prime[i] * x[i+1]
     end
 
-    return soln
+    return x
 
 end
 
--- a/contents/thomas_algorithm/code/julia/thomas.jl
+++ b/contents/thomas_algorithm/code/julia/thomas.jl
@@ -1,23 +1,47 @@
-function(a::Vector{Float64}, b::Vector{Float64}, c::Vector{Float64},
-         d::Vector{Float64}, soln::Vector{Float64})
-    # Setting initial elements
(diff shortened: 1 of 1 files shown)


Butt4cak3:
This new implementation that returns the solution vector still modifies the `c` vector. Should we just make it so that the algorithm doesn't modify the input vectors at all? The C implementation is an example of how that could happen. It creates an intermediate `double y[size]`.

jiegillet:
+1 on not modifying any input. Functional programming FTW :)

Butt4cak3:
It's not that difficult to change the algorithm to not modify `c` and `d`. Create two more arrays, one is the result and the other is there to be modified instead of `c` and you're good to go.

jiegillet:
We are not questioning the notations `a`, `b` and `c` I think, just the fact that the inputs should not be modified by the algorithm.

leios:
I was bringing up the topic because `a, b, c,` and `d` are really unclear and it's something I thought we could clean up while we are at it. Otherwise, we can do precisely what you guys are saying by adding the `c_prime` and `d_prime` variables (which is common mathematical notation).

jiegillet:
I like `c_prime` and `d_prime`, but the chapter also mentions `x` with is kind of `d_prime_prime`, so maybe just using `c_prime` and `x` would be good.

leios:
`x` is the set of solutions, which we can use `d_prime` for. I'll rework this with `c_prime` and `x`.

contents/thomas_algorithm/code/julia/thomas.jl
```diff
@@ -1,23 +1,45 @@
...
-    d[0] = d[0] / b[0]
-
-    for i = 1:n
-        # Scale factor is for c and d
-        scale = 1 / (b[i] - c[i-1]*a[i])
-        c[i] = c[i] * scale
-        d[i] = (d[i] - a[i] * d[i-1]) * scale
-    end
+    c_prime[1] = c_prime[1] / b[1]
+    x[1] = x[1] / b[1]
```
berquist: I usually suggest things like `a = a / b` get converted to `a /= b`, but what do you think? No opinion here.

contents/thomas_algorithm/code/julia/thomas.jl
```diff
@@ -1,23 +1,45 @@
-function(a::Vector{Float64}, b::Vector{Float64}, c::Vector{Float64},
-         d::Vector{Float64}, soln::Vector{Float64})
+function thomas(a::Vector{Float64}, b::Vector{Float64}, c::Vector{Float64},
+                d::Vector{Float64}, n::Int64)
+
+    x = d
+    c_prime = c
```
berquist: These need to be
```julia
x = copy(d)
c_prime = copy(d)
```

berquist: (otherwise you're still overwriting the input.)

berquist:
I was not here for the modifying input vs. returning conversation, but my 2 cents:

I don't think the average numerical algorithm is more understandable by being made pure, and most people who want to understand the algorithm so they can implement it should rather implement the version that mutates a preallocated input argument. Maybe it should go as far as using other input arguments as scratch, as this and other implementations used to. This is how many calls to BLAS and LAPACK work. Passing matrices around with copy-by-value semantics is a death sentence for performance.

Perhaps the solution is to say "there should be no surprises". This is well-documented behavior in real code, but not in the AAA, so being pure is good. Sorry for the inconclusive language, but I never read this rebuttal against "make everything more functional" even though writing pure code prevents necessary memory optimizations in numerical code.

/rant over

TL;DR looks good.

leios:
I agree with @berquist , though. I am not sure what is the best way to do this. I know we are returning values for readability, but in some languages, this does kill performance and code shouldn't be written in general like that.

jiegillet:
We were never concerned with performance though, so I'm not worried about that. 
I'm just thinking a function like this one should be pure in case you need to use your matrix for other purposes later :)

berquist:
That is possible, yes. It is also possible that the people who would be affected already know about this sort of thing. I am fine with making everything pure, but only pointing out something I have never seen written down that I think should be revisited in some form later (chapter?).

PR #503, opened 2018-10-12
Forward euler in x86_64

Gathros:


contents/forward_euler_method/code/asm-x64/euler.s
```diff
@@ -0,0 +1,96 @@
...
+  movsd  xmm1, threshold
+  call   check_result                # Check if results are correct
+  mov    rdi, OFFSET fmt
+  mov    rsi, rax
+  xor    rax, rax
+  call   printf                      # Print out success code
+  add    rsp, 800                    # Deallocating array
+  pop    rbp
+  xor    rax, rax
+  ret
```
Liikt: Please insert a newline to avoid warnings

contents/forward_euler_method/code/asm-x64/euler.s
```diff
@@ -0,0 +1,96 @@
...
+  comisd xmm0, xmm1                  # Check if abs(...) > threshold
+  jbe    if_false
+  mov    rdi, OFFSET error_fmt       # If true print out array[i] and solution
+  mov    rax, 1
+  call   printf
+  mov    rax, 1                      # and set sucess code to failed (rax = 1)
+if_false:
+  add    r13, 1
+loop_check:
+  cmp    r13, r14                    # Check if index is less the array size
```
Liikt: I'm guessing you mean r12 here and not r14. r14 is 0x0 and thus the loop never get's executed once.

Liikt:
Well apart from the weird error I still get, this is okay.

Gustorn:
I get the newline warning as well. They should be there according to the `.editorconfig` anyway so might as well add it @Gathros

PR #504, opened 2018-10-12
Tree traversal in X86_64

Gathros:


Gathros:
I used a Circular array Queue in C so i just copied it here.

Gustorn:
@Liikt Linked lists are horrible data structures. While I can see _some_ of their educational appeal I don't think we should advocate them. Data structures using linear memory are pretty much always better, even in operations where linked lists should win out (except for very uncommon access patterns or a _huge_ amount of elements)

Gathros:
`bfs_queue` is there.

Issue #499, opened 2018-10-12
Scripts to clean up AAA files 


jiegillet:
It would be great to have a set of scripts we can use to clean up some files in the AAA regularly. 

- [ ] `.editorconfig` languages need to be sorted alphabetically
- [ ] languages in `book.json` need to be sorted alphabetically and indented properly

More ideas that we should probably discuss:

- [ ] script that runs linters with the `.editorconfig` rules for "all" languages in bulk

I'll edit ideas in as they get discussed

Butt4cak3:
> script that runs linters with the .editorconfig rules for "all" languages in bulk

That would probably be a pain in the butt. We'd have to create linter configurations from templates for all sorts of different linters that get installed in different ways on different operating systems.

jiegillet:
It would not be meant for users, just for us, we can run it once in a while. This way we wouldn't need to ask contributors to use spaces instead of tabs for example.

julianschacher:
We can do that, but even then we should try to merge as clean and as good formatted PRs as possible.

jiegillet:
Agreed. But some things fall through the cracks sometimes, I've merged code with tabs before because I hadn't noticed.

leios:
I agree with both points. We need to be more vigilant, but having a safety net is nice.

Butt4cak3:
> It would not be meant for users, just for us

I know. My comment still applies. Everyone of us would have to have a linter for every language installed.

jiegillet:
Or maybe we could have linters for the languages we use. I don't need to format JS code if you know how to do that, and I'll handle Haskell.

Butt4cak3:
Then that would require every one of us to regularly run the tool. If I always make sure that indentation is correct, it's almost useless for me to run it. If someone forgets to check indentation every time, they're likely to forget to run the tool too. It would greatly reduce the tool's usefulness.

julianschacher:
@jiegillet These scripts would then be executed by the user (or the reviewer) during the PR process, right?
That’s way better then the general clean-up updates, since it wouldn’t have the issue @Butt4cak3 mentioned and it wouldn’t spam the commit history.

julianschacher:
@Butt4cak3 it doesn’t hurt to have them tho. Maybe it catches something you as a reviewer miss. Humans aren’t perfect.

jiegillet:
[...]

PR #506, opened 2018-10-12
Cleanup files

jiegillet:
Two quick scripts, as discussed in #499 
Please test and review :)

### DIFF
--- a/tools/cleanup/clean_book_json.py
+++ b/tools/cleanup/clean_book_json.py
@@ -16,7 +16,7 @@ with open(path) as f:
 
 book = json.loads(text)
 book["pluginsConfig"]["api-language-selector"]["languages"]\
-                                   .sort(key=lambda x: x[u'name'])
+                         .sort(key=lambda x: x[u'name'].lower())
 
 with open(path + "_", "w") as f:
     f.write(json.dumps(book, indent=2))
--- a/tools/cleanup/clean_book_json.py
+++ b/tools/cleanup/clean_book_json.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env python
+"""
+Sorts languages in book.json alphabetically into book.json_.
+Check the file manually, then rename.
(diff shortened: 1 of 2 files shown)


PudottaPommin:
Looks good! Although I was at 1st confused it's order by name of language and not file extension. Which makes sense, but I would expect order by file extension.

jiegillet:
I don't think extension would make sense, because they are not really extensions, they are arbitrary. For example, we have
```
{
          "lang": "LabVIEW", 
          "name": "LabVIEW"
        },
```
I know that it would be better to stick to extensions in general, but my point holds. 
Language names, on the other hand, are meaningful. 
I'm also sorting `.editorconfig` by language name.

PudottaPommin:
I understand both points. 

One thing. Intention is to run this script from cding into tools/cleanup folder or from root folder of repo.

jiegillet:
Right now you have to be in tools/cleanup and run it from there. 
I'm not actually super happy with this, do you have a suggestion for a better choice? I can't use an absolute path.

PudottaPommin:
Maybe pass folder in which perform cleanup in parameter? Or remove `../../` from paths and considering running it from root folder.

jiegillet:
Parameter is better. No way to mess that up, thank you!

tools/cleanup/clean_book_json.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+"""
+Sorts languages in book.json alphabetically into book.json_.
+Check the file manually, then rename.
+You need to path the path to the file as an argument.
```
Butt4cak3: "pass"?

jiegillet: Gaaaaaah! 
Thank you -_-

jiegillet:
Muhahaha! Seems I could merge it myself :D

PR #507, opened 2018-10-13
Make Python Jarvis March conform to PEP8

berquist:


jiegillet:
It's hard to see the differences when GitHub considers a name change like a file deletion. Maybe in the future we should encourage people to change the name in a commit after the changes have been reviewed :(

Butt4cak3:
> change the name in a commit after the changes have been reviewed

Or just in a separate commit. It's easy enough to exclude that from a diff locally.

PR #497, opened 2018-10-12
Euclid algorithm in bash

Ariana1729:
umm quite useless but slightly better than writing in bf haha, also displayed nicely on gitbook too:>

PR #509, opened 2018-10-13
Added bubblesort in bash

Ariana1729:
More bash algos!

```
(( tmp = arr[(( j + 1 ))] ))
(( arr[(( j + 1 ))] = arr[j] ))
```

these look pretty crippled tho not sure how to make it clearer tbh

### DIFF
--- a/contents/bubble_sort/code/bash/bubble_sort.bash
+++ b/contents/bubble_sort/code/bash/bubble_sort.bash
@@ -9,13 +9,13 @@ bubble_sort() {
     (( len = ${#arr[@]} ))
 
     for ((i = 0; i <= len - 1; i++)); do
-	    for ((j = 0; j <= len - 2; j++)); do
-	        if (( arr[j] > arr[(( j + 1 ))] )); then
-		        (( tmp = arr[(( j + 1 ))] ))
-		        (( arr[(( j + 1 ))] = arr[j] ))
-		        (( arr[j] = tmp ))
-	        fi
-	    done
+        for ((j = 0; j <= len - 2; j++)); do
+            if (( arr[j] > arr[(( j + 1 ))] )); then
+                (( tmp = arr[(( j + 1 ))] ))
+                (( arr[(( j + 1 ))] = arr[j] ))
+                (( arr[j] = tmp ))
+            fi
+        done
     done
     echo ${arr[*]}
 }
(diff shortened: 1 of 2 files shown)


contents/bubble_sort/code/bash/bubble_sort.bash
```diff
@@ -0,0 +1,27 @@
...
+bubble_sort() {
+    local i
+    local j
+    local tmp
+    local len
+    local arr
+    arr=("$@")
+    (( len = ${#arr[@]} ))
+
+    for i in $(seq 0 $(( len - 1 ))); do
```
berquist: Don't shell out when you can use a built-in:
```bash
for ((i = 0; i <= len - 1; i++)); do
```

Ariana1729: ah ok

jiegillet:
I just want to say that I really appreciate the bash submissions and reviews, I'm learning a lot and it makes me want to learn proper bash instead the usual SO copy/paste that I hate ^^

Liikt:
I don't think this is how one would typically write bash.

```bash
#!/usr/bin/env bash
bubble_sort() {
    local i
    local j
    local tmp
    local len
    local arr
    arr=("$@")
    len=${#arr[@]}
    for i in $(seq 0 $(calc -e $len-1)); do
            for j in $(seq 0 $(calc -e $len-2)); do
                if [ ${arr[j]} -gt ${arr[j+1]} ]; then
                        tmp=${arr[j+1]}
                        arr[j+1]=${arr[j]}
                        arr[j]=$tmp
                fi
            done
    done
    echo ${arr[*]}
}
arr=(1 45 756 4569 56 3 8 5 -10 -4)
echo "Unsorted array: ${arr[*]}"
tmp=$(bubble_sort "${arr[@]}")
echo "Sorted array: ${tmp[*]}"```

This is how i would write bubblesort in bash

Ariana1729:
Hmm I was refering to [this pull request](https://github.com/algorithm-archivists/algorithm-archive/pull/497) for which one to choose

berquist:
@Liikt It's not how one would normally write bash, because we have the bad habit of shelling out (`seq`, `calc`) when bash internals should be used instead. Using `[...]` is also bad; that's the original Bourne shell, not the Bourne-again shell.

If this were pure `sh`, then yes, you would need to call out to other programs.

berquist:
@Ariana1729 I'll look over this again tonight my time.

Liikt:
I mean sure. I just have never seen even bash written like that. But then again I can't say i have seen that much bash at all. I personally still prefer my style but I can agree that it is more sh than bash. What I found is that the current code looks a bit like lisp :D

berquist:
:+1:

PR #512, opened 2018-10-13
Lisp bogo sort

Trashtalk217:
In the spirit of bogo sort I've patched this one together rather quickly, the shuffle is slow and the sortedp could maybe be simpler. But it overal works, ship it! :)

### DIFF
--- a/contents/bogo_sort/code/lisp/bogo-sort.lisp
+++ b/contents/bogo_sort/code/lisp/bogo-sort.lisp
@@ -1,0 +1,26 @@
+;;;; Bogo sort implementation
+
+(defun sortedp (list)
+  "Checks if a list is sorted"
+  (if (< (length list) 2)
+      t
+      (if (< (first list) (second list))
+          (sortedp (rest list))
+          nil)))
+
+(defun shuffle (list)
+  "Returns a shuffled list"
+  (loop for i from 0 upto (- (length list) 1)
+    do
+      (rotatef 
+        (nth i list)
+        (nth (random (length list)) list))
+    finally (return list)))
+
+(defun bogo-sort (list)
+  "Sorts a given list (eventually)"
+  (if (sortedp list)
+      list
+      (bogo-sort (shuffle list))))
+
+(print (bogo-sort (list 1 3 2 4)))
--- a/contents/bogo_sort/code/lisp/bogo-sort.lisp
+++ b/contents/bogo_sort/code/lisp/bogo-sort.lisp
@@ -9,12 +9,12 @@
           nil)))
 
 (defun shuffle (list)
-  "Returns a shuffled list"
-  (loop for i from 0 upto (- (length list) 1)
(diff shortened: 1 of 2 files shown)


contents/bogo_sort/code/lisp/bogo-sort.lisp
```diff
@@ -0,0 +1,26 @@
...
+          (sortedp (rest list))
+          nil)))
+
+(defun shuffle (list)
+  "Returns a shuffled list"
+  (loop for i from 0 upto (- (length list) 1)
+    do
+      (rotatef 
+        (nth i list)
+        (nth (random (length list)) list))
```
jiegillet: This is not correct (if you are implementing [Fisher-Yates algorithm](https://en.wikipedia.org/wiki/Fisher–Yates_shuffle)), you should pick a second index `j` in the range `i <= j < length list `.

contents/bogo_sort/code/lisp/bogo-sort.lisp
```diff
@@ -0,0 +1,26 @@
+;;;; Bogo sort implementation
+
+(defun sortedp (list)
+  "Checks if a list is sorted"
+  (if (< (length list) 2)
+      t
+      (if (< (first list) (second list))
```
jiegillet: This should be `<=` I think

jiegillet:
Also you didn't include it in the `.md` file

contents/bogo_sort/code/lisp/bogo-sort.lisp
Trashtalk217: I was actually implementing the Knuth shuffle (see [here](https://www.rosettacode.org/wiki/Knuth_shuffle#Common_Lisp)) and it's really hard not to copy the code straight so this was the best I could do.

Trashtalk217: If you know shuffling algorithms which are faster or easier to implement in FP, please let me know. The current one is not very efficient: `nth` and `rotatef` both have O(n). 
The knuth shuffle seems to be designed for arrays instead of lists. 
Bogo sort is, on the other hand, not exactly made to be very efficient.

jiegillet: The Knuth shuffle is another name for Fisher-Yates. You have to either loop for `i` from `0` to `length list - 2` and pick `j` from `i` to `length list - 1` (incl) OR loop for `i` from `length list - 1` to `0` and pick `j` from `0` to `i` (incl). Right now you're doing a mix of both.

Trashtalk217: I think I got the shuffle method down, (I even found a nice shortcut for in- and decrementing by one) and I'm not going to bother using arrays. Lists are clear and easy to use and this is bogo sort after all.

Trashtalk217:
I think I fixed it?

PR #514, opened 2018-10-14
Lisp Monte Carlo

Trashtalk217:
This one is pretty self explanatory actually.


Trashtalk217:
Prettified the output, should be good now

contents/monte_carlo_integration/code/lisp/monte-carlo.lisp
```diff
@@ -0,0 +1,18 @@
...
+  (loop repeat samples
+    with count = 0 
+    do
+      (when (in-circle-p (random 1.0) (random 1.0))
+            (incf count))
+    finally (return (* (/ count samples) 4.0))))
+
+(defvar pi-estimate (monte-carlo 5000000))
+(format t "Estimate: ~D ~%" pi-estimate)
+(format t "Error: ~D" (* (/ (- pi-estimate pi) pi) 100))
```
jiegillet: It's not a huge deal, but if you could take the absolute value of the difference, you would be consistent with the other implementations.
Also, is it me or did you add `~%` to the wrong line?

Trashtalk217: The tilde + % signals a new line, but I did forget the percentage sign on the error and taking the absolute value isn't a huge deal.

jiegillet: Let's test GH's new function: a suggestion that you can commit online
```suggestion
(format t "Error: ~D" (* (/ (- pi-estimate pi) pi) 100))
```

PR #515, opened 2018-10-15
PHP unify code style

PudottaPommin:
This PR unifies PHP implementations, which were pushed before defining code style.
Functionality is same, except one less output in TreeTraversal, which was JSON of the tree.

### DIFF
--- a/contents/tree_traversal/code/php/tree_traversal.php
+++ b/contents/tree_traversal/code/php/tree_traversal.php
@@ -1,121 +1,143 @@
 <?php
+declare(strict_types=1);
 
 class Tree implements JsonSerializable
 {
-  private $id;
-  private $children = [];
+    private $id;
+    private $children = [];
 
-  public function __construct(int $id, array $children = [])
-  {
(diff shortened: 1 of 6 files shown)


contents/tree_traversal/code/php/tree_traversal.php
```diff
@@ -1,121 +1,143 @@
...
-TreeTraversal::dfs_stack($node);
+TreeTraversal::DFSStack($node);
 
 echo 'DFS Queue:' . PHP_EOL;
-TreeTraversal::dfs_queue($node);
+TreeTraversal::DFSQueue($node);
 
 // If you want try binary order non-binary tree
 // Comment generation of new tree bellow
 // If you do that, exception will be thrown
```
Butt4cak3: I'd rephrase this and fix some of the spelling/grammar mistakes to make it more clear what you're trying to say here.

```
// If you want to try to run binary order on a non-binary tree,
// comment out the generation of the new tree below.
// If you do that, an exception will be thrown.
```

PR #519, opened 2018-10-19
Added monte carlo in bash

Ariana1729:
Code would need more echo ""|bc or dc if I used radius of 1 so used radius of 32767(limit of random), should be ok?

### DIFF
--- a/contents/monte_carlo_integration/code/bash/monte_carlo.bash
+++ b/contents/monte_carlo_integration/code/bash/monte_carlo.bash
@@ -13,10 +13,10 @@ monteCarlo() {
     local count
     local i
     ((count = 0))
-    for((i = 0; i < $1; i++));do
-    if(($(inCircle RANDOM RANDOM) == 1)); then
-        ((count++))
-    fi
+    for ((i = 0; i < $1; i++)); do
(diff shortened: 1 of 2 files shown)


contents/monte_carlo_integration/code/bash/monte_carlo.bash
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+inCircle() {
+    local ret
+    local mag
+    ((ret = 0))
+    if (($1 ** 2 + $2 ** 2 < 1073676289)); then # 1073676289 = 32767 ** 2
+	((ret = 1))
```
berquist: This should be indented.

contents/monte_carlo_integration/code/bash/monte_carlo.bash
```diff
@@ -0,0 +1,26 @@
...
+    fi
+    printf "%d" $ret
+}
+
+monteCarlo() {
+    local count
+    local i
+    ((count = 0))
+    for((i = 0; i < $1; i++));do
+	if(($(inCircle RANDOM RANDOM) == 1)); then
```
berquist: Everything in this block should be indented.

PR #520, opened 2018-10-19
Added cooley tukey to js

Ariana1729:
Should I use complex or complex.min, complex.js is easier to read but complex.min.js is smaller
Source: https://github.com/infusion/Complex.js


Ariana1729:
Only that file had ';'s in it so just removed those for consistency

Butt4cak3:
I don't think we should add libraries to the AAA like that. I'm not saying that we can't use them. I'm just saying that we shouldn't add the files.

I'd suggest adding a comment to the top of the files saying something like

```
// This code example makes use of complex.js.
// You can install it with `npm install complex.js`.
```

jiegillet:
This is not a proper review, but your indentation is all over the place.

leios:
Hey @Ariana1729, we screwed up. We should have looked at this PR before merging the newer FFT implementation in js (#599). As it stands now, we have an implementation in the algorithms archive, so I will close this PR.

We really appreciate the submission and I personally apologize for the mix up.

PR #522, opened 2018-10-20
Implemented Thomas Algorithm in Lua

Vexatos:
I chose not to mutate the input tables. Is that important?

### DIFF
--- a/contents/thomas_algorithm/code/lua/thomas.lua
+++ b/contents/thomas_algorithm/code/lua/thomas.lua
@@ -1,0 +1,35 @@
+function thomas(a, b, c, d)
+
+  -- Create tables and set initial elements
+  local c_prime = {c[1] / b[1]}
+  local result = {d[1] / b[1]}
+
+  for i = 2, #a do
+    -- Scale factor is for c_prime and result
+    local scale = 1.0 / (b[i] - a[i] * c_prime[i - 1])
+    c_prime[i] = c[i] * scale
+    result[i] = (d[i] - a[i] * result[i - 1]) * scale
+  end
+
+  -- Back-substitution
+  for i = #a-1, 1, -1 do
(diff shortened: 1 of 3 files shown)


jiegillet:
It looks good, and not mutating the input is great. We should probably ask someone (paging @Gorzoid ) who knows lua for an actual review though.
 One problem is how you print the system, it doesn't really work since a b and c are diagonal values of the matrix, not rows (and I realized you copied the Julia code, paging @leios ). If you want to prettify it, you can do something like the PHP implementation.

Vexatos:
I just based this one on the Julia implementation.

contents/thomas_algorithm/code/lua/thomas.lua
```diff
@@ -0,0 +1,36 @@
...
+local a = {0.0, 2.0, 3.0}
+local b = {1.0, 3.0, 6.0}
+local c = {4.0, 5.0, 0.0}
+local d = {7.0, 5.0, 3.0}
+
+print("The system")
+print(table.unpack(a))
+print(table.unpack(b))
+print(table.unpack(c))
+print(table.unpack(d))
```
Gorzoid: As jiegillet said, the system isn't actually this matrix. a,b and c are instead diagonals and d is the "result" column. To properly print the system we would need something like
```lua
print(b[1],	c[1], "",   "",   "|", d[1])
print("",   a[2], b[2], c[2], "|", d[2])
print("",   "",   a[3], b[3], "|", d[3])
```
I don't see much use in making code to properly any size augmented matrix so manually entering the indices should be fine.
It should be fine to use print(table.unpack(... for the solution however.

Vexatos: I agree, I based this on the Julia implementation. I will PR a fix to that as well I guess.

PR #526, opened 2018-10-20
Added exclamation marks to mutating julia functions

Vexatos:
By convention, Julia functions that mutate one of their parameters need to have an exclamation mark at the end of their name; this fixes all instances of that that I could find.

### DIFF
--- a/contents/bogo_sort/code/julia/bogo.jl
+++ b/contents/bogo_sort/code/julia/bogo.jl
@@ -9,15 +9,15 @@ function is_sorted(a::Vector{Float64})
     return true
 end
 
-function bogo_sort(a::Vector{Float64})
+function bogo_sort!(a::Vector{Float64})
     while(!is_sorted(a))
         shuffle!(a)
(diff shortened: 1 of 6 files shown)


contents/bogo_sort/code/julia/bogo.jl
```diff
@@ -9,15 +9,15 @@ function is_sorted(a::Vector{Float64})
...
 
-function bogo_sort(a::Vector{Float64})
+function bogo_sort!(a::Vector{Float64})
     while(!is_sorted(a))
         shuffle!(a)
     end
 end
 
 function main()
     a = [1, 3, 2, 4]
```
leios:
```suggestion
    a = [1.0, 3, 2, 4]
```

This is my fault, but there is a type error here where the bogo_sort function wants floats and we are reading in ints. Could we fix this?

Vexatos: Done.

PR #530, opened 2018-10-20
Implemented Bogo Sort in Crystal

Vexatos:
Roughly based on the Lua implementation.


Vexatos:
Good idea, probably.

contents/bogo_sort/code/crystal/bogo.cr
```diff
@@ -0,0 +1,22 @@
...
+    if a[i] > a[i + 1]
+      return false
+    end
+  end
+  true
+end
+
+def bogo_sort!(a)
+  while !is_sorted?(a)
+    a = a.shuffle
```
NIFR91: The function shuffle without `!` creates a new array, remember that in crystal  objects are passed by reference. 
```suggestion
    a.shuffle!
```

Vexatos: Not that it matters, but sure.

PR #534, opened 2018-10-24
Implemented Stable Marriage in Ruby

BjoernAkAManf:
Code should be self explanatory

### DIFF
--- a/contents/stable_marriage_problem/code/ruby/stable_marriage.rb
+++ b/contents/stable_marriage_problem/code/ruby/stable_marriage.rb
@@ -1,0 +1,113 @@
+class Generator
+    def initialize(size)
+        @size = size
+    end
+
+    def generate(prefix, r)
+        Array.new(@size){|i|
+            Person.new(
+                i,
+                "#{prefix} #{i}",
+                (0 ... @size).to_a.shuffle!(random: r)
+            )
+        }
+    end
+end
+
+class Person
+    def initialize(id, name, prefs)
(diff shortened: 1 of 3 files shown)


contents/stable_marriage_problem/code/ruby/stable_marriage.rb
```diff
@@ -0,0 +1,113 @@
...
+gen = Generator.new(4)
+men = gen.generate("Man", r)
+women = gen.generate("Woman", r)
+
+# Assume no Name is longer than 20 characters
+spacer = '-' * 20 * cols 
+
+round = 0
+# Solve the Problem
+loop do
```
nic-hartley:
```suggestion
1.step do |round|
```

(and of course remove the definition/modification of `round`)

contents/stable_marriage_problem/code/ruby/stable_marriage.rb
```diff
@@ -0,0 +1,113 @@
...
+        ]
+    end
+
+    protected
+    attr_reader :id, :name
+    attr_writer :partner
+
+    # Acts upon a given Proposal
+    def onPropose(partner)
+        if !self.lonely
```
nic-hartley:
```suggestion
        unless lonely
```

(plus or minus a question mark, if you take the previous suggestion about that)

BjoernAkAManf:
Will look into your suggestions next week.

Thanks!

PR #537, opened 2018-10-28
Create CODE_OF_CONDUCT.md

Neverik:


Butt4cak3:
People already know my opinion on whether, when and how we should add a code of conduct. I don't think we can just take a CoC and slap it on top of our project. There's a little more to that.

Other than that, I didn't know you're the entire project team.

> Instances of abusive, harassing, or otherwise unacceptable behavior may be
reported by contacting the project team at neverliberty@gmail.com.

PR #540, opened 2018-11-03
Add Bogo sort in Scala

jaoel:


jiegillet:
It looks great to me, the only thing I'd like to have is more output, like "The unsorted list is ... The sorted list is ...".
@kenpower would you like to try your hand at reviewing? Let's have a review by someone who actually knows Scala for a change :)

contents/bogo_sort/code/scala/bogo.scala
```diff
@@ -0,0 +1,24 @@
...
+  def bogoSort(list: List[Int]): List[Int] =
+    isSorted(list) match {
+      case false => bogoSort(shuffle(list))
+      case _ => list
+    }
+
+  def main(args: Array[String]): Unit = {
+    val unsorted = List(5, 2, 7, 1, -5)
+    val sorted = bogoSort(unsorted)
+    println(sorted)
```
kenpower: I agree with @jiegillet, for consistency with the other examples, we should output the sorted and unsorted list e.g from bubble sort.
`    val unsorted = List(9, 2, 0, 5, 3, 8, 1, 9, 4, 0, 7, 0, 9, 9, 0)

    println("Unsorted list is " + unsorted)
    println("  Sorted list is " + bubbleSort(unsorted))`

contents/bogo_sort/code/scala/bogo.scala
```diff
@@ -0,0 +1,24 @@
+import scala.util.Random.shuffle
+
+object BogoSort {
+
+  def isSorted(list: List[Int]): Boolean =
+    list match {
+      case a :: b :: _ if a < b => isSorted(list.tail)
+      case _ :: Nil => true
+      case _ => false
+    }
```
kenpower: This is probably just a mater of personal taste, but I like to put the terminal conditions of a recursive function first, like this:

    list match {
      case Nil => true
      case a :: b :: _ if a > b => false
      case _ :: tail => isSorted(tail)
    }

I'm not suggesting you change it, but for me it's easier to read the recursion this way.

jaoel: I also noticed that my implementation doesn't treat an empty list as sorted, causing the program to freeze, so I think it is safe to say that your version is better :)

kenpower: ok just commit the changes and we can merge

jaoel: It's done

PR #541, opened 2018-11-05
Adding Convolutions in X86_64

Gathros:


Gathros:
@Liikt What should I do?

Liikt:
hm i guess a couple of sentences that say that all zeros is the wanted output or something. or explaining what the zeros mean. because python i.e. has a complete different output than the asm and the C version of fft.

Gathros:
Should I print it in the terminal or as a comment?

Liikt:
no no to the terminal would be better

contents/convolutions/code/asm-x64/convolution.s
```diff
@@ -0,0 +1,328 @@
...
+  sub    rsp, 24
+  mov    r12, rdi
+  mov    r13, rsi
+  mov    r14, rdx
+  lea    rax, [rcx + r8]                     # Save the sum sizes into the stack
+  mov    QWORD PTR [rsp], rax
+  pxor   xmm0, xmm0
+  movsd  QWORD PTR [rsp + 8], xmm0           # Save 0 + 0i into the stack
+  movsd  QWORD PTR [rsp + 16], xmm0
+  mov    rbp, rcx                            # Find smallest array size
```
Liikt: You are using rbp here which is reserved for the base pointer and probably will do bad things with your stack. I guess you mean rbx.

contents/convolutions/code/asm-x64/convolution.s
```diff
@@ -0,0 +1,328 @@
...
+  cmp    rbp, r8
```
Liikt: You are using rbp here.

contents/convolutions/code/asm-x64/convolution.s
```diff
@@ -0,0 +1,328 @@
...
+  jge    conv_keep_size
+  mov    rbp, r8
```
Liikt: You are using rbp here.

contents/convolutions/code/asm-x64/convolution.s
```diff
@@ -0,0 +1,328 @@
...
+conv_keep_size:
+  xor    r15, r15
+conv_loop_i:
+  cmp    r15, QWORD PTR [rsp]                # Go through all entries in out array
+  je     conv_end_i
+  xor    rbx, rbx
+conv_loop_j:
+  cmp    rbx, r15                            # Check if the rbx is less then r15
+  je     conv_end_j
+  cmp    r15, rbp                            # Check if r15 is less then the smallest array size
```
Liikt: You are using rbp here.

contents/convolutions/code/asm-x64/convolution.s
```diff
@@ -0,0 +1,328 @@
...
+  je     conv_end_j
+  lea    rdi, [r12 + rbx]                    # Get signal1[j] * signal2[i-j]
```
Liikt: This is more a style thing but `rdi`/`rsi` aren't really general purpose registers but used for arguments. So you should change that to `rXx` or `rX`

Liikt:
I'm happy with this now. If @Gustorn wants to take a look at we can wait. Otherwise this is mergeable :+1:

PR #542, opened 2018-11-05
Adding FFT in X86_64

Gathros:


contents/cooley_tukey/code/asm-x64/fft.s
```diff
@@ -0,0 +1,398 @@
...
+  lea    rdi, [rsp + 1024]
+  mov    rsi, 1024
+  call   cooley_tukey
+  mov    rdi, rsp
+  lea    rsi, [rsp + 1024]
+  mov    rdx, 1024
+  call   approx
+  add    rsp, 2048
+  pop    r12
+  ret
```
Liikt: Before you return you should add a `xor rax, rax` because now the program returns the num of bytes of the last write as a return code since that is what is in `rax` at that time. But since the program executed correctly we want a good `0` as a return value.

Gathros: You're right, I forgot.

contents/cooley_tukey/code/asm-x64/fft.s
```diff
@@ -0,0 +1,400 @@
...
+  cvttsd2si rax, xmm0
+  mov    QWORD PTR [rsp], rax     # Save it to the stack
+  mov    r14, 1
+iter_ct_loop_i:
+  cmp    r14, rax                 # Check if r14 is greater then log2(N)
+  jg     iter_ct_end_i
+  movsd  xmm0, two                # Calculate stride = 2^(r14)
+  cvtsi2sdq xmm1, r14
+  call   pow
+  cvttsd2si rbp, xmm0
```
Liikt: Did you mean rbx here? Because one could run into big trouble messing with the base pointer.

contents/cooley_tukey/code/asm-x64/fft.s
```diff
@@ -0,0 +1,400 @@
...
+  xor    r15, r15
+iter_ct_loop_j:
+  cmp    r15, r13                 # Check if r15 is less then array size
+  je     iter_ct_end_j
+  movsd  xmm4, one                # Save 1 + 0i to stack
+  pxor   xmm5, xmm5
+  movsd  QWORD PTR [rsp + 24], xmm4
+  movsd  QWORD PTR [rsp + 32], xmm5
+  xor    rbx, rbx
+  mov    rax, rbp                 # Calculate stride / 2
```
Liikt: Here you use rbp again

contents/cooley_tukey/code/asm-x64/fft.s
```diff
@@ -0,0 +1,400 @@
...
+  sar    rax, 1
+iter_ct_loop_k:
+  cmp    rbx, rax                 # Check if rbx is less then stride / 2
+  je     iter_ct_end_k
+  mov    r8, r15                  # Saving pointers to X[k + j + stride / 2] and X[k + j]
+  add    r8, rbx
+  sal    r8, 4
+  mov    r9, rbp
```
Liikt: Here you use rbp again

contents/cooley_tukey/code/asm-x64/fft.s
```diff
@@ -0,0 +1,400 @@
...
+  movsd  QWORD PTR [r8 + 8], xmm1
+  movsd  xmm0, QWORD PTR [rsp + 24] # Calculating v * w
+  movsd  xmm1, QWORD PTR [rsp + 32]
+  movsd  xmm2, QWORD PTR [rsp + 8]
+  movsd  xmm3, QWORD PTR [rsp + 16]
+  call   __muldc3
+  movsd  QWORD PTR [rsp + 24], xmm0 # Saving answer
+  movsd  QWORD PTR [rsp + 32], xmm1
+  add    rbx, 1
+  mov    rax, rbp
```
Liikt: Here you use rbp again

contents/cooley_tukey/code/asm-x64/fft.s
```diff
@@ -0,0 +1,400 @@
...
+  sar    rax, 1
+  jmp    iter_ct_loop_k
+iter_ct_end_k:
+  add    r15, rbp
```
Liikt: Here you use rbp again

contents/cooley_tukey/code/asm-x64/fft.s
```diff
@@ -0,0 +1,400 @@
...
+  and    rax, rdi
+  cmp    rax, r14                  # Check if rax is greater then r14
+  jle    bit_reverse_no_swap       # If so then swap entries
+  shl    rax, 4                    # Times index by 16 to get bytes to entry
+  shl    r14, 4
+  movaps xmm0, XMMWORD PTR [r12 + rax]
+  movaps xmm1, XMMWORD PTR [r12 + r14]
+  movaps XMMWORD PTR [r12 + rax], xmm1
+  movaps XMMWORD PTR [r12 + r14], xmm0
+  shr    r14, 4
```
Liikt: You may be missing a `shr rax, 4` here. I am not sure about that but I assume

Liikt:
Also we should consider the problem described in #558. I know it's a bit harder and very verbose to do in asm, but maybe we can come up with a good solution

contents/cooley_tukey/code/asm-x64/fft.s
Gathros: No, I don't use that value of `rax` again. `rax` gets reset at line 189.

contents/cooley_tukey/code/asm-x64/fft.s
Gathros: No, I was using `rbp` like a `r12-15` register.

contents/cooley_tukey/code/asm-x64/fft.s
Liikt: alright that is fine then :+1:

contents/cooley_tukey/code/asm-x64/fft.s
Liikt: but you really shouldn't use `rbp` like a general purpose register. as soon as you hit a return everything is borked. also it is very much against convention

Gathros: Yeah, I didn't know that at the time. Good thing I pushed and popped it.

Liikt:
This looks good now. Maybe @Gustorn wants to take a look as well? If not we can merge this.

PR #543, opened 2018-11-08
added a nim implementation of the forward euler method

c252:
I added a nim implementation of the forward euler method.


contents/forward_euler_method/code/nim/forwardeuler.nim
```diff
@@ -0,0 +1,32 @@
+import math
+
+proc solve_euler(timestep: float, n:int): seq[float] =
```
berquist:
```suggestion
proc solveEuler(timestep: float, n: int): seq[float] =
```

berquist:
Thanks!

PR #552, opened 2018-12-03
Added implementation of Gaussian Elimination in Go

ravenbluedragon:


leios:
I think @Liikt and @depate might be able to review this.

PR #553, opened 2018-12-04
Added Monte Carlo Implementation in TI-83 BASIC

theinventor13:
Example code is also a visualization demo

### DIFF
--- a/contents/monte_carlo_integration/code/ti83basic/MONTE.TXT
+++ b/contents/monte_carlo_integration/code/ti83basic/MONTE.TXT
@@ -1,0 +1,29 @@
+"MONTE PROGRAMMED BY CODING WITH ETHANOL
+ClrHome
+Prompt M
+0->C
+0->I
+StoreGDB 0
+ClrDraw
+AxesOff
+CoordOff
+GridOff
+LabelOff
+0->Xmin:1->Xmax
+0->Ymin:1->Ymax
+Circle(0,0,1)
+For(I,0,M,1)
+::rand->X
+::rand->Y
+::Pt-On(X,Y)
+::If ((X*X+Y*Y)<1)
+::Then
+::::C+1->C
+::End
+End
+ClrHome
+(C/M)*4->E
+Text(28,43,E)
+Pause 
+RecallGDB 0
+ClrHome


nic-hartley:
"Coding with ethanol"

Ahh, that explains why you wrote Monte Carlo in _TI-BASIC_, of all languages.

theinventor13:
lol thats my moniker
i usually put it all as one word, e.g. codingwithethanol
but yes i also coded it because i was bored and somebody already did bubblesort in ti-basic

contents/monte_carlo_integration/code/ti83basic/MONTE.TXT
```diff
@@ -0,0 +1,29 @@
...
+LabelOff
+0->Xmin:1->Xmax
+0->Ymin:1->Ymax
+Circle(0,0,1)
+For(I,0,M,1)
+::rand->X
+::rand->Y
+::Pt-On(X,Y)
+::If ((X*X+Y*Y)<1)
+::Then
```
Butt4cak3: I see what you did here and I have to admit that it makes it slightly more readable on a PC screen, but I ran into the issue that `::Then` breaks on my TI-82. Also, the colons make the line above long enough to be broken up into two, which decreases readability more than it's worth in my opinion. What do you think about getting rid of the colons?

contents/monte_carlo_integration/code/ti83basic/MONTE.TXT
```diff
@@ -0,0 +1,29 @@
+"MONTE PROGRAMMED BY CODING WITH ETHANOL
+ClrHome
+Prompt M
+0->C
+0->I
+StoreGDB 0
```
Butt4cak3: The StoreGDB function changed between the TI-82 and TI-83. On an 82 this has to be `StoreGDB GDB1` instead. Do these GDB vars still exist on the 83? If not, then we'll keep it like this.

Butt4cak3:
To anyone who wants to pick this PR up:
I remember testing the code and it worked just fine. The changes I requested were merely for compatibility with the TI-82. If you want to leave the code sample as strictly TI-83 code, it's ready to merge.

contents/monte_carlo_integration/code/ti83basic/MONTE.TXT
thecnoNSMB: I'm not familiar with how the GDB system works but I can confirm that those variables do still exist on my TI-84 Plus.

thecnoNSMB: After checking [an unofficial changelog](https://brownmath.com/ti83/diff8384.htm#Diff8384), there is nothing to suggest that these variables wouldn't be on a TI-83 either.

leios:
This PR is almost ready to go. If anyone knows anything about TI83 and can change the few lines necessary (even in a separate PR), we can probably merge this.

leios:
Is this the code we want in the end?
```
"MONTE PROGRAMMED BY CODING WITH ETHANOL AND TOMMYTORTY10
ClrHome
Prompt M
0->C
0->I
StoreGDB 1

ClrDraw
AxesOff
CoordOff
GridOff
LabelOff
0->Xmin:1->Xmax
0->Ymin:1->Ymax
Circle(0,0,1)

For(I,0,M,1)
rand->X
rand->Y
Pt-On(X,Y)

If ((X*X+Y*Y)<1):Then
C+1->C
End

End

ClrHome
(C/M)*4->E
Text(28,43,E)
Pause 
RecallGDB 1
ClrHome

AxesOn
CoordOn
GridOn
LabelOn
```

Amaras:
@theinventor13 if you are ready to add your code to the relevant `.md` file, I am ready to merge this PR, since Butt4cak3 judged it was nice enough. :slightly_smiling_face: 
Otherwise, I think we can do it ourselves.

PR #556, opened 2018-12-23
adding go implementation to bogo_sort.md

leios:
We missed the go implementation in the bogo_sort chapter. I'm adding it in.


contents/bogo_sort/bogo_sort.md
```diff
@@ -129,6 +131,8 @@ We are done here!
 [import, lang:"crystal"](code/crystal/bogo.cr)
 {% sample lang="r" %}
 [import, lang:"r"](code/r/bogo_sort.r)
+{% sample lang="go" %}
+[import:3-, lang:"go"](code/go/bogo_sort.go)
```
julianschacher:
```suggestion
[import, lang:"go"](code/go/bogo_sort.go)
```
I think we should display the whole file in this part of the chapter.
We also have the "Submitted by [...]" included for other languages and the removal of those comments is a whole other issue we need to discuss.

Issue #96, opened 2018-04-29
Broken syntax highlighting for some C/C++ code examples


Butt4cak3:
**Note:** This issue exists primarily for documentation. I don't expect anyone else to step in. If you still want to help me investigate the issue, feel free to join me!

---

This is not a new one, but I found what causes the issue in some cases. I looked through the stable marriage code and found out that the second of these two lines breaks it. The first one is fine, but I left it in for comparison:

```cpp
static thread_local std::random_device global_random_device;
static thread_local std::mt19937 global_rng(global_random_device());
```

The issue is with highlight.js, which we have already [forked](https://github.com/algorithm-archivists/highlight.js) because I fixed another issue in it. The next step is to figure out what *exactly* causes it.

ubsan:
@Butt4cak3 is it maybe the `thread_local` keyword?

Butt4cak3:
@ubsan It's not the `thread_local` keyword. It's only the *second* of the two lines of code that break the highlighting. In other words...

This is fine:
```
static thread_local std::random_device global_random_device;
```

This isn't:
```
static thread_local std::mt19937 global_rng(global_random_device());
```

ubsan:
@Butt4cak3 ah, okay; that's direct-initialization. I can't believe that that actually breaks the highlighter... You should be able to fix it by switching to copy-initialization:

```cpp
static thread_local auto global_rng = std::mt19937(global_random_device());
```

Butt4cak3:
I'd rather try to patch the highlighter before I change the code. I've done it before. It's not *that* difficult. But I'll use that line if I can't get it to work. Thanks!

Gathros:
@Butt4cak3 wanted me to add that c has broken syntax too in tree traversal and huffman encoding.

Butt4cak3:
Link to tree traversal for reference https://algorithm-archivists.github.io/chapters/tree_traversal/tree_traversal.html

leios:
So it seems our problem here is related to nested parentheses: https://github.com/highlightjs/highlight.js/issues/1451, A solution is to use [prism](https://plugins.gitbook.com/plugin/prism). PR on the way and we'll close this issue after merging.

leios:
This has been solved after #561. Please let us know if there are further issues.

PR #561, opened 2018-12-25
Add the Prism Syntax Highlighting Plugin

julianschacher:
Add the prism syntax highlighting plugin to the AAA, set its config and modify some import statements.
The plugin on GitBook Plugins: https://plugins.gitbook.com/plugin/prism
Solves #96

### DIFF
--- a/book.json
+++ b/book.json
@@ -9,7 +9,9 @@
     "api-language-selector@https://github.com/algorithm-archivists/gitbook-plugin-api-language-selector.git",
     "include-codeblock",
     "ga",
-    "bulk-redirect"
+    "bulk-redirect",
+    "prism",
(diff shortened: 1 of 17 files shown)


leios:
Cool! I'll build this and check it out when I get home.

julianschacher:
Added WIP so that I can squash the commits before we rebase the commits onto master (to keep the signature).

leios:
Overall, it looks good. There still seem to be some errors on building certain esoteric languages, but I guess that is to be expected.

julianschacher:
Yeah, Prism drops errors for Chip-8, TI-83 Basic, Factor and Whitespace since they aren't on supported languages (yet).

PR #562, opened 2018-12-25
Make Python Jarvis March conform to PEP8, take 2

berquist:
A redo of https://github.com/algorithm-archivists/algorithm-archive/pull/507. In order to make the changes clearer, this shouldn't be a squash/rebase and merge, so the filename change and line changes are kept separate properly.

Check the second commit for the actual diff. Or, maybe the rename should be a separate PR?


leios:
It looks like the filename was changed, could you also update the markdown file to reflect this?

berquist:
Ok, check again.

leios:
Shoot, screwed up. I squashed and merged. Sorry about that. We can fix it, but I think we might just need to live with the changes. I'll be more careful from now on.

berquist:
I don't think it matters _too_ much; it was more about the review itself so one can see the lines that were actually changed.

PR #565, opened 2018-12-29
Fixed conv.jl + Shortened

Gorzoid:
Julia code was incorrect and failed to run for any input. I addressed those issues and also rewrote it in a more Julia style with list comprehensions and generator expressions.

### DIFF
--- a/contents/convolutions/code/julia/conv.jl
+++ b/contents/convolutions/code/julia/conv.jl
@@ -1,22 +1,14 @@
-function conv(signal1::Vector{Complex}, signal2::Vector{Complex})
-    n = length(signal1) + length(signal2)
-    out = Vector{Complex}(n)
-    sum = 0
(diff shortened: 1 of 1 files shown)


leios:
Yeah, the julia code in the algorithm archive is specifically written to be easy to read and extended to other languages. I'll be honest, this code was originally written as pseudocode and then translated to Julia before the [style guide](https://web.mit.edu/julia_v0.6.0/julia/share/doc/julia/html/en/manual/style-guide.html) existed, which is why it might be a bit wonky and the index starts as 0.

This PR made me realize that the entire chapter should be removed and re-worked heavily because none of the implementations match and the quality of the chapter is far below what I would consider appropriate for the Algorithm Archive. I already took down the video for similar reasons.

leios:
Thanks for bringing this to my attention, by the way!

leios:
Unfortunately going to have to close this one after #566, but I really appreciate the help!

PR #580, opened 2019-01-24
Verlet integration implemented in R

tsik1:

PR #585, opened 2019-01-30
Thomas algorithm R implementation

tsik1:


berquist:
See https://github.com/algorithm-archivists/algorithm-archive/pull/580#issuecomment-657100415.

PR #596, opened 2019-03-17
An implementation of Euclidean Algorithm in D

ADedecker:


leios:
Thanks for the implementation! I don't really know D too well, but would it be alright to review the code from the perspective of C?

c252:
The main algorithm seems to be just like the C code. Everything seems correct according to the D style guide https://dlang.org/dstyle.html

contents/euclidean_algorithm/code/d/euclidean_algorithm.d
```diff
@@ -0,0 +1,41 @@
...
+
+    return a;
+}
+
+void main()
+{
+    auto check1 = euclid_mod(64 * 67, 64 * 81);
+    auto check2 = euclid_sub(128 * 12, 128 * 77);
+
+    writeln(check1,'\n',check2);
```
c252: The only thing I would add (to keep it consistent with Leios's Julia code) would be to 
have the print statements say:
``` writeln("Modulus-based euclidean algorithm result: ", check1);```
```writeln("Subtraction-based euclidean algorithm result: ", check2); ```

PR #598, opened 2019-04-07
Verlet Integration in Nim

c252:
Implemented verlet integration in nim.


contents/verlet_integration/code/nim/verlet.nim
```diff
@@ -0,0 +1,56 @@
...
+let time_v: float = verlet(5.0, -10.0, 0.01)
+echo "Time for Verlet integration is: ", time_v
+
+let (time_sv, vel_sv) = stormer_verlet(5.0, -10.0, 0.01)
+echo "Time for Stormer Verlet integration is: ", time_sv
+echo "Velocity for Stormer Verlet integration is: ", vel_sv
+
+let (time_vv, vel_vv) = velocity_verlet(5.0, -10.0, 0.01)
+echo "Time for velocity Verlet integration is: ", time_vv
+echo "Velocity for velocity Verlet integration is: ", vel_vv
```
Liikt: Can you add a newline at the end of the file? Makes it better to read in a terminal.

c252: Fixed. Thank you.

contents/verlet_integration/code/nim/verlet.nim
```diff
@@ -0,0 +1,56 @@
+proc verlet(pos_in, acc, dt: float): float =
+  var
+    pos: float = pos_in
+    prev_pos: float = pos
```
berquist: Don't forget variable and procedure names should be camelCase, not snake_case.

contents/verlet_integration/code/nim/verlet.nim
```diff
@@ -0,0 +1,56 @@
...
+    pos: float = pos_in
+    time: float = 0.0
+    vel: float = 0.0
+
+  while pos > 0.0:
+    time += dt
+    pos += vel * dt + 0.5 * acc * dt * dt
+    vel += acc * dt
+
+  return (time, vel)
```
berquist: For here and the above two, implicit return is more idiomatic:
```suggestion
  (time, vel)
```

contents/verlet_integration/code/nim/verlet.nim
```diff
@@ -0,0 +1,56 @@
...
+
+let time_v: float = verlet(5.0, -10.0, 0.01)
```
berquist: You don't need the annotation here, inference works just fine:
```suggestion
let time_v = verlet(5.0, -10.0, 0.01)
```

contents/verlet_integration/code/nim/verlet.nim
c252: Oh neat, I did not know about implicit returning in Nim. Should I change the other implementations to use it too?

berquist: Sure, you can if you want.

It would be good if we could find a repo that has "good style practices" being used, rather than just looking through the style guide, which contains some confusing stuff IMO. Do you know of any? There was a ML or matrix math library, I think.

berquist:
Basically all the code I'm reading in [Arraymancer](https://github.com/mratsim/Arraymancer) and [nimble](https://github.com/nim-lang/nimble) uses implicit return. However I am perfectly fine with explicit returns, and if anyone would like to revisit this in the future, that's fine too.

PR #599, opened 2019-04-10
FFT Javascript Implementation

benchislett:
Made use of [Complex.js](https://github.com/infusion/Complex.js/) for dealing with complex numbers.

Implementation is fairly similar to the Python one, a key difference being the bit reversal step.
Instead of dealing with bits, the double-concat algorithm is implemented, described [here](https://codegolf.stackexchange.com/questions/83373/bit-reversal-permutations).


contents/cooley_tukey/code/javascript/fft.js
```diff
@@ -0,0 +1,92 @@
...
+const X = Array.from(Array(8), () => new Complex(Math.random(), 0));
+const Y = cooley_tukey(X);
+const Z = iterative_cooley_tukey(X);
+const T = dft(X);
+
+let ydiff = 0;
+let zdiff = 0;
+for (let i = 0; i < X.length; i++) {
+  ydiff += (Y[i].sub(T[i])).abs();
+  zdiff += (Z[i].sub(T[i])).abs();
```
Trashtalk217: In the Julia code the approximation-check has been made a seperate function, it might be a good idea to do that here as well.

contents/cooley_tukey/code/javascript/fft.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+}
+
+function bit_reverse(x) {
+  const N = x.length;
+  const indexes = bit_reverse_idxs(Math.log2(N));
+  return x.map((_, i) => x[indexes[i]]);
+}
+
+  // Assumes log_2(N) is an integer
```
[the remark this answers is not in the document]
benchislett: This is a [radix-2 cooley-tukey implementation](https://en.wikipedia.org/wiki/Cooley%E2%80%93Tukey_FFT_algorithm#The_radix-2_DIT_case) if I'm not mistaken, and requires that N be a power of two. The [bit reversal algorithm](https://en.wikipedia.org/wiki/Bit-reversal_permutation#Applications) also operates with radix-2 so it requires that N be a power of two, and thus that log2(N) be integral.
Here in particular, the displacement array is generated, and will always be of power-of-two length. The recursive nature of the `bit_reverse_idxs` function requires that the input be a positive integer or it will infinitely recurse as n goes into the negative non-integers.

contents/cooley_tukey/code/javascript/fft.js
benchislett: For consistency or syntactically? I personally think that it's fine where it is and that another function would be too much for this simple piece of code, but I'm not set in that opinion.

leios: I think it is a bit more clear in a function. At the very least, a comment would be appreciated.

benchislett: Done!

leios:
I'm happy with the implementation and the line numbers seem correct. If @Trashtalk217 is happy with the javascript as well, I believe this can be merged.

leios:
@Trashtalk217 If you have no further comments, I think we are ready to merge this one.

PR #603, opened 2019-05-31
Euclidean proof

Trashtalk217:
I thought it may be helpfull to provide a proof with the euclidean algorithm, since it is not directly obvious why the algorithm computes the greatest common divisor.

I don't know how the attribution, but I'll probably just give full credit to James Schloss, to avoid any confusion.

### DIFF
--- a/contents/euclidean_algorithm/euclidean_algorithm.md
+++ b/contents/euclidean_algorithm/euclidean_algorithm.md
@@ -166,6 +166,12 @@ Here's a video on the Euclidean algorithm:
 <iframe width="560" height="315" src="https://www.youtube.com/embed/h86RzlyHfUE" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 </div>
 
+## Proof
+
+Some intuition as to why the Euclidean Algorithm works lies in it's proof. Only a proof for the subtraction method will be given at this point, but the modular version follows the same line of reasoning.
+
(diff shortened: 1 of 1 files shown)


julianschacher:
Regarding the attribution:
If you add [this](https://gist.github.com/julianschacher/89c88731413d8e779ddda5f57a1c9e8c) as a comment to this PR (replacing AUTHOR with your name), you can add something like the following to the license section of `contents/euclidean_algorithm/euclidean_algorithm.md`:
```diff
##### Text
The text of this chapter was written by [James Schloss](https://github.com/leios) and is licensed under the [Creative Commons Attribution-ShareAlike 4.0 International License](https://creativecommons.org/licenses/by-sa/4.0/legalcode).
+ The proof was written by [YOUR NAME](LINK TO YOUR GITHUB/WEBSITE/WHATEVER) and is licensed under the [Creative Commons Attribution-ShareAlike 4.0 International License](https://creativecommons.org/licenses/by-sa/4.0/legalcode).
```

Trashtalk217:
I'd rather just give the rights to James since I don't want my real name in the book (yet) and I feel like writing "written by Trashtalk217" is a bit unprofessional.

julianschacher:
> I'd rather just give the rights to James since I don't want my real name in the book (yet) and I feel like writing "written by Trashtalk217" is a bit unprofessional.

Fair enough. We can always change it in the future after all :)

leios:
Thanks for the submission, we should certainly be adding proofs!

Should we provide a common format for proofs in the case that future proofs are added? (For example, should all proofs look like this: http://cheng.staff.shef.ac.uk/proofguide/proofguide.pdf).

I need to look at this proof a bit more rigorously and decide what we need for this section, in particular.

dovisutu:
Yup, there should be a proof format.
like this?
```
for all x>y which in N, exists z = (x,y)
thus                     x mod z = 0, y mod z = 0
thus                 (x-y) mod z = 0
thus                       (x,y) = (y,(x-y))
```
which `for all`,`exists` and `thus` is replaced by logic operator.

Trashtalk217:
Honestly I don't feel comfortable writing easy to read proofs yet. I still think proofs in the algorithm archive are a neat idea, but I don't think I'm qualified to write them. Maybe later.

dovisutu:
Yeah, it feels strange even when _I_ am reading it, and that's written long ago...
I don't think I'd rewrite this though, for I really can't get the time to do so.
So maybe someone can write a more viable version than mine. :D

Issue #510, opened 2018-10-13
Purity of numerical algorithms


berquist:
We had a discussion in https://github.com/algorithm-archivists/algorithm-archive/pull/495 about the purity of functions. The argument is that for many algorithms, specifically numerical ones, the implementation should not be pure for performance reasons, specifically memory. While the goal is not to write for performance, an algorithm should be written idiomatically for both 1. the language and (potentially) 2. how it is used. Also, from my experience, while it is easier to reason about the output of a pure numerical function by definition, it is not always easier to read. For algorithms of the scale in this book, this is probably a non-issue.

My proposal was to write all algorithms as being pure, but have text somewhere explaining the difference in the context of numerical algorithms, because it is an important topic not often touched upon.

jiegillet:
+1. I suggest adding something in the [how to contribute](https://www.algorithm-archive.org/contents/how_to_contribute/how_to_contribute.html) page. Maybe add a point after style named purity and explain the reason why we use pure functions here.

leios:
I think it fits best in the [Code style guide](https://www.youtube.com/watch?v=ijktorPAHGs) as a general principle.

I also agree that we should create a "disclaimer" chapter in place of the current how to contribute chapter (since the functionality of that one has been replaced by the how to contribute wiki). Basically, it would be nice to have some text describing the purpose of the code in the AAA.

Butt4cak3:
I'm the one who originally pointed out that the current Thomas algorithm implementations aren't pure and I suggested we'd change that. The more I thought about it, the less convinced I became, though. The reason is that it adds a certain degree of complexity to the algorithms and makes it a bit more difficult to understand it by reading the code. The code is not as self-explanatory anymore.

So one thing we could do to counteract this is to add some comments in the reference (Julia) code and *require* other implementations to also include them. I'm thinking of something like what I did in my Thomas algorithm cleanup: https://github.com/algorithm-archivists/algorithm-archive/blob/9ca87e1af2c01ef314361a0c4780d0409fdaa2c4/contents/thomas_algorithm/code/java/Thomas.java#L4

Yordrar:
I think that the goal of the code sample should be readability over performance, as there are a lot of possible ways to optimize code with many of them having some sort of obfuscation. The code should be as readable as possible so that anyone can understand it easily and adapt it to their needs. Now if that means that some functions should be pure and some of them shouldn't, so be it.

leios:
[...]

PR #604, opened 2019-06-01
Thomas algorithm implementation in Common Lisp

Trashtalk217:
While I probably could've done some funky recursion with this. I'm keeping it simple for this first submission. The `svref` and the `setf` syntax is wordy and annoying so I might make a specific function or even a macro for that. The function is destructive so I either have to find a copy command for a vector or re-write the entire function.

It'd be great to first have a review though.


contents/thomas_algorithm/code/clisp/thomas.lisp
```diff
@@ -0,0 +1,29 @@
...
+    for i from (- (length a) 2) downto 0 do
+      (decf (svref d i) (* (svref c i) (svref d (1+ i))))
+    finally (return d)))
+
+(defvar diagonal-a #(0 2 3))
+(defvar diagonal-b #(1 3 6))
+(defvar diagonal-c #(4 5 0))
+(defvar last-column #(7 5 3))
+
+(print (thomas diagonal-a diagonal-b diagonal-c last-column))
```
berquist: Because your inputs are arrays of integers, and none of your internal operations convert to a floating point representation, you're printing rational numbers. It would be good to also print the final result as floats for comparison with the other implementations.

contents/thomas_algorithm/code/clisp/thomas.lisp
```diff
@@ -0,0 +1,29 @@
+;;;; Thomas algorithm implementation in Common Lisp
+
+(defun thomas (a b c d)
+  "Returns the solutions to a tri-diagonal matrix destructively"
+  (setf (svref c 0) (/ (svref c 0) (svref b 0)))
+  (setf (svref d 0) (/ (svref d 0) (svref b 0)))
+
+  (loop
+    for i from 1 upto (1- (length a)) do
+      (setf
```
berquist: This block looks over-indented.

contents/thomas_algorithm/code/clisp/thomas.lisp
```diff
@@ -0,0 +1,29 @@
+;;;; Thomas algorithm implementation in Common Lisp
+
+(defun thomas (a b c d)
+  "Returns the solutions to a tri-diagonal matrix destructively"
```
berquist: Like you said, it'd be better for consistency to not destroy the inputs after the discussions in https://github.com/algorithm-archivists/algorithm-archive/pull/495 and https://github.com/algorithm-archivists/algorithm-archive/issues/510.

contents/thomas_algorithm/code/clisp/thomas.lisp
```diff
@@ -0,0 +1,29 @@
...
+
+  (loop
+    for i from 1 upto (1- (length a)) do
+      (setf
+        (svref c i)
+        (/ (svref c i) (- (svref b i) (* (svref a i) (svref c (1- i))))))
+      (setf
+        (svref d i)
+        (/
+          (- (svref d i) (* (svref a i) (svref d (1- i))))
```
berquist: If you want to write a macro, `(- (svref d i) (* (svref a i) (svref d (1- i))))` is probably the level of granularity that you want, with a separate macro for the `1+` version. Since you aren't shortening the code by using macros, make sure you use a descriptive macro name.

contents/thomas_algorithm/code/clisp/thomas.lisp
```diff
@@ -0,0 +1,29 @@
+;;;; Thomas algorithm implementation in Common Lisp
+
+(defun thomas (a b c d)
+  "Returns the solutions to a tri-diagonal matrix destructively"
+  (setf (svref c 0) (/ (svref c 0) (svref b 0)))
+  (setf (svref d 0) (/ (svref d 0) (svref b 0)))
```
berquist: Same comment about the macro here. Give them good docstrings.

berquist:
Poke for rebase. I think the only thing I care about from the review is printing in floating point.

contents/thomas_algorithm/code/clisp/thomas.lisp
Trashtalk217: true

contents/thomas_algorithm/code/clisp/thomas.lisp
Trashtalk217: fixed very dirtily

Trashtalk217:
I'm almost done, I found that (helper b a c i) is used twice so I might make that a separate variable in the loop. I also need an explanation for what the helper function actually does. I discovered that I didn't need macros and that functions could do the job.

Trashtalk217:
I'm almost done, I found that (helper b a c i) is used twice so I might make that a separate variable in the loop. I also need an explanation for what the helper function actually does. I discovered that I didn't need macros and that functions could do the job.

contents/thomas_algorithm/code/clisp/thomas.lisp
```diff
@@ -1,29 +1,34 @@
...
-(defvar diagonal-c #(4 5 0))
-(defvar last-column #(7 5 3))
+(defparameter diagonal-a #(0 2 3))
+(defparameter diagonal-b #(1 3 6))
+(defparameter diagonal-c #(4 5 0))
+(defparameter last-column #(7 5 3))
 
-(print (thomas diagonal-a diagonal-b diagonal-c last-column))
+;; should print 0.8666667 1.5333333 -0.26666668
+(format t "~{~f ~}" (coerce (thomas diagonal-a diagonal-b diagonal-c last-column) 'list))
```
berquist, line 34: This is actually printing
```
5.133333 0.46666667 0.26666668
```
for me with ECL, SBCL, CCL, and CMUCL (CLISP isn't working on my machine right now).

Trashtalk217: I was doing the wrong test, the actual output is 0.866666 1.533333 -0.266666. The div-set also wasn't working, it had to be a macro.

Trashtalk217: Still need additional comments though

PR #605, opened 2019-06-02
Graham scan implementation in Common Lisp

Trashtalk217:
I'm on a roll.

Generally happy with this one except the `graham-scan` function itself. It's not that clean and quite slow; `nth` and `append` are somewhat expensive. Otherwise I'm mostly good and I appreciate any review. 

I messed up with git so that's why it looks so messy.


berquist:
Can you do a rebase on top of current master to fix the conflict?

contents/graham_scan/code/clisp/graham-scan.lisp
```diff
@@ -0,0 +1,60 @@
+;;;; Graham scan implementation in Common Lisp
+
+(defstruct (point (:constructor make-point (x y))) x y)
+
+(defun counterclockwise-p (p1 p2 p3)
```
berquist: If you read the chapter, the explanation describes a routine that returns a number which, if <= 0, indicates that the given points don't form a CCW turn. This version isn't wrong, it's just that it makes the implementation different, so it should be changed for consistency.

Trashtalk217: Done and solved @berquist !

Trashtalk217:
Basically as good as it's going to get. `graham-scan` Is still destructive because the standard sort is destructive in Common Lisp. But because `sort` only shuffles things around and the order of these things doesn't matter, I'm sure it won't matter that much. You could always pass a copy of the gift to the function (`(copy gift)`).

PR #612, opened 2019-06-13
using isapprox(a,b) function in julia instead of rolling our own

leios:
The julia implementation of fft had an approximate function we created ourselves, but there was no reason to do so when the `isapprox(a,b)` function exists

### DIFF
--- a/contents/cooley_tukey/code/julia/fft.jl
+++ b/contents/cooley_tukey/code/julia/fft.jl
@@ -110,23 +110,17 @@ function butterfly(x)
     return y
 end
 
-function approx(x, y)
-    val = true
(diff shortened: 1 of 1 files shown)


c252:
I personally like the Unicode syntax in Julia ``` a ≈ b ```, but I am probably in the minority.

leios:
I'll be honest, I just don't want to have to change text editors to put in the LaTeX. The unicode-ness of Julia also makes it less readable in my opinion... But I know people who absolutely love unicode

c252:
I understand, also with ```isapprox``` you can set the tolerance value, which you cannot do with the unicode.

PR #610, opened 2019-06-12
Bubble Sort Scratch Implementation

dovisutu:
In the building, I met a strange problem:
It can't pass. It said 
```
Template render error: (C:\Users\w\Downloads\algorithm-archive\contents\bubble_sort\bubble_sort.md) [Line 1605, Column 35]
parseSignature: expected comma after expression
```
what? Line 1605?
Also, I tried to paste the code of tree traversal here(with ../tree_traversal), but still fail.

PR #614, opened 2019-06-15
implemented euclidean algorithm in Scratch

c252:
I mostly did this to see what the processes for implementing scratch 3 code in the AAA would be like. It seemed like the easiest way was to write the scratch code in text/block-plugin syntax and use this tool https://s3blocks.github.io/# (like what was done in #610 ). I picked 200 as the image width because I thought it looked best.


jiegillet:
Looks good. However, I would also suggest to always have a Scratch project linked with the code in it. For more complicated algorithms, it's nice to be able to see it in action and test it, because there are no easy way to import the code from the AAA on Scratch.

c252:
Thanks @jiegillet , I added the link to the scratch project.

PR #616, opened 2019-06-21
implemented graham scan in C plus plus

AkashDhiman:
I also took the edge case when the algorithm can encounter multiple points with the same polar angle or same minimum y co-ordinate, and sorting is done via cross product itself so I didn't include trigonometric functions to do sorting. this is my first PR, let me know if there is something I can improve upon.


contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,102 @@
+#include<iostream>
+#include<vector>
+
+struct point {
+  double x;
+  double y;
+};
+
+point origin;
```
dovisutu: Don't use a global `origin` variable. Localize it in `grahamScan` and as argument of `ccwCheck`.
```suggestion
```

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+  int crossProduct = (b.y - a.y) * (c.x - b.x) - (b.x - a.x) * (c.y - b.y);
+  if (crossProduct == 0)
+  {
+    return 0;
+  }
+  return ( crossProduct > 0)? -1 : 1;
+}
+
+int compare(const void *vp1, const void *vp2){
```
dovisutu: The `compare` function is totally useless, I think. It's OK to have points colinear, as the one inwards will be replaced.

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,102 @@
...
+}
+
+int ccwCheck(point a, point b, point c){
+
+  int crossProduct = (b.y - a.y) * (c.x - b.x) - (b.x - a.x) * (c.y - b.y);
+  if (crossProduct == 0)
+  {
+    return 0;
+  }
+  return ( crossProduct > 0)? -1 : 1;
```
dovisutu: It's unnececcary to return 1,0 or -1. It's not a bool, so we have to compare it again, therefore, why not just return it?
Also, the previous `if` is useless, too.
```suggestion
  return (b.y - a.y) * (c.x - b.x) - (b.x - a.x) * (c.y - b.y);
```

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if ((y < yMin) || (yMin == y && points[i].x < points[min].x)){
+      yMin = points[i].y;
+      min = i;
+    }
+  }
+  swap(points[0], points[min]);
+  origin = points[0];
+
+  //sorting by polar angle and removing duplicates
+  qsort(&points[1], points.size() - 1, sizeof(point), compare);
```
dovisutu: I prefer to use `sort<_Ranit, _Pr>(const _RanIt _First, const _RanIt _Last, _Pr _Pred)` in <algorithm\> instead of `qsort(void*  _Base, size_t _NumOfElements, size_t _SizeOfElements, int (*_CompareFunction)(const void*, const void*)` in <corecrt_search.h\>.
Also, since `compare` is useless, the _Pr can be a simple lambda `[&](point& a, point& b){return ccwCheck(points[0],a,b)`

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,102 @@
...
+  std::vector<point> pointsSorted;
```
dovisutu: You don't needed to create new one. original one is OK to edit

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+void print(std::vector<point> points)
+{
+  std::cout << "the points of hull are as follows:\n";
+  for (size_t i = 0; i < points.size(); i++) {
+    std::cout << "(" << points[i].x << "," << points[i].y << ")\n";
+  }
+}
+
+std::vector<point> grahamScan(std::vector<point> points){
```
dovisutu: I prefer to make this function more _c++y_ by taking the begin and end iterator as argument instead of copying the input.
 ```suggestion
 template<typename Iter>
std::vector<point> grahamScan(Iter first, Iter last){
```

File: contents/graham_scan/code/c++/graham_scan.cpp [hunk 7]
```diff
@@ -0,0 +1,102 @@
...
+  for(size_t i = 1; i < points.size(); i++){
+    while(i < points.size() - 1 && ccwCheck(origin, points[i],
+      points[i+1]) == 0){
+      i++;
+    }
+    pointsSorted.push_back(points[i]);
+  }
+
+  //creating the convex hull
+  std::vector<point> hull;
```
dovisutu: As mentioned, the algorithm shouldn't create a new vector to store hull.
>To save memory and expensive append() operations, we ultimately look for points that should be on the hull and swap them with the first elements in the array. If there are Melements on the hull, then the first M elements in our output random distribution of points will be the hull.

AkashDhiman:
Thank you for considering to look at my code for graham scan in c++, I will
try to improve upon the things you mentioned and will modify the code
accordingly as soon as possible.

leios:
Thanks for the review here @dovisutu!

Are you on discord? I apologize if you are there already and I missed you. A lot of discussion happens there.

dovisutu:
> Thanks for the review here @dovisutu!
> Are you on discord? I apologize if you are there already and I missed you. A lot of discussion happens there.

I'd like to, but I can't login discord here in China...
Maybe try my VPN again...
(It worked, but 3 months ago it broke.)

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
...
+                i++;
+            }
+    }
+    m++;
+    std::swap<point>(points[i],points[m]);
+  }
+  std::vector<point>::const_iterator first = points.begin();
+  std::vector<point>::const_iterator last = points.begin()+m;
+  std::vector<point> hull(first, last);
+  return hull;
```
dovisutu: I prefer to inline these statements:
```suggestion
  return std::vector<point>(points.begin(), points.begin() + m);
```

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
...
+                m--;
+                continue;
+            } else if (i == points.size()) {
+                break;
+            } else {
+                i++;
+            }
+    }
+    m++;
+    std::swap<point>(points[i],points[m]);
```
dovisutu: You don't have to specify type. Automatic Type Inference will infers that `_Ty` is `Point`.
Also, spacing please.
```suggestion
    std::swap(points[i], points[m]);
```

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
...
+  }
+  std::vector<point>::const_iterator first = points.begin();
+  std::vector<point>::const_iterator last = points.begin()+m;
+  std::vector<point> hull(first, last);
+  return hull;
+}
+
+void print(std::vector<point> points){
+  std::cout << "points in hull are as follows: \n";
+  for (size_t i = 0; i < points.size(); i++) {
```
dovisutu: I prefer to use `for` *with range* similar with `for each` in other languages. Like this:
```suggestion
  for (auto p : points) {
```
Then use `p` instead of `points[i]` below.

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
+#include<iostream>
```
dovisutu: Please space between `#include` and header in order to be consistent with other implementations in AAA.

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+std::vector<point> grahamScan(std::vector<point> points){
+  //selecting lowest point as pivot
+  int lowIndex = 0;
+  for (size_t i = 1; i < points.size(); i++) {
+    if(points[i].y < points[lowIndex].y){
+      lowIndex = i;
+    }
+  }
+  std::swap<point>(points[0], points[lowIndex]);
```
dovisutu: You don't have to specify type. Automatic Type Inference will infers that `_Ty` is `Point`.
```suggestion
  std::swap(points[0], points[lowIndex]);
```

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
...
+  point pivot = points[0];
+
+  //sorting points by polar angle
+  sort(points.begin()+1, points.end(), [pivot](point a, point b){
```
dovisutu: `sort` is in namespace `std`.
```suggestion
  std::sort(points.begin()+1, points.end(), [pivot](point a, point b){
```

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
...
+  for (size_t i = 0; i < points.size(); i++) {
+    std::cout <<"(" << points[i].x << ", "<< points[i].y <<")\n";
+  }
+}
+
+int main(){
+  std::vector<point> points = {{-5, 2}, {5, 7}, {-6, -12}, {-14, -14}, {9, 9},
+                              {-1, -1}, {-10, 11}, {-6, 15}, {-6, -8}, {15, -9},
+                              {7, -7}, {-2, -9}, {6, -5}, {0, 14}, {2, 8}};
+  std::vector<point> hull = grahamScan(points);
```
dovisutu: I'd like a printing statment *before* finding hull, in order to show the points.

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,66 @@
+#include<iostream>
+#include<vector>
+#include<algorithm>
+
+struct point{
+  float x;
+  float y;
```
dovisutu: Why use float here? I think double is OK.

leios:
@dovisutu No pressure on the discord, then! I just wanted to make sure you were kept in-the-loop with the latest changes, but most of that discussion is at least duplicated here on github.

Thanks for all the help!

dovisutu:
> @dovisutu No pressure on the discord, then! I just wanted to make sure you were kept in-the-loop with the latest changes, but most of that discussion is at least duplicated here on github.
> Thanks for all the help!

BTW, I just fixed my VPN...
I'll try to join discord chat then.

AkashDhiman:
I am getting a somewhat odd runtime error, the code compiles and the output is also shown but (only sometimes) with the error:
> *** Error in `./a.out': free(): invalid next size (fast): 0x00005577d36da230 ***

after which it prints a memory map which I have not shown here.
I suspect it to be a buffer overflow maybe? Can anyone give me a suggestion on how to deal with it?

AkashDhiman:
I think I fixed the runtime error, I believe that in the previous commit the runtime error was due to line 32 for loop, where the iterator could access location beyond the last element of vector.

contents/graham_scan/code/c++/graham_scan.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <vector>
+#include <algorithm>
+
+struct point{
+  double x;
+  double y;
+};
+
+bool ccw(point a, point b, point c){
+  return ((b.x - a.x)*(c.y - a.y) >= (b.y - a.y)*(c.x - a.x));
```
dovisutu: This is not a __*`strict weak ordering`*__ Predication.
```suggestion
  return ((b.x - a.x)*(c.y - a.y) > (b.y - a.y)*(c.x - a.x));
```

PR #618, opened 2019-07-03
Update images of Scratch Implemantation to Scratch3

dovisutu:
**Current** Implementations in Scratch:
- [x] Thomas Algorithm
- [x] Tree Traversal
  - combine `dfs.svg` and `dfs-from.svg` into `dfs-pre.svg`
- [ ] Verlet Intergration (not gonna cover)
  - It's imcomplete, and needs to update separatly.
- [ ] FFT (not gonna cover)
  - It only contains a link to the scratch screenshot, and needs to update separatly.

**In-progress** (aka. in-PR) implementations are already in Scratch*3* and well-formatted.(2 are merged.)
About Verlet, *maybe* @jiegillet can update it.


c252:
We should get everyone's opinion on this, but I personally think having the images be a width of 200 looks the best.

dovisutu:
Well, I think `width` shouldn't be a constant, as while some pictures getting too big, some pictures are still too small.
I think it's better to keep each charactar the same size, but it'll be extremly hard.

c252:
Oh that's a good point, I agree.

leios:
For Verlet integration, we were still figuring out how to put scratch into the AAA. As such, it is a bit different than the other implementations.

I guess we should write a new verlet integration method for scratch.

Thanks for all the work here!

leios:
Looking at the PR now, would it be possible to add the full tree traversal code to the bottom of the chapter, similar to other languages?

I need to go through everything with a fine comb, but I think we can deal with Verlet in another PR, if you like.

dovisutu:
Yeah, we can put the full tree traversal code in the end (except the initializing of nodes, maybe I should add a `define`, but a little inconsistent with original version). Just put the whole `tree_traversal.md`.
It's OK to separate verlet away since it may need longer time to update.
BTW: I missed FFT... Which was worse, only a single lonely link in the end.

dovisutu:
BTW, I find that keeping each character the same size isn't hard... Just need to look in the properties of the `svg` tag in the .svg file, and the `witdh` property is the `width` property we need when import into the AAA.

leios:
Cool, keep up the good work! No rush here!

leios:
As another note here, for tree traversal, we are just linking to one of @jiegillet's scratch projects. Maybe we could put some full code in there, like the other languages?

Outside of that, I am happy with this PR and am willing to merge it, if it is no longer a work-in-progress.

dovisutu:
The scratch project for tree traversal is inface a *maze* (generator??) used to show a tree... Probably not fit with other implementations since they only test the algorithm.

leios:
Yeah, that's my point. It doesn't fit any other implementation.

jiegillet:
[My project](https://scratch.mit.edu/projects/174017753/editor/) does contain tree traversal code, even though it's maybe not used directly in the project. Check the "Tree traversal" sprite.

dovisutu:
Yeah, I see it now.

PR #619, opened 2019-07-03
Monte Carlo Scratch Implementation

dovisutu:


contents/monte_carlo_integration/code/scratch/monte_carlo.txt
```diff
@@ -0,0 +1,13 @@
+<(((x) * (x)) + ((y) * (y))) < (1) >
+
+
+define Monte Carlo with point numer (num)
+set [Count v] to (0)
+repeat (num)
+    set [x v] to ((pick random (0) to (10000000)) / (10000000))
```
leios: I don't know if there is a defined `RAND_MAX` in scratch like there is for C (or other programming languages), but this value of `10000000` falls out of the sky. We could read it in as another variable, maybe?

It doesn't matter that much and I am fine with the current version, but it would be nice to have this set as another variable somewhere or something. I am not sure what is the cleanest implementation.

dovisutu: Nope, there isn't `RAND_MAX` defined.
The real thing is, The randomizer of scratch is just a **`Integer`** generetor, in order to generate proper double, I have to do that.
But it's OK to handle `10000000` with another variable, although it may be confusing since there are only *variable*s in scratch, no *const*s.

leios: In that case, can we leave a comment, like so:

```
<(((x) * (x)) + ((y) * (y))) < (1) >


define Monte Carlo with point number (num)
set [Count v] to (0)
repeat (num)
    set [x v] to ((pick random (0) to (10000000)) / (10000000)) // 10000000 is arbitrarily chosen
    set [y v] to ((pick random (0) to (10000000)) / (10000000))
    if <(((x) * (x)) + ((y) * (y))) < (1) > then
        change [Count] by (1)
    end
end
set [EstPI] to ((4) * ((Count) / (num)))
```

dovisutu: OK。

leios:
@c252 You mentioned you wanted to tackle this one; however, the only inconsistency I see is that there is no `radius` parameter. And this is inconsistent for monte carlo in general (#625).

What are your thoughts?

c252:
If it is alright with you then I am fine with it.

dovisutu:
Maybe I can add `radius` in, along with a note `// the radius is 1`.

leios:
Sure, that would work for now at least... And we have the scratchblocks code, so it's no where near as hard to modify as before.

dovisutu:
BTW, after I looked up the source code of scratch3, I realized the random generator *can* generate `double`s as long as at least one of `FROM` and `TO` is `double`.

leios:
Oh, nice! So we don't need the arbitrarily large value anymore!

contents/monte_carlo_integration/code/scratch/monte_carlo.txt
```diff
@@ -0,0 +1,13 @@
+<(((x) * (x)) + ((y) * (y))) < (1) >
+
+
+define Monte Carlo with point number (num)
+set [Count v] to (0)
+repeat (num)
+    set [x v] to ((pick random (0) to (10000000)) / (10000000)) // 10000000 is arbitarily chosen
```
leios: Now that we know we can change this to floating-point precision, could we change this to read:
```
   set [x v] to (pick random (0.0) to (1.0))
```

(this is just a reminder for me when looking back at this pr later)

leios:
There seems to be a lot of empty space under the `Algorithm.svg` image. Could this be cropped?

c252:
I think somewhere in the program you might need to set the radius variable to 1 because in scratch3 the variables will default to 0.

dovisutu:
About `radius`, I globally ran a block (set radius to 1) and never change it afterwards.
I don't add it in algorithm, because I add a comment `// the radius is 1`. If add initialization, the comment can be removed.

c252:
I figured that was how you did it, but my worry was that if someone else copied this scratch code and ran it the radius would be 0 and it wouldn't work properly. (It's not a very big deal)

dovisutu:
About `Algorithm.svg`, It's because I go back and use [the tool here](http://scratchblocks.github.io/) instead of [this one](http://s3blocks.github.io) because of better quality at comments, but it seemly like has a wrong size.
The blank can be cropped simply by changing the `width` and `height` approximately.

contents/monte_carlo_integration/code/scratch/monte_carlo.txt
```diff
@@ -0,0 +1,13 @@
...
+define Monte Carlo with point number (num)
+set [Count v] to (0)
+repeat (num)
+    set [x v] to (pick random (0.0) to (1.0))
+    set [y v] to (pick random (0.0) to (1.0))
+    if <(((x) * (x)) + ((y) * (y))) < ((radius) * (radius)) > then // the radius is 1
+        change [Count] by (1)
+    end
+end
+set [EstPI] to ((4) * ((Count) / (num)))
```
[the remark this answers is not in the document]
dovisutu: The problem is my main loop contains some out-of-topic thing (visualization) with a broadcast (`draw`), maybe I should remove it.
Except that, it should work fine.

dovisutu:
What I thought is if they see the comment, they should know they shouldset `radius` to 1.
Although it's possible that somebody who know nothing about scratch just mechanically copied it...?

contents/monte_carlo_integration/code/scratch/monte_carlo.txt
```diff
@@ -10,4 +10,13 @@ repeat (num)
         change [Count] by (1)
     end
 end
-set [EstPI] to ((4) * ((Count) / (num)))
\ No newline at end of file
+set [EstPI v] to ((4) * ((Count) / (num)))
+
+when @greenflag clicked
+set [PI v] to (3.1415926535897932384626433832795028841971493993)
+ask (Point size?) and wait
```
c252: This looks really good, however I think this should be "Number of points" not "Point size".

dovisutu: OK.

c252:
Sorry for late response I have been crazy busy. This looks good to !

leios:
The `Algorithm.svg` does not seem to be viewable on my end anymore.

dovisutu:
Can't view *again*? Umm... I'll fix it if I have time...
*(possibly not because schoolwork)*

leios:
This PR is ready to go, except that the svg image has whitespace under it when rendering the Algorithm Archive. I can try to take a look at it in more detail if you cannot figure this out. There are some issues with css + svg.

We can relatively easily convert everything to png and it will work fine; however, there might be a way to get this to work with a [viewbox](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox)

This was tested in Firefox and chromium on linux.

dovisutu:
Yeah, that is confirmed. Setting `transform: scale(1)` fixes it, as what I did in #682.
Now this should be fixed - tested on (old) Edge and Chrome.

PR #622, opened 2019-07-17
Plotting chapter addition

leios:
This is a rough draft for a plotting chapter in the Algorithm Archive.

I have implemented everything in gnuplot; however, I should have allowed for other people to add in whatever plotter they want later (eg python, julia). In these cases, all the images should probably be regenerated in the plotter of choice.

As a note, this is the first step in an overhaul to the algorithm archive. Next up is domain coloring, followed by FFT's, then convolutions, and then everything else. There's a lot to do and I'm busy right now with my thesis, but I'm doing what I can.


dovisutu:
It seems like `data/scatterplot_data.dat` is missing, as you try to import it at L539 of `plotting.md`.

leios:
Thanks for catching that! I'll fix it soon.

leios:
I am relatively happy with how this chapter reads; however, I have considered adding an addition section on format specifiers as described [here](http://gnuplot.sourceforge.net/docs_4.2/node184.html). I decided not to put it in this release because it isn't really used in any of the AAA plots so far (I think).

Let me know if you think I should add it in!

leios:
Hey guys, I need this chapter for the next one... So I will probably merge it in tomorrow.

c252:
Sounds good, I just read over the chapter and everything looks fine to me.

PR #623, opened 2019-07-17
Tree Traversal in Common Lisp

Trashtalk217:
I'm working from a new machine so all the git stuff might be messed up. But this is the best version of tree traversal I can make. Reviews are welcome!


Trashtalk217:
I agree with you, although the current fix I've implemented is a bit jank. The recursion makes it very difficult to differentiate between different trees and subtrees, so I have to treat every element the same. They are all on newlines or none of them are. Hence the extra `(format t "~%")` in the place where the code is tested.

There is a possible third solution. I could define a function inside the tree-traversal functions where the recursion happens:
```
(defun recursion (tree)
    (labels ((actual-recursion (tree) (actual-recursion (changed-tree))))
    (format t "~%"))
```
That would work, but I'm worried it might be slightly over-engineered. The `bfs-queue` and `dfs-stack` are easier fixes as they don't relies on recursion, but if we only changed those two we would lose uniformity.

I'm currently leaning to the over-engineered as I only need to over-engineer 3 methods which is doable.

Trashtalk217:
Here's the over-engineered solution. I hope this is an improvement.

berquist:
I was thinking of the un-overengineered solution, where you need the extra newline prints in between each call. I have no strong preference either way. I do like replacing the `let` in `bfs-queue` entirely with the loop. Tell me which version you want, then I'll merge.

Trashtalk217:
Simple is probably best, this is the version I'm happiest with.

berquist:
Nah, not that cruel.

PR #627, opened 2019-10-01
Unified Lua code style across pages

Vexatos:
* Proper use of `local`
* Proper use of snake_case
* Removed a bunch of dead spaces
* Fixed monte carlo code to be equivalent to other implementations
* Added comment from the julia implementation to monte carlo code.

### DIFF
--- a/contents/huffman_encoding/code/lua/huffman.lua
+++ b/contents/huffman_encoding/code/lua/huffman.lua
@@ -1,10 +1,10 @@
 local function frequency_array(str)
   -- Collect all frequency values into a dict
   local map = {}
-  for c in str:gmatch"." do -- Iterate over each character in str
+  for c in str:gmatch(".") do -- Iterate over each character in str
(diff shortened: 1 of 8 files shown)


contents/huffman_encoding/code/lua/huffman.lua
```diff
@@ -1,10 +1,10 @@
 local function frequency_array(str)
   -- Collect all frequency values into a dict
   local map = {}
-  for c in str:gmatch"." do -- Iterate over each character in str
+  for c in str:gmatch(".") do -- Iterate over each character in str
```
DavidLoftus: What's the reasoning against omitting parentheses, is it just for consistency reasons?

Vexatos: Yea, mostly consistency. It can be confusing for people who are not quite as used to Lua's oddities since it is a rather rare language feature.

PR #626, opened 2019-10-01
Add Euclidean Algorithm in PowerShell

PaddyKe:

PR #630, opened 2019-10-01
Added Forward Euler Method in PowerShell

PaddyKe:
This PR is currently work in progress since #626 is not merged yet. It contains the languagedefinition for PowerShell. As soon as it is merged, I‘ll remove it.


leios:
#626 was merged, so I think this can be removed

PR #631, opened 2019-10-02
bubblesort in v

Liikt:
I found a cool new language in the making called V. so yeah this is bubblesort in V.

### DIFF
--- a/contents/bubble_sort/bubble_sort.md
+++ b/contents/bubble_sort/bubble_sort.md
@@ -68,6 +68,8 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 [import:2-14, lang:"emojicode"](code/emojicode/bubble_sort.emojic)
 {% sample lang="bash" %}
 [import:2-21, lang:"bash"](code/bash/bubble_sort.bash)
+{% sample lang="v" %}
+[import:2-21, lang:"v"](code/v/bubble_sort.v)
 {% sample lang="scratch" %}
 <p>
   <img  class="center" src="code/scratch/bubble_sort.svg" width="400" />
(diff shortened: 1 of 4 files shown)


contents/bubble_sort/code/v/bubble_sort.v
```diff
@@ -0,0 +1,20 @@
...
+			}
+		}
+	}
+}
+
+fn main() {
+	arr := [1, 45, 756, 4569, 56, 3, 8, 5, -10, -4]
+	println('Array unsorted:')
+	println(arr)
+	bubblesort(mut arr)
```
berquist: Seems like a compiler bug that you can mutate this here when it isn't `mut arr` above.

contents/bubble_sort/bubble_sort.md
```diff
@@ -68,6 +68,8 @@ This means that we need to go through the vector $$\mathcal{O}(n^2)$$ times with
 [import:2-14, lang:"emojicode"](code/emojicode/bubble_sort.emojic)
 {% sample lang="bash" %}
 [import:2-21, lang:"bash"](code/bash/bubble_sort.bash)
+{% sample lang="v" %}
+[import:2-21, lang:"v"](code/v/bubble_sort.v)
```
berquist: Your bounds aren't correct.

contents/bubble_sort/code/v/bubble_sort.v
Liikt: no i think it is intentional that this works. but i don't think it's idiomatic. i will change that. 

this one i wrote waaaay back when i first learned about vlang and when doing `function(mut var)` would actually create a faulty program

contents/bubble_sort/bubble_sort.md
Liikt: they probably aren't correct in any vlang pr :sweat: will fix

contents/bubble_sort/code/v/bubble_sort.v
berquist: Considering that most of the bubble sort implementations mutate their input, if you declare `mut arr := ...`, I'll accept this.

PR #632, opened 2019-10-02
jarVis march

Liikt:
jarvis march in V


berquist:
Can you revert your changes to `book.json` and `.editorconfig`?

PR #633, opened 2019-10-02
V forward euler

Liikt:
Add forward euler in V

### DIFF
--- a/contents/forward_euler_method/code/v/euler.v
+++ b/contents/forward_euler_method/code/v/euler.v
@@ -1,0 +1,38 @@
+import math
+
+fn forward_euler(timestep f64, n int) []f64 {
+    mut res := [f64(0.0)].repeat(n)
+    res[0] = f64(1)
+    for x := 1; x < n; x++ {
+		res[x] = res[x-1] - 3.0*res[x-1]*timestep
+	}
+	return res
+}
+
+fn check(result []f64, threshold, timestep f64) bool {
+	mut approx := true
+	for x := 0.0; int(x) < result.len; x++ {
+		solution := math.pow(math.E, -3.0 * x * timestep)
+		if math.abs(result[int(x)]-solution) > threshold {
+            tmp := result[int(x)]
+			println(tmp)
+			println(solution)
+			approx = false
+		}
+	}
+	return approx
+}
+
+fn main() {
+	timestep := .01
+    threshold := .01
+	n := 100
+
+	result := forward_euler(timestep, n)
+
+	if check(result, threshold, timestep) {
+		println("All values within threshold")
+	} else {
+		println("Value(s) not within threshold")
+	}
+}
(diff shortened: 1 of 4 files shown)


contents/forward_euler_method/code/v/euler.v
```diff
@@ -0,0 +1,38 @@
...
+    res[0] = f64(1)
+    for x := 1; x < n; x++ {
+		res[x] = res[x-1] - 3.0*res[x-1]*timestep
+	}
+	return res
+}
+
+fn check(result []f64, threshold, timestep f64) bool {
+	mut approx := true
+	for x := 0.0; int(x) < result.len; x++ {
```
berquist:
```suggestion
	for x := 0; x < result.len; x++ {
```
then you avoid the two other integer conversions and are left with the one float conversion in the `solution` calculation.

contents/forward_euler_method/code/v/euler.v
```diff
@@ -0,0 +1,38 @@
...
+		solution := math.pow(math.E, -3.0 * x * timestep)
```
berquist: All consts are lowercase:
```suggestion
		solution := math.pow(math.e, -3.0 * x * timestep)
```
though really you can be using `exp`:
```suggestion
		solution := math.exp(-3.0 * x * timestep)
```

contents/forward_euler_method/code/v/euler.v
```diff
@@ -0,0 +1,38 @@
...
+		if math.abs(result[int(x)]-solution) > threshold {
+            tmp := result[int(x)]
+			println(tmp)
+			println(solution)
```
berquist: I'm ok with this, but maybe use string interpolation along with these values to say "there is a mismatch at this index", etc.

contents/forward_euler_method/code/v/euler.v
```diff
@@ -0,0 +1,38 @@
...
+			approx = false
+		}
+	}
```
berquist: It looks like here and in many other places, you've mixed tabs in with spaces.

contents/forward_euler_method/code/v/euler.v
```diff
@@ -0,0 +1,38 @@
+import math
+
+fn forward_euler(timestep f64, n int) []f64 {
+    mut res := [f64(0.0)].repeat(n)
```
berquist: 💯 for the array method

contents/forward_euler_method/code/v/euler.v
```diff
@@ -0,0 +1,38 @@
...
+    res[0] = f64(1)
+    for x := 1; x < n; x++ {
+		res[x] = res[x-1] - 3.0*res[x-1]*timestep
```
berquist: My personal stylistic preference is to be like Python and have spaces around every operator:
```suggestion
		res[x] = res[x - 1] - 3.0 * res[x - 1] * timestep
```
which, now that I look at it, you can factor out `res[x - 1]`...

berquist:
Actually, something I just noticed when reading https://github.com/vlang/v/blob/master/vlib/builtin/array.v is that they use tabs in the stdlib, not spaces.

PR #634, opened 2019-10-02
V thomas

Liikt:
thomas algorithm in vlang

### DIFF
--- a/contents/thomas_algorithm/code/v/thomas.v
+++ b/contents/thomas_algorithm/code/v/thomas.v
@@ -1,32 +1,34 @@
-fn thomas(a []f32, b []f32, c mut []f32, d mut []f32) []f32 {
-	c[0] = c[0] / b[0]
-	d[0] = d[0] / b[0]
+fn thomas(a []f32, b []f32, c []f32, d []f32) []f32 {
+	mut new_c := c
+	mut new_d := d
+	new_c[0] = new_c[0] / b[0]
+	new_d[0] = new_d[0] / b[0]
 
 	for i := 1; i < d.len; i++ {
-		scale := 1. / (b[i] - c[i-1]*a[i])
-		c[i] *= scale
-		d[i] = (d[i] - a[i]*d[i-1]) * scale
+		scale := 1. / (b[i] - new_c[i-1]*a[i])
+		new_c[i] *= scale
+		new_d[i] = (new_d[i] - a[i]*new_d[i-1]) * scale
 	}
 
 	for i := d.len - 2; i >= 0; i-- {
-		d[i] -= c[i] * d[i+1]
+		new_d[i] -= new_c[i] * new_d[i+1]
 	}
 
-	return d
+	return new_d
 }
 
 fn main() {
 	a := [0.0, 2.0, 3.0]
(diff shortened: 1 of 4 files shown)


contents/thomas_algorithm/code/v/thomas.v
```diff
@@ -0,0 +1,33 @@
...
+		scale := 1. / (b[i] - c[i-1]*a[i])
+		c[i] *= scale
+		d[i] = (d[i] - a[i]*d[i-1]) * scale
+	}
+
+	for i := d.len - 2; i >= 0; i-- {
+		d[i] -= c[i] * d[i+1]
+	}
+
+	return d
```
berquist: Returning something that you've also mutated as an input argument isn't good no matter the language. You should return `d`, remove `mut` for `c` and `d`, and make copies of those two input arrays.

contents/thomas_algorithm/code/v/thomas.v
```diff
@@ -0,0 +1,33 @@
...
+	b := [1.0, 3.0, 6.0]
+	mut c := [4.0, 5.0, 0.0]
+	mut d := [7.0, 5.0, 3.0]
+
+	println("The system,")
+	println("[1.0  4.0  0.0][x] = [7.0]")
+	println("[2.0  3.0  5.0][y] = [5.0]")
+	println("[0.0  3.0  6.0][z] = [3.0]")
+	println("has the solution:")
+	solve := thomas(a, b, mut c, mut d)
```
berquist:
```suggestion
	solution := thomas(a, b, mut c, mut d)
```
is more descriptive to me.

contents/thomas_algorithm/code/v/thomas.v
Liikt: You are of course right. I changed it.

Thanks :+1:

berquist:
Can you revert your changes to `book.json` and `.editorconfig`?

contents/thomas_algorithm/code/v/thomas.v
```diff
@@ -0,0 +1,35 @@
+fn thomas(a []f32, b []f32, c []f32, d []f32) []f32 {
+	mut new_c := c
+	mut new_d := d
```
berquist: This isn't actually making a copy.

Liikt: this defeats the whole purpose of `mut`, as the current implementation of V mutates the original array as well.

berquist: Yikes. Filed https://github.com/vlang/v/issues/2548.

In the meantime, can you do it "by hand"? Write a function that, given an array, makes a new one, iterates over the old one pushing each item into the new one, then returns the new one?

berquist: Meaning, while I acknowledge there is a compiler bug, we should implement the workaround anyway (at least for now) in the interest of safety.

Liikt: I asked Alex (the main dev) already about it. he knows and is currently reworking the memory stuff. after that is done this bug should be gone as well

PR #641, opened 2019-10-11
Apply clang-format to Graham Scan in C++

berquist:
Unsure if I should also take this opportunity to C++-ify things (things should be passed by const reference).


leios:
Feel free to update this to be more C++y, if you like! Also: don't forget to update the text for the new line numbers

berquist:
Ok, this is ready for re-review.

PR #643, opened 2019-10-11
Update Graham Scan in JavaScript to use points from Julia implementation

berquist:


berquist:
I don't think that's out of scope for the same PR since it's related cleanup. Atomic commits are good but requiring separate PRs is overkill.

Liikt:
yeah I can agree to that.

PR #644, opened 2019-10-11
Make Graham Scan in Python PEP8 compliant

berquist:


Trashtalk217:
Woops forgot to comment. I checked these changes against the [official styleguide](https://www.python.org/dev/peps/pep-0008/) and it looks great. So I approve this PR.

PR #650, opened 2019-10-19
CGparallel.cpp

pratik07101999:


julianschacher:
Hi!
First of all: Thanks for making the effort of contributing to the Algorithm Archive.
Unfortunately it seems like this isn't an algorithm that exists in the Algorithm Archive yet and we don't take code that isn't associated with any existing chapter describing an algorithm.

PR #651, opened 2019-10-19
implemented bubblesort in coffeescript and added it to book.json

c252:


contents/bubble_sort/code/coffeescript/bubblesort.coffee
```diff
@@ -0,0 +1,13 @@
+bubble_sort = (a) ->
```
berquist: It looks like camelCase is idiomatic:
```suggestion
bubbleSort = (a) ->
```

c252:
Sorry that took so long. I added it to the .editorconfig

PR #654, opened 2019-11-14
Implementation of Euclidean Algorithim in VimL

mukundan-v-senthil:


leios:
As a quick review, this looks good; however, it does not perfectly conform to the vimscript style guide: https://google.github.io/styleguide/vimscriptguide.xml, "when in doubt, treat vimpscript like python style."

This guide said 2 space indenting, but 4 is probably also fine.

Admittedly, I am not sure if this is the right style guide or not, so if you were following a different guide, let me know!

Also: If you prefer not to use a guide for vimscript, that's fine for me. I don't mind merging it as is. Just let me know!

PR #658, opened 2019-12-18
Fix incorrect explanation in comments of monte_carlo.py

thecnoNSMB:


Trashtalk217:
Looks good, can be merged

thecnoNSMB:
On further inspection, this error seems to also be implied by the actual text of the chapter:
> we are only integrating the upper right quadrant of the circle from  0<x,y<1 . This [...] requires us to multiply our output by 4.
Hmmm.

jiegillet:
I agree. Integrating only a quarter of the box because the circle is symmetric and centered is using too much prior knowledge. While in real applications identifying and exploiting symmetries is very important, it's not the best for teaching purposes. It might be good to discuss this in the text too actually.

About setting the radius in relation to the box size, I now agree that the code should allow for different sizes, but it should come with a paragraph about precision: you get the best estimation with the smallest possible box around the shape. This could come with examples of answers for different radii.

contents/monte_carlo_integration/code/python/monte_carlo.py
```diff
@@ -19,8 +19,8 @@ def monte_carlo(n_samples, radius = 1):
 		if(in_circle(x, y, radius)):
 			in_circle_count += 1
 
-	# Since we've generated points in upper left quadrant ([0,radius], [0, radius])
-	# We need to multiply the number of points by 4	
+	# Since the ratio this calculates is pi/4 (pi*r^2 / 4*r^2)
```
jiegillet: Actually the ratio this calculates is ` pi/4 = (pi*r^2/4) / (r^2)` because we are integrating a quarter of a circle in a square of side `r`. Then we need to multiply by 4 to get the full circle in units of `r^2`.

If we were integrating over the full box of side `2r` we would be calculating ` pi/4 = (pi*r^2) / (4 r^2)` in units of `4r^2` that we would need to multiply by 4 to get the answer in units of `r^2`. 

This is quite subtle and potentially confusing IMO, we need to be careful.

leios:
Right, this should be a quick rework for the text, but I am not sure if we should also make huge changes to the provided code?

I feel like a better MC example might be integrating various shapes instead of finding pi.

jiegillet:
Finding pi is the `Hello World` of MC though :)

contents/monte_carlo_integration/code/python/monte_carlo.py
thecnoNSMB: Thanks for catching that, I'll fix the code comment.

thecnoNSMB:
I was collecting how other resources handle this on the community Discord server, and was asked to copy it here:
- [Wikipedia](https://en.wikipedia.org/wiki/Monte_Carlo_integration#Example) integrates the entire circle
- [Scratchapixel](https://www.scratchapixel.com/lessons/mathematics-physics-for-computer-graphics/monte-carlo-methods-in-practice/monte-carlo-rendering-practical-example) uses the McBeth chart as its example (fitting, as it's a computer graphics website)
- [Wolfram MathWorld](http://mathworld.wolfram.com/MonteCarloIntegration.html) uses no example
- [towards data science](https://towardsdatascience.com/monte-carlo-integration-is-magic-c7f817d89610) integrates the entire circle
- [Rosetta Code](https://rosettacode.org/wiki/Monte_Carlo_methods) either integrates the entire circle or a quarter of the circle depending on the language
- [Academo](https://academo.org/demos/estimating-pi-monte-carlo/) integrates the entire circle
- [GeeksforGeeks](https://www.geeksforgeeks.org/estimating-value-pi-using-monte-carlo/) appears to integrate a quarter of the circle, but their math is a bit weird
- [Think Twice](https://www.youtube.com/watch?v=ELetCV_wX_c) integrates the entire circle

ntindle:
[lang: python]

PR #640, opened 2019-10-10
Implemented Huffman encoding in Kotlin

neverix:

PR #660, opened 2019-12-23
Kotlin huffman

Kroppeb:


Gathros:
This code doesn't work for a string of one unique character multiple times. I don't know Kotlin at all but I have a fix you can look at and improve on.

```Kotlin
if (priorityQueue.size == 1) {
    val left = priorityQueue.remove()
    val right = Leaf(0, '\u0000')
    root = Branch(left, right)
} else {
    root = priorityQueue.remove() as? Branch ?: error("No support for string of one unique letter")
}
```
I'm not sure how to get `right` to be a null object but this will do.

berquist:
I just saw this. Why are these changes on top of https://github.com/algorithm-archivists/algorithm-archive/pull/640?

ntindle:
[lang: kotlin]

PR #662, opened 2019-12-23
Fixing Huffman Encoding edge case failure for assembly x64

Gathros:


leios:
Is anyone actually able to review this? If not, I am happy to run it locally and try to figure it out, but I would need a little help figuring out how to do that.

ntindle:
[lang: asm]

stormofice (approved):
The code looks good to me and it produces the correct output. Thanks!


[If anyone wants to test this themselves, change the `text` variable to something like `aaaaaaaaaa` and compile it with `gcc -no-pie huffman.s`]

stormofice:
I  think that [code] line length should be checked by a CI task / GitHub action or something similar in the future, so I think we can merge this one (and fix the line length later on with other ones that will fail when this gets enforced).

ntindle:
Imo we should aim for the editor to support wordwrap rather than have the maintainers be responsible for checking line length

stormofice:
I created an issue (see #962) to separate the discussion about long lines and unblock this PR.

Amaras:
Since this has been approved and a specific issue has been opened, I'm merging this in

PR #664, opened 2019-12-29
Split Operator method in go

Gathros:


contents/split-operator_method/code/go/split_op.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return energy_final * par.dx
+}
+
+func main() {
+	var par params
+	var opr operators
+
+	initParams(&par, 5.0, 0.05, 256, 100, true)
+	initOperators(&opr, par, 0.0, -1.0)
```
Liikt: `initOperators` could be a method of a `params` struct, but I would definitely make it return an `operator` struct instead of passing a pointer.

contents/quantum_systems/code/go/energy.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	fft(energy_k, true)
+
+	for i := 0; i < len(wfc); i++ {
+		energy_k[i] *= wfc_c[i]
+		energy_r[i] = wfc_c[i] * h_r[i] * wfc[i]
+	}
+
+	energy_final := 0.0
```
DavidLoftus:
```suggestion
	energyFinal := 0.0
```
Prefer camelCase over snake_case in Go.

contents/split-operator_method/code/go/split_op.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	for i := 0; i < len(x); i++ {
+		x[i] = y.At(i) / cmplx.Sqrt(complex(float64(len(x)), 0))
+	}
+
+}
+
+func initParams(xmax, dt float64, res, timesteps int, im bool) params {
+	var par params
```
DavidLoftus: Use inline initialization instead, i.e. ```
par := params{
    xmax: xmax,
    ...
}
```

File: contents/split-operator_method/code/go/split_op.go [hunk 4]
```diff
@@ -0,0 +1,197 @@
...
+		plan.Destroy()
+	}
+
+	for i := 0; i < len(x); i++ {
+		x[i] = y.At(i) / cmplx.Sqrt(complex(float64(len(x)), 0))
+	}
+
+}
+
+func initParams(xmax, dt float64, res, timesteps int, im bool) params {
```
DavidLoftus: `newParams` might be a better fit now that it returns a value

contents/split-operator_method/code/go/split_op.go
```diff
@@ -0,0 +1,197 @@
...
+	size int
+	v    []complex128
+	pe   []complex128
+	ke   []complex128
+	wfc  []complex128
+}
+
+func fft(x []complex128, inv bool) {
+	y := fftw.NewArray(len(x))
+	for i := 0; i < len(x); i++ {
```
DavidLoftus: Same thing as cooley-tukey, use range loops.

contents/split-operator_method/code/go/split_op.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+
+			sum *= par.dx
+
+			for j := 0; j < opr.size; j++ {
+				opr.wfc[j] /= complex(math.Sqrt(sum), 0)
+			}
+		}
+
+		filename := "output" + strconv.Itoa(i) + ".dat"
```
DavidLoftus: Could we move this to another function? It's already considerably big.

contents/split-operator_method/code/go/split_op.go
```diff
@@ -0,0 +1,197 @@
...
+		file, err := os.Create(filename)
+
+		if err != nil {
+			fmt.Println("Cant write: ", err)
+			continue
```
DavidLoftus: Not very nice error handling. Move this code to a function that returns an error and let main() handle it by calling log.Fatal(err) if err != nil

contents/split-operator_method/code/go/split_op.go
```diff
@@ -0,0 +1,197 @@
...
+
+		fft(opr.wfc, true)
+
+		for j := 0; j < opr.size; j++ {
+			opr.wfc[j] *= opr.pe[j]
+		}
+
+		for j := 0; j < opr.size; j++ {
+			density[j] = math.Pow(cmplx.Abs(opr.wfc[j]), 2)
+		}
```
DavidLoftus: Can these be collapsed into one? Not sure if that would make it look better.

contents/quantum_systems/code/go/energy.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+func fft(x []complex128, inv bool) {
+	y := fftw.NewArray(len(x))
+	for i := 0; i < len(x); i++ {
+		y.Set(i, x[i])
+	}
+
+	if inv {
+		plan := fftw.NewPlan(y, y, fftw.Backward, fftw.Estimate)
```
DavidLoftus: Having a `var plan *fftw.Plan` variable could avoid code duplication for Execute() and Destroy()

contents/split-operator_method/code/go/split_op.go
Gathros: I was following the C and Julia versions and this is how it was done there.

contents/quantum_systems/code/go/energy.go
Gathros: I don't know how to do that. Any suggestions?

contents/split-operator_method/code/go/split_op.go
leios: I don't mind doing the naive error handling here (unless it's really, really against Go standards). I don't think Go readers will have any problem seeing this as error handling without the function.

contents/split-operator_method/code/go/split_op.go
leios: What do you want to move to a function? The definition of the output file?

contents/split-operator_method/code/go/split_op.go
leios: Yeah, I think they can without any issues. I think we should probably change other similar languages (like C / C++) to do a single loop here as well? Not really sure what's more readable.

contents/split-operator_method/code/go/split_op.go
DavidLoftus: The saving the output data to a file, seeing as the function is already super big.

contents/split-operator_method/code/go/split_op.go
DavidLoftus: Possibly just use `log.Fatal("Can't write: ", err)` rather than continuing, as that's how the Julia code behaves.

contents/split-operator_method/code/go/split_op.go
DavidLoftus: Yeah in some ways its cleaner to keep loops separated, especially since density does depend on wfc, it indicates that they occur as disjoint operations.

contents/quantum_systems/code/go/energy.go
DavidLoftus: Essentially this: ```go
var plan *fftw.Plan
if inv {
	plan = fftw.NewPlan(y, y, fftw.Backward, fftw.Estimate)
} else {
	plan = fftw.NewPlan(y, y, fftw.Forward, fftw.Estimate)
}
plan.Execute()
plan.Destroy()```
or: ```go
var dir fftw.Direction
if inv {
	dir = fftw.Backward
} else {
	dir = fftw.Forward
}
plan := fftw.NewPlan(y, y, dir, fftw.Estimate)
plan.Execute()
plan.Destroy()
```

Go doesn't have ternaries so this is the only option

Comment (DavidLoftus) [reply contents/quantum_systems/code/go/energy.go]: Neither makes it look especially better the what you already have

File: contents/split-operator_method/code/go/split_op.go (hunk 7 above)
Comment (Gathros) [reply contents/split-operator_method/code/go/split_op.go]: So should I replace lines 143 and 144 with that code?

File: contents/split-operator_method/code/go/split_op.go (hunk 8 above)
Comment (Gathros) [reply contents/split-operator_method/code/go/split_op.go]: I would keep it on the grounds that it is more readable.

File: contents/split-operator_method/code/go/split_op.go [hunk 10]
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	for i := range(x) {
+		x[i] = y.At(i) / cmplx.Sqrt(complex(float64(len(x)), 0))
+	}
+
+}
+
+func newParams(xmax, dt float64, res, timesteps int, im bool) params {
+	par := params{
```
radu-munteanu, line 56: @DavidLoftus `new` makes me think the function returns an object/pointer. I'd either return &params / &operators or change the name to something else. From what I see how these structures are used in main, it's better to return the addresses. I'll explain at the splitOp definition.

contents/split-operator_method/code/go/split_op.go
```diff
@@ -0,0 +1,198 @@
...
+			opr.pe[i] = cmplx.Exp(-0.5 * complex(par.dt, 0) * opr.v[i])
+		} else {
+			opr.ke[i] = cmplx.Exp(complex(0, -0.5*par.dt*math.Pow(par.k[i], 2)))
+			opr.pe[i] = cmplx.Exp(complex(-imag(opr.v[i]), -0.5*par.dt*real(opr.v[i])))
+		}
+	}
+	return opr
+}
+
+func splitOp(par params, opr operators) {
```
radu-munteanu, line 102: This function changes the elements in the given opr structure. The opr parameter should be a pointer. The code works because the `size` element of the opr doesn't change (nor the slices as a whole, just their element's values), but if it would have changed, it wouldn't be changed outside the function, because opr is a copy of the structure given, not a pointer. The changes on the slices work because the copied structure has the same copied pointers of those slices and, also, because the slices don't change (with `append` or with `[:]`), just the element's values do.

ntindle:
[lang: go]

PR #666, opened 2020-01-23
adding FUNDING.yml file

leios:
I was recently accepted into the github sponsors program and would like to add a funding button here and also potentially on the front page. As this is a somewhat significant change to the AAA, it is worth discussing a bit.

This would involve adding a new chapter for sponsors and such and on the main page as well. I am up for editing the current tiers as well, so let me know if you think of better reward tiers: https://github.com/sponsors/leios

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,12 @@
+# These are supported funding model platforms
+
+github: leios
+patreon: # Replace with a single Patreon username
+open_collective: # Replace with a single Open Collective username
+ko_fi: # Replace with a single Ko-fi username
+tidelift: # Replace with a single Tidelift platform-name/package-name e.g., npm/babel
+community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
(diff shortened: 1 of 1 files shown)


leios:
Ok, I have been thinking about how to cycle funding back into the community, and I can only think of 2 ways:
1. Hold all the money until we have enough to hire a "full-time" dev for a few months to work with me as an intern (at MIT, I guess?). The advantage here is that this person could also help with video editing, animations, and other related problems.
2. Give $X rewards for certain key issues (AAA rewrite, reviews, cleaning up language implementations, etc). The advantage is that we can start tacking core issues *now*.

I have also found a number of grants that would certainly work with the AAA in the future if we decide to go down the academic route or create some sort of open research nonprofit.

What do you guys think?

Ellipse0934:
I certainly think you should hold onto it for a full time dev hiring or maybe even treat yourself. 
In the future there might be a significant task which happens to be boring too like say new infrastructure for the gitbook rewrite, having a large sum to hire a dev to complete that task makes more sense to me.

PR #667, opened 2020-01-27
A couple of GoLang code improvements

radu-munteanu:
bogo_sort - there was no need for references everywhere, because a slice is a pointer by nature (values can be changed) and the code doesn't use append or anything else that would change the size, and, possibly, the initial address of the slice

jarvis - the equal function had a typo, perhaps the example should be improved, in the future, to catch this case

### DIFF
--- a/contents/bogo_sort/code/go/bogo_sort.go
+++ b/contents/bogo_sort/code/go/bogo_sort.go
@@ -8,10 +8,10 @@ import (
     "time"
 )
 
-func shuffle(a *[]int) {
-    for i := len(*a) - 1; i > 0; i-- {
+func shuffle(a []int) {
+    for i := len(a) - 1; i > 0; i-- {
         j := rand.Intn(i + 1)
-        (*a)[i], (*a)[j] = (*a)[j], (*a)[i]
+        a[i], a[j] = a[j], a[i]
     }
 }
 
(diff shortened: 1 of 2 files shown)


c252:
Hi, could you please submit separate PRs for each edit? It makes things a bit easier.

radu-munteanu:
@c252 I'm not sure how to do this. The PR is for branches (I see there's a tag option). If I commit something new, wouldn't that be captured in the current PR (similar to how GitLab PR works)? The only thing I can do is to commit one thing, make a PR, then, after the PR gets approved and merged, I can do another PR with a newer commit.

I saw I can put exactly the commit hash in the PR. But, I see it only works for the first commit. For the last one, I can only pick master, and that contains both commits. I guess it has something to do on the git hashing algorithm, u can't make a secondary commit PR without including its predecessor, otherwise it would have a different hash. The only thing that might work is to create separate branches, and do a PR from them, but that would look ugly on the git tree, unless you make some kind of rebase on the upstream, then I'd have to overwrite my whole fork tree.

Let me know what you think.

c252:
It's not that big of a deal, don't worry about it. Just for the future, we try to have 1 implementation per PR.

radu-munteanu:
OK. Noted.

I've made two commits, so, even if you have just one PR, in the commit history, each change is logged individually.

leios:
Thanks for taking a look. I think these are fine changes as well.

radu-munteanu:
@c252 now I see what you mean by one implementation - one PR. You guys squash the commits and make a new commit out of one PR. :)

PR #675, opened 2020-04-18
MAJOR RESTRUCTURING

Trashtalk217:
This has been a bloody ride. I started to merge the building tool with the actual data source and it's a bit of thing alright. But it works. (At least on my machine...)


Trashtalk217:
Known issues:
[ ] data files don't seem to be tracked (see the plotting chapter)

Trashtalk217:
The bibtex parser is more scuffed than I could've possibly imagined...

leios:
Great work! We need a few people to help with the remaining issues. I can try my hand at some of them after the IFS chapter

Bugs:
- [ ] Images not centered
- [ ] Certain links (to CC website) have broken images, but working links
- [ ] No "hot reload" -- every time you change language, the page is refreshed
- [ ] Red box around all unicode characters?

Wishlist:
- Language switcher not obvious (Also not obvious in current AAA)

PudottaPommin:
This is draft for a light update of language selector. Unless you want to do any big updates considering whole page looks ( including HTML ).
Number of columns could be worked out better ( it's done via CSS ), so should be easy change if it needed add or remove columns as columns are calculated as % width of container.

**Mobile**
image

**Desktop**
image

I was also thinking about placing language selector on right side of top bar, which could distinguish it a bit more -> maybe even change background colour of button to something dark with light text colour.

leios:
Right, I agree that the language selector should be more visually distinct. I am definitely supportive of any changes to make that happen!

berquist:
Now that we've switched from GitBook to HonKit, is this worth keeping around?

leios:
That's a good question. We should probably close this for now because the immediate fire has been dealt with, but I think we should keep the project in the back of our minds because there could be a situation where we want to do something honkit doesn't allow.

berquist:
In that case, I think you should make a copy of the remote branch in the main repo, in case the remote ever disappears.

leios:
Good idea / done.

berquist:
Closed in favor of https://github.com/algorithm-archivists/algorithm-archive/pull/720 (for now).

PR #677, opened 2020-04-24
Improved julia implementation of Computus

Vexatos:
Replaced division and flooring with actual floor division. Also prettified formatting and print statements.


Vexatos:
It may be considered whether to use `%` and `÷` instead of `mod` and `fld`.

leios:
Yup, thanks for cleaning this up!

PR #678, opened 2020-04-24
Implemented Computus in Crystal

Vexatos:
Based on the julia implementation.


leios:
If Jie is happy, so am I

PR #682, opened 2020-04-28
Huffman Encoding Scratch Implementation

dovisutu:
I finally find a way to fix the buggy image... just set `transform: scale(1)`.


leios:
I am impressed! Good work! We'll get to a review soon!

jiegillet:
I can do a review, I also implemented (dynamic) Huffman encoding with Scratch (unshared). Is there an actual Scratch project? It would be easier to play with.

dovisutu:
Yeah, I shared it (with extra interacting part which is a simple *enter-text-and-say-result* loop). [The link](https://scratch.mit.edu/projects/389604255/) is already in the chapter.

Maybe drawing the tree out can be better, but I didn't implement it partially because I didn't have enough time. Anyway, that can be done independently without revisiting the chapter and code.

TimothyJu1985:
TimothyJu1985 archived:false

leios:
I am unfortunately not a scratch master like @jiegillet, but can take a look at this, if needed. Let me know!

dovisutu:
Hi, is the review going on?

leios (approved):
At this stage, I checked the project and it looks good! The code also looks good, so I am happy to give this a positive review!

Sorry again for the delay!

PR #686, opened 2020-05-06
Add Split-Operator in Rust

fanninpm:
@Gathros and @strega-nil do you want to help make this code more rusty, especially with those very short for-loops. (I essentially translated the preexisting C++ implementation and kept fighting the borrow checker until it worked.)

### DIFF
--- a/contents/split-operator_method/code/rust/split_op.rs
+++ b/contents/split-operator_method/code/rust/split_op.rs
@@ -8,6 +8,8 @@ use std::fs::File;
 use std::io::Write;
 use std::path::Path;
 
+// This implementation is based on the C and C++ implementations.
+
 #[derive(Clone)]
 struct Parameters {
     xmax: f64,
@@ -104,26 +106,22 @@ fn split_op(par: &Parameters, opr: &mut Operators) {
     let mut density: Vec<f64>;
 
     for i in 0..par.timesteps {
-        // this should use an iterator
         for j in 0..par.res {
             opr.wfc[j] *= opr.pe[j];
         }
 
         fft(&mut opr.wfc, false);
 
-        // this should use an iterator
         for j in 0..par.res {
             opr.wfc[j] *= opr.ke[j];
         }
 
         fft(&mut opr.wfc, true);
 
-        // this should use an iterator
         for j in 0..par.res {
             opr.wfc[j] *= opr.pe[j];
         }
 
-        // this is a good example (c.f. the cpp implementation)
         density = opr.wfc.iter().map(|x| x.norm().powi(2)).collect();
 
         if par.im_time {
--- a/contents/split-operator_method/code/rust/split_op.rs
+++ b/contents/split-operator_method/code/rust/split_op.rs
@@ -1,0 +1,199 @@
+extern crate num;
+extern crate rustfft;
+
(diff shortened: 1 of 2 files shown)


contents/split-operator_method/code/rust/split_op.rs
```diff
@@ -0,0 +1,199 @@
...
+    for i in 0..x.len() {
+        x[i] = y[i] / (x.len() as f64).sqrt();
+    }
+}
+
+fn split_op(par: &Parameters, opr: &mut Operators) {
+    let mut density: Vec<f64>;
+
+    for i in 0..par.timesteps {
+        // this should use an iterator
```
Liikt, line 107: Not sure how you turn this into an iterator as you need the index for accessing a different array than the one you are iterating over. Not saying it is impossible, just harder. I think this is perfectly fine. This also makes the code a bit clearer.

Gathros:
I don't know rust enough to be able to review it, but I think @Liikt can help you.

leios:
Cool, I am happy to merge if @fanninpm is happy as well. Maybe remove the `this should be an iterator` comment?

Thanks for hopping in folks!

Also, no problem if you want to keep fiddling with it!

fanninpm:
If you guys think this is fine as is, I'm happy to see it merged.

leios:
I'll do a final look-over later. Probably on stream.

contents/split-operator_method/code/rust/split_op.rs
fanninpm: I just realized I could've used the [`enumerate()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.enumerate) method, but (as you point out) that may come at the cost of clarity.

leios: Honestly, I don't mind using that either. It just depends on what Rust users are used to reading.

leios:
I ran this locally and got the expected results, so it is good to merge as soon as you are happy with the code (just let me know if you want to use `enumerate()` or not).

As a question, I am not good with rust, how do you normally compile this type of stuff? Would it be worth adding a cargo.toml file or something?

fanninpm:
If this were in a standalone project, I would be using a `Cargo.toml` file and running `cargo build` or `cargo run` in lieu of `rustc`, because it depends on crates not found in the standard library. (Specifically, I make use of the latest versions of [num](https://crates.io/crates/num) and [rustfft](https://crates.io/crates/rustfft) as of 5 May 2020.)

Since I didn't want to disturb the existing directory structure, I developed this on [repl.it](https://repl.it/). Per #691, if you want to add a `Cargo.toml` file, be my guest.

~~Anyway, I'm happy with the code as is.~~ @Liikt would it hurt to remove the two instances of [`extern crate`](https://doc.rust-lang.org/reference/items/extern-crates.html), as they're no longer required in Rust 2018? That would be my only other outstanding reservation I have with the code. (By the way, if you're interested in learning more about iterators vs. for loops in Rust, I'd recommend reading [this section](https://doc.rust-lang.org/book/ch13-04-performance.html) of the Rust Book.)

leios:
I am happy with this PR as is, great work! Thanks for the code!

PR #690, opened 2020-05-06
Tree traversal in MATLAB

foldsters:
All code sections are included. Accidentally deleted some lines during debugging and added them back in.


foldsters:
Accidentally pushed to the wrong branch, fixing now

foldsters:
Okay that should be fixed

foldsters:
I'm not a fan of cell arrays either, but it seems it's the only way to get type-agnostic tree structures.

I'll work on the updates now.

foldsters:
That should be good to go!

PR #693, opened 2020-05-07
Add Thomas Algorithm in Rust

fanninpm:
For leios: This one is implemented without any external crates, so you can compile it just with `rustc`. (Let me know if what I did for pretty-printing on line 37 of the code was too much.)

### DIFF
--- a/contents/thomas_algorithm/code/rust/thomas.rs
+++ b/contents/thomas_algorithm/code/rust/thomas.rs
@@ -1,0 +1,38 @@
+// This implementation is based on the C++ implementation.
+
+fn thomas(a: &[f64], b: &[f64], c: &[f64], x: &mut Vec<f64>) {
+    let size = a.len();
+    let mut y = vec![0.0; size];
+
+    y[0] = c[0] / b[0];
+    x[0] /= b[0];
+
+    for i in 1..size {
+        let scale = 1.0 / (b[i] - a[i] * y[i - 1]);
+        y[i] = c[i] * scale;
+        let x_i = x[i];
+        x[i] = (x_i - a[i] * x[i - 1]) * scale;
+    }
+
+    for i in (0..(size - 1)).rev() {
+        x[i] -= y[i] * x[i + 1];
+    }
+}
+
+fn main() {
+    let a = vec![0.0, 2.0, 3.0];
+    let b = vec![1.0, 3.0, 6.0];
+    let c = vec![4.0, 5.0, 0.0];
+    let mut x = vec![7.0, 5.0, 3.0];
+
+    println!("The system");
+    println!("[{:?} {:?} {:?}][x] = [{:?}]", a[0], b[0], c[0], &x[0]);
+    println!("[{:?} {:?} {:?}][x] = [{:?}]", a[1], b[1], c[1], &x[1]);
+    println!("[{:?} {:?} {:?}][x] = [{:?}]", a[2], b[2], c[2], &x[2]);
+    println!("has the solution");
+
+    thomas(&a, &b, &c, &mut x);
+
+    x.iter()
+        .for_each(|i| println!("[{:>19}]", format!("{:18}", format!("{:?}", i))));
+}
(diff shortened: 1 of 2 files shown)


contents/thomas_algorithm/code/rust/thomas.rs
```diff
@@ -0,0 +1,38 @@
+// This implementation is based on the C++ implementation.
```
Liikt, line 1: I would probably remove this comment

Liikt: Other than that this is perfectly fine

contents/thomas_algorithm/code/rust/thomas.rs
```diff
@@ -0,0 +1,38 @@
...
+
+fn thomas(a: &[f64], b: &[f64], c: &[f64], x: &mut Vec<f64>) {
```
Liikt, line 3: If you change the `&[f64]` to a `Vec<f64>` you can get rid of the references in the main which would be a bit cleaner.

Liikt, line 3: I would also probably return a `Vec<64>` instead of mutating the Vec. Makes it a bit more cleaner to read imho.

so in the end i think the best signature would be `fn thomas(a: Vec<f64>, b: Vec<f64>, c: Vec<f64>, mut x: Vec<f64>) -> Vec<f64>`

berquist: I think we agreed in https://github.com/algorithm-archivists/algorithm-archive/issues/510 that we want functions to be pure, even though at least the Thomas algorithm functions are all over the place in that regard. With that in mind, I think the best signature is
```rust
fn thomas(a: &[f64], b: &[f64], c: &[f64], d: &[f64]) -> Vec<f64> {
```
so that you have the efficiency of the function not taking ownership of its arguments, but also being pure.

contents/thomas_algorithm/code/rust/thomas.rs
fanninpm: With your and @leios's blessing, I would probably go on a crusade to remove all such comments (e.g. "This implementation was written by @strega-nil") throughout the codebase.

leios: Honestly, I think that's a fine idea.

They were originally implemented as a way of letting readers know that this code *could* have been written by someone who did not write previous implementations in the same language.

Now that the review process is more rigorous, it doesn't have as much meaning.

contents/thomas_algorithm/code/rust/thomas.rs
```diff
@@ -0,0 +1,38 @@
...
+    println!("The system");
+    println!("[{:?} {:?} {:?}][x] = [{:?}]", a[0], b[0], c[0], &x[0]);
+    println!("[{:?} {:?} {:?}][x] = [{:?}]", a[1], b[1], c[1], &x[1]);
+    println!("[{:?} {:?} {:?}][x] = [{:?}]", a[2], b[2], c[2], &x[2]);
+    println!("has the solution");
+
+    let y = thomas(&a, &b, &c, &x);
+
+    y.iter()
+        .for_each(|i| println!("[{:>19}]", format!("{:18}", format!("{:?}", i))));
```
berquist, line 37: One question before I override @Liikt and approve. Why is this nested and not just a single `println!`?

fanninpm: I essentially wanted to pretty-print the result in a certain way. (I want positive values to lead with a space.) If you want, I can get rid of the pretty-printing.

berquist: Isn't this the same?
```rust
y.iter().for_each(|i| println!("[{:>19}]", i));
```

fanninpm: Not quite. The result I was looking for satisfies these guidelines:

1. The sign should either be a negative or a space.
2. The decimal point should (ideally) be aligned.

I guess we don't have to pretty-print the result if we don't have to.

berquist: I understand better now. As long as the explanation is in the PR thread like this, it's ok.

PR #696, opened 2020-05-11
adding matlab implementation to monte carlo integration

foldsters:


contents/monte_carlo_integration/code/matlab/monte.m
```diff
@@ -0,0 +1,20 @@
+monte_carlo(10000000);
```
berquist, line 1: This should go at the bottom of the file.

contents/monte_carlo_integration/code/matlab/monte.m
```diff
@@ -0,0 +1,20 @@
...
+    squares_array = xy_array.^2;            
+    
+    % sum the xs and ys and check if it's in the quarter circle
+    incircle_array = sum(squares_array)<1;  
+
+    % determine the average number of points in the circle
+    pi_estimate = 4*sum(incircle_array)/n;  
+    
+    fprintf("The pi estimate is: %f\n", pi_estimate);
+    fprintf("Percent error is: %f%%", 100 * abs(pi_estimate - pi) / pi);
```
berquist, line 18: In general, separate the printing from the computation: have `monte_carlo` compute the estimate for pi, and print the result after the function.

contents/monte_carlo_integration/code/matlab/monte.m
foldsters: I understand that it looks weird, but matlab instructions in an .m file _must_ come before all function definitions.

berquist: Welp, that's what I get for trying to run this in Octave and not knowing that.

contents/monte_carlo_integration/code/matlab/monte.m
foldsters: I was following the julia implementation: it also just prints and does not return the value. I can still make the change but I believe what I have is consistent with what's already in the archive.

contents/monte_carlo_integration/code/matlab/monte.m
foldsters: Haha yeah, matlab is weird.

contents/monte_carlo_integration/code/matlab/monte.m
berquist: I just took a look at a few of the other code samples, and most have the printing outside the function *except* for Julia, so I think the Julia code should be changed.

foldsters: Agreed. I will make the change now.

contents/monte_carlo_integration/code/matlab/monte.m
```diff
@@ -0,0 +1,21 @@
+pi_estimate = monte_carlo(10000000);
```
foldsters, line 1: Changes added!

contents/monte_carlo_integration/code/matlab/monte.m
```diff
@@ -0,0 +1,21 @@
...
+
+fprintf("The pi estimate is: %f\n", pi_estimate);
+fprintf("Percent error is: %f%%", 100 * abs(pi_estimate - pi) / pi);
```
foldsters, line 4: Changes added!

contents/monte_carlo_integration/code/matlab/monte.m
foldsters: I assumed this is resolved then.

contents/monte_carlo_integration/code/matlab/monte.m
berquist: Just one thing, can you add a newline after `Percent error is`?

foldsters:
I can, but is that really necessary? Any other print that happens after running the file in a program will automatically be run on a new line since it's a new command.

berquist:
Yes, but I don't think it looks good:
```
To get started, type doc.
For product information, visit www.mathworks.com.
 
>> run monte
The pi estimate is: 3.141513
Percent error is: 0.002529%>> run monte
The pi estimate is: 3.140780
Percent error is: 0.025855%>> run monte
The pi estimate is: 3.142591
Percent error is: 0.031772%>> 
```
What if the printed line was really long? Your prompt would be on the right-hand side of the screen.

foldsters:
Oh I see. I was using a jupyter terminal and that didn't happen. I'll get that done now.

foldsters:
Change added

foldsters:
chrome_cZUw66XbEq

For reference here's what it looked like for me using a jupyter console. Thanks for checking on the original ide!

contents/monte_carlo_integration/code/matlab/monte.m
leios: Ok. I'll update Julia today. Thanks for the mention of inconsistency!

PR #700, opened 2020-05-21
Add IFS in python

jonathanvanschenck:
I tried to adapt the julia code as closely as possible, with just a few python-specific tweaks.
I opted to only use base python packages so that the code would be usable for more people (with numpy, I think this is a one-liner).
Additionally, I intentionally used some python specific syntax and structuring (like "[x for x in y]" list comprehension and the zip function, etc) because I felt it was truer to the language, than doing the usually thing of making a full for loop block that loops over an index. This whole thing would be much cleaner if base python had element-wise addition. . . *sigh*

### DIFF
--- a/contents/IFS/IFS.md
+++ b/contents/IFS/IFS.md
@@ -128,6 +128,8 @@ Here, instead of tracking children of children, we track a single individual tha
 {% method %}
 {% sample lang="jl" %}
 [import:4-17, lang:"julia"](code/julia/IFS.jl)
+{% sample lang="py" %}
+[import:5-13, lang:"python"](code/python/IFS.py)
 {% endmethod %}
 
 If we set the initial points to the on the equilateral triangle we saw before, we can see the Sierpinski triangle again after a few thousand iterations, as shown below:
@@ -180,6 +182,8 @@ In addition, we have written the chaos game code to take in a set of points so t
 {% method %}
 {% sample lang="jl" %}
 [import, lang:"julia"](code/julia/IFS.jl)
+{% sample lang="py" %}
+[import, lang:"python"](code/python/IFS.py)
 {% endmethod %}
 
 ### Bibliography
--- a/contents/IFS/IFS.md
+++ b/contents/IFS/IFS.md
@@ -129,7 +129,7 @@ Here, instead of tracking children of children, we track a single individual tha
 {% sample lang="jl" %}
 [import:4-17, lang:"julia"](code/julia/IFS.jl)
 {% sample lang="py" %}
-[import:5-13, lang:"python"](code/python/IFS.py)
+[import:5-12, lang:"python"](code/python/IFS.py)
 {% endmethod %}
 
 If we set the initial points to the on the equilateral triangle we saw before, we can see the Sierpinski triangle again after a few thousand iterations, as shown below:
--- a/contents/IFS/IFS.md
+++ b/contents/IFS/IFS.md
@@ -131,7 +131,7 @@ Here, instead of tracking children of children, we track a single individual tha
 {% sample lang="cpp" %}
 [import:39-52, lang:"cpp"](code/c++/IFS.cpp)
 {% sample lang="py" %}
-[import:1-12, lang:"python"](code/python/IFS.py)
+[import:5-12, lang:"python"](code/python/IFS.py)
 {% endmethod %}
 
 If we set the initial points to the on the equilateral triangle we saw before, we can see the Sierpinski triangle again after a few thousand iterations, as shown below:
--- a/contents/IFS/IFS.md
+++ b/contents/IFS/IFS.md
@@ -179,7 +179,7 @@ In addition, we have written the chaos game code to take in a set of points so t
 
 {% method %}
 {% sample lang="jl" %}
-[import:4-17, lang:"julia"](code/julia/IFS.jl)
+[import, lang:"julia"](code/julia/IFS.jl)
 {% endmethod %}
 
 ### Bibliography
--- a/contents/IFS/IFS.md
+++ b/contents/IFS/IFS.md
@@ -131,7 +131,7 @@ Here, instead of tracking children of children, we track a single individual tha
 {% sample lang="cpp" %}
 [import:39-52, lang:"cpp"](code/c++/IFS.cpp)
 {% sample lang="py" %}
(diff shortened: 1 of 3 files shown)


leios:
Not a rigorous review yet, but would you mind outputting the set of points and writing them to file afterwards? I will be modifying the Julia code to do this as well.

This was done in PR #701, and I think it's a good addition

jonathanvanschenck:
For sure!

jonathanvanschenck:
Alright, I made a few tweaks.

First, I opted to make `chaos_game` a python generator rather than to have it build up the list and return the whole thing. If you think it obscures the algorithm too much, I'll switch it to a regular function like you were mentioning above. I'm still pretty new to publishing code, so I'm still learning what code is "clear" and what code is not.

Second, I also added a function to generate the points of a regular N-gon.

contents/IFS/code/python/IFS.py
```diff
@@ -0,0 +1,34 @@
...
+    point = [random(), random()]
+
+    for _ in range(n):
+        # Update the point position and yield the result
+        point = [(_p + _s) / 2 for _p, _s in zip(point, choice(shape_points))]
+        yield point
+
+# This function generates the shape points for an N-gon inscribed inside a
+#  unit circle, with one of the edges oriented along the bottom
+def ngon_shape_points(N):
```
leios, line 16: Though I appreciate this, it is not quite consistent with the other implementations. That said, it's clever and I appreciate that.

There's no real reason for a chaos game to be constrained in this way. People can choose points arbitrarily, so this function *might* give the wrong impression to some readers. I'm inclined to remove it.

contents/IFS/code/python/IFS.py
```diff
@@ -0,0 +1,34 @@
...
+from math import sqrt, sin, cos, pi
+
+# This generator simulates a "chaos game"
+def chaos_game(n, shape_points):
+    # Initialize the starting point
+    point = [random(), random()]
+
+    for _ in range(n):
+        # Update the point position and yield the result
+        point = [(_p + _s) / 2 for _p, _s in zip(point, choice(shape_points))]
```
leios, line 11: I think this is fine, I didn't know `choice()` was a thing before. That's kinda cool.

contents/IFS/code/python/IFS.py
jonathanvanschenck: Good point!
My original thinking was that the article emphasizes regular polygons, and generating the Cartesian vertices of anything other than the triangle and square is nontrivial. In that way, I was thinking of this as a sort of "lemma" for the main algorithm.
That being said, I think I now agree with you that the inclusion of `ngon_shape_points` probably would stifle creativity more than it would encourage people to try things out.
Actually, on that point, it might be interesting to include an example of a non-regular polygon in the article itself!

leios: Yeah, I was thinking of doing this with the restricted chaos game chapter upcoming. I have to take baby steps because of my workload with other projects right now.

jonathanvanschenck:
Removed the `ngon_shape_points` function.

contents/IFS/code/python/IFS.py
jonathanvanschenck: I hear you! Thanks for all the great work you have done, I love the video and it is fun getting to participate

contents/IFS/code/python/IFS.py
```diff
@@ -0,0 +1,25 @@
...
+#     A = (0.0, 0.0)
+#     B = (0.5, sqrt(0.75))
+#     C = (1.0, 0.0)
+# It will output the file sierpinski.dat, which can be plotted after
+shape_points = [[0.0, 0.0],
+                [0.5, sqrt(0.75)],
+                [1.0, 0.0]]
+with open("sierpinski.dat", "w") as f:
+    for point in chaos_game(10000, shape_points):
+        f.write("{0},{1}\n".format(*point))
```
berquist, line 25: For consistency with the Julia implementation, the plot output will be easier to read if it's tab-separated:
```suggestion
        f.write("{0}\t{1}\n".format(*point))
```

contents/IFS/code/python/IFS.py
```diff
@@ -0,0 +1,25 @@
...
+from math import sqrt, sin, cos, pi
+
+# This generator simulates a "chaos game"
+def chaos_game(n, shape_points):
+    # Initialize the starting point
+    point = [random(), random()]
+
+    for _ in range(n):
+        # Update the point position and yield the result
+        point = [(_p + _s) / 2 for _p, _s in zip(point, choice(shape_points))]
```
berquist, line 11: You don't need the underscores here, since `_p` and `_s` can't escape the scope of the list comprehension. Up to you whether or not to keep them.

contents/IFS/code/python/IFS.py
```diff
@@ -0,0 +1,25 @@
...
+        yield point
```
berquist, line 12: :+1:

contents/IFS/IFS.md
```diff
@@ -130,6 +130,8 @@ Here, instead of tracking children of children, we track a single individual tha
 [import:4-17, lang:"julia"](code/julia/IFS.jl)
 {% sample lang="cpp" %}
 [import:39-52, lang:"cpp"](code/c++/IFS.cpp)
+{% sample lang="py" %}
```
berquist, line 133: I would fix the range so that the imports in the source code file aren't included.

jonathanvanschenck: okay!

contents/IFS/code/python/IFS.py
jonathanvanschenck: I'll change them

contents/IFS/code/python/IFS.py
jonathanvanschenck: done

contents/IFS/code/python/IFS.py
```diff
@@ -0,0 +1,25 @@
+from random import random, choice
+from math import sqrt, sin, cos, pi
```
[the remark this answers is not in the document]
jonathanvanschenck: Oh! yup, my bad

contents/IFS/code/python/IFS.py
jonathanvanschenck: :)

PR #702, opened 2020-05-22
fixing Julia code and embedding latest video

leios:
This fixes the Julia code to match #700 and #701, and also embeds the lated youtube video


leios:
Great, Thanks for that!

PR #703, opened 2020-05-23
Add computus in C++

VincentZalzal:


VincentZalzal:
Sure, no problem. I also added braces to the for-loop, for consistency. Would be nice to have a style guide somewhere for that kind of stuff. At least, clang-format handles a big chunk of that already.

VincentZalzal:
Sorry, my comment was unclear. I meant, at least you provide a clang-format config, which handles much style issues already, which is nice. I used it before submitting :)

PR #704, opened 2020-05-24
[Haskell] IFS

jiegillet:
I could not resist generalizing the algorithm a little bit, in my version Hutchinson Operators are an argument of the function. 

I used [Ormolu](https://github.com/tweag/ormolu) as the code formatter for the first time. I like formatter without config files so I'll probably keep using that.


ntindle:
[lang: haskell]

rampion (approved):

jiegillet:
Thank you @rampion.

leios (approved):
We have an approval (admittedly from someone I have not seen before), but as long as 2 Haskellers are happy, I am happy as well. We can update it later if need be

PR #706, opened 2020-05-24
Update C++ Jarvis March to use points from Graham Scan

berquist:


berquist:
I should have commented here, but https://github.com/algorithm-archivists/algorithm-archive/pull/708#issuecomment-633322788 needed to be resolved before this got merged.

PR #708, opened 2020-05-24
Update JavaScript Jarvis March to use points from Graham Scan

berquist:


leios:
I think @jiegillet did something with the Haskell code such that these numbers were not "magic," but instead pulled from some distribution, but I don't remember the trick. I think he just randomly picked a distribution from within a circle, but I don't remember how he knew which points should be excluded.

I can look into this in a bit.

Thanks a bunch for all the PR work recently!

berquist:
I see. If we can find out, then that information should be at the end of the chapter. Of course, hold off on all the related PRs until we get that information.

jiegillet:
I made the points this way:

```haskell
let pts = [(x,y) | x<-[-5..5], y<-[-5..5], x^2+y^2<=5^2]
```

which means, take all points on a square grid from (-5,-5) to (5,5) and keep the ones inside or on a circle of radius 5.

ntindle:
[lang: javascript]

Amaras (approved):
LGTM, I'll merge this soon

PR #709, opened 2020-05-24
Update Julia Jarvis March to use points from Graham Scan

berquist:

It revealed a bug in this implementation that needs to be fixed.

### DIFF
--- a/contents/jarvis_march/code/julia/jarvis.jl
+++ b/contents/jarvis_march/code/julia/jarvis.jl
@@ -48,8 +48,11 @@ function jarvis_march(points::Vector{Pos})
 end
 
 function main()
-
-    points = [Pos(2,1.5), Pos(1, 1), Pos(2, 4), Pos(3, 1)]
+    points = [
+        Pos(-5, 2), Pos(5, 7), Pos(-6, -12), Pos(-14, -14), Pos(9, 9),
+        Pos(-1, -1), Pos(-10, 11), Pos(-6, 15), Pos(-6, -8), Pos(15, -9),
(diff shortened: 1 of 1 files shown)


leios:
Could you elaborate on the bug a bit?

berquist:
Sure. The output from the Python code is
```
(-14, -14)
(15, -9)
(9, 9)
(0, 14)
(-6, 15)
(-10, 11)
```
but, from the same initial points, the Julia output is
```
Pos[Pos(-14.0, -14.0), Pos(15.0, -9.0), Pos(-6.0, 15.0), Pos(-14.0, -14.0)]
```
I haven't bothered to debug this. If someone sees the obvious fix, feel free to modify this PR or issue a new one.

leios:
So I was looking at this today. There is definitely a bug in some implementations, but it is unclear what that is because the python code is radically different. We can rewrite the julia code to match the python code, but right now we probably need a better way of defining all the points to find the convex hull.

I really need to revise these chapters anyway, so I'll bump that up on the list.

I feel like we had this discussion somewhere on github before, but I can't find it. Sorry!

berquist:
The discussion was in https://github.com/algorithm-archivists/algorithm-archive/pull/708#issuecomment-633322788. What solution do you think is better, that one, or this hard-coded set? The points inside or on a circle from a square grid makes more sense to me.

leios:
I agree. The only caveat is that we need to somehow ensure all output is the same. A hard-coded set is easier to do that with.

berquist:
Yes, also my thoughts. Well, we are already going on a system of trust with PR review. One thing that might help is sorting the output. Ideally the output is also not big, which is easily checked.

ntindle:
[lang: julia]

PR #715, opened 2020-06-12
Gale-Shapley implementation in Rust

lazyprop:
This is, more or less, a translation of the Python implementation.

### DIFF
--- a/contents/stable_marriage_problem/code/rust/stable_marriage.rs
+++ b/contents/stable_marriage_problem/code/rust/stable_marriage.rs
@@ -1,136 +1,107 @@
-#[derive(Eq, PartialEq, PartialOrd)]
+use std::vec::Vec;
+use std::collections::HashMap;
+use std::fmt;
+
+#[derive(Hash, Eq, PartialEq, Copy, Clone)]
+struct PersonId(pub char);
+
+impl fmt::Display for PersonId {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+#[derive(Clone)]
 struct Person {
-    name: char,     // a single character denoting the person e.g 'A', 'B', 'F'
-    index: usize,   // index in men/women vecotr
-    preference: Vec<usize>,
-    pref_index: usize,
-    candidates: Vec<usize>, // all those men/women who have proposed to self
-    partner: Option<usize>
+    id: PersonId,
+    partner: Option<PersonId>,
+    preferences: Vec<PersonId>,
 }
 
-impl Person {
-    fn propose_to_next(&mut self, women: &mut Vec<Person>) {
-        /* propose to next most preferred person */
+fn gale_shapley(people: Vec<Person>) -> HashMap<PersonId, PersonId> {
+    let mut stable_matching: HashMap<PersonId, PersonId> = HashMap::new();
 
-        if self.pref_index >= self.preference.len() {
-            ()
-        }
+    for person in people {
+        let partner = stable_matching
+            .get(&person.id)
+            .or_else(||     // execute this if person doesn't have partner
+                person.preferences
+                .iter()
+                // skip all preferences which already have a pair
+                .skip_while(|id| stable_matching.contains_key(id))
+                .next()
+            )
+            .map(|&id| id);
 
-        let person = self.preference[self.pref_index];
-        // add self to the preferred woman's candidates
-        women[person].candidates.push(self.index);
-
-        self.pref_index += 1;
-
-    }
-
-    fn pick_preferred(&mut self, men: &mut Vec<Person>) {
-        /* pick the highest preferred man among candidates */
-
-        for person in &self.preference {
-            if Some(*person) == self.partner {
-                break
-            } else if self.candidates.contains(&person) {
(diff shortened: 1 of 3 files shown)


Amaras:
Hello, thanks for your first PR here! As I'm just a contributor here, and I've only just started learning Rust, I cannot really review your code. That said, it does seem quite clean. I have not tried to build it, but I trust you on that for now before any real review is done.

Problem is, there is an open PR (https://github.com/algorithm-archivists/algorithm-archive/pull/687) on my part to improve the Python implementation, so you might want to branch off from simply translating the current Python implementation, and maybe write more idiomatic Rust if possible.
The review in that PR might also give you some ideas to improve your code from the direct translation it seems to be.

In any case, thanks again for your interest in the AAA!

n1trob4ct3r:
I have already kept in mind the changes that you have made in the Python implementation.

- No 'getters/setters' are used. There is just an `initperson` function which I think is necessary because initializing a person requires a lot of repetitive lines.
- Currently, the number and preferences of people hardcoded because a variable number of people would require randomly generating preferences. The problem is that `rand` is an external crate in Rust which has to be downloaded from the Internet. 
- I am new to Rust myself. Nevertheless, I have tried my best to write idiomatic Rust.
- I have updated `println!("Pairs: ")` to `println!("Stable Matching: ")` to make the output more descriptive.
- I have added my name to CONTRIBUTORS.md

contents/stable_marriage_problem/code/rust/stable_marriage.rs
```diff
@@ -0,0 +1,136 @@
...
+        cont = false;
+        for man in &men {
+            if man.partner.is_none() {
+                cont = true;
+                break;
+            }
+        }
+    }
+
+    println!("\nStable Matching: ");
```
Liikt, line 126: This is only personal preference but i would rather write
```rust
print!("\n");
print!("Stable Matching: \n");
```
than `println!("\nStable Matching: ");`

contents/stable_marriage_problem/code/rust/stable_marriage.rs
```diff
@@ -0,0 +1,136 @@
...
+    for man in &men {
+        match man.partner {
```
Liikt, line 128: You can do a simple unwrap here as all man must have a partner at this point.

n1trob4ct3r:
@Liikt I have rewritten the entire implementation. It is much more concise and 'Rusty' now. The initial declaration and display of the vectors takes up a lot of lines though.

contents/stable_marriage_problem/code/rust/stable_marriage.rs
```diff
@@ -0,0 +1,107 @@
...
+    let stable_matching: HashMap<PersonId, PersonId> = gale_shapley(people);
+
+    // display stable matches
+    for man in men {
+        let partner_id = stable_matching.get(&man.id);
+        match partner_id {
+            Some(p) => println!("{} + {}", man.id, p),
+            None => ()
+        }
+    }
```
fanninpm, line 106: I get the feeling that `iter` and [`.filter_map()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.filter_map) might be useful here. (I'm not sure, though.)

n1trob4ct3r: @fanninpm How would we use `.filter_map()` here? Won't it just over complicate things? We will have to iterate and then collect into vector and then print the vector. I am not sure though. Is there a better way to do this?

fanninpm: @n1trob4ct3r You can use [`.for_each()`](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.for_each) on the iterator, but again, I doubt its usefulness. Besides, you refer to both `man.id` and `partner_id` in your solution.

n1trob4ct3r: @fanninpm Yes that would be a problem too.

Is there anything else? Is it ready for merging?

contents/stable_marriage_problem/code/rust/stable_marriage.rs
```diff
@@ -0,0 +1,107 @@
+use std::vec::Vec;
+use std::collections::HashMap;
+use std::fmt;
+
+#[derive(Hash, Eq, PartialEq, Copy, Clone)]
+struct PersonId(pub char);
+
+impl fmt::Display for PersonId {
```
jonay2000, line 8: Maybe you want to just derive debug, instead of manually implementing display, to simplify the entire code a bit. Even if you don't want to remove the Display implementation, deriving debug is never a bad practice

jonay2000: I now see that your code relies on the exact representation quite a bit, so then maybe just add a derive debug?

contents/stable_marriage_problem/code/rust/stable_marriage.rs
jonay2000: I think rust recommends to use just a `println!()` invocation without parameters. I think clippy even has a lint for that.

so:

```rust

println!();
println!("Stable Matching: ");
```

PR #720, opened 2020-06-30
Use HonKit instead of GitBook

fanninpm:
From what I've seen, [HonKit](https://github.com/honkit/honkit) appears to pick up where GitBook left off in development. Things were pretty much smooth sailing except for the fact that `honkit install` didn't exist (this was so that they could remove their reliance on `npm` so others could use `yarn`, for instance). I then [learned](https://github.com/honkit/honkit/issues/72) that I had to create the `package.json` file below.

Alongside this pull request, I would also recommend changing [this Wiki page](https://github.com/algorithm-archivists/algorithm-archive/wiki/How-to-Contribute) to reflect the change to HonKit. Below, I have provided some sample text as a basis for what the relevant sections would look like.
____

## Step 1.2 - HonKit

The next thing we need is [HonKit](https://github.com/honkit/honkit) (successor to GitBook), which is the tool that builds the AAA. By downloading HonKit, you can also build and serve the AAA locally on your computer.

HonKit can be retrieved via NPM or yarn. If you don't have NPM/Node.js installed on your system, you can download it here: <https://nodejs.org/en/>.

## Step 2.3 - Building and running the AAA

Now that you have the AAA-repo on your machine, you first need to get all the packages the AAA requires to build. This can be done by running the following command in the same directory as your copy of the AAA-repo:
```sh
npm install  # or `yarn install`
```
After the all packages are downloaded, you can run/serve the AAA with this command:
```sh
npm run serve  # or `yarn run serve`
```
Once HonKit finishes building and starts serving, your terminal should've printed out something like the following: 

This indicates that you successfully built the AAA! You can check out your local AAA by navigating to http://localhost:4000 in your browser.

Here you can see the AAA, just like it is on <https://algorithm-archive.org>:

When you're done with your local server, you can use `ctrl + C` to close the server.

## Step 3.3 - Adding code to your AAA-repo

First of all, there are some things you need to know ("Skip a bit, brother.")

…

The instances mark places where code gets imported into the page. The very last place should import all files you created. The other places often just target specific functions (for example just a `Sort()` function as an explanation, but not all other code). To identify which exact function is needed, run the book (described in Step 2.3) via `npm run serve` and go to each code example.

…

The `multLangIdentifier` is important for HonKit to provide the language switcher that you see at the top of the algorithm pages. The `multLangIdentifier` is simply the file ending of the language you want to add.

…

Now you can run `npm run serve` again (or just let it rebuild automatically, if it is still running) and look at the page of the algorithm you're targeting. You should now be able to select your chosen language at the top and you should see your code in all places where code examples should be.

…

## Step 3.7 - After the Pull Request: The Review

So, after someone reviewed your PR, you can just go ahead and correct the things the person pointed out. Doing that is fairly simple. Just checkout your branch locally (via `git checkout your-branch-name`), make corrections, run the book to check that everything is correct (via `npm run serve`) and then finally push the changes to your branch (via `git push`). Since the PR uses your branch, the changes immediately show up after you pushed them.


fanninpm:
Here's the Travis build for commit efc855e: https://travis-ci.org/github/algorithm-archivists/algorithm-archive/builds/703556596

fanninpm:
Here's the Travis build for commit 1c6c4f3, which **passed**: https://travis-ci.org/github/algorithm-archivists/algorithm-archive/builds/703575094

leios:
This looks really promising! Is it ready for review?

fanninpm:
@leios This is ready; however, I'll be offline for the next couple of hours.

leios:
Looks great! For some reason, the images are left-justified. That's the only problem I could find after building it locally. I'll dig deeper a bit later.

fanninpm:
@leios Does the latest commit work? It uses the latest version of HonKit, which fixes the problem you were seeing (c.f. <https://github.com/honkit/honkit/issues/89>). You'd probably have to run `npm i` followed by `npm run build --reload` (or `yarn install` followed by `npm run build --reload` if you use `yarn` instead of NPM).

leios:
As a few quick notes:

The images are still left-justified, when they should be center-justified. All imaged have `class="center"` set in html, so I am not sure how to fix this.

In addition,the word count seems to be broken as well.

leios:
Also, as a note: I can no longer build with gitbook on my computer because of an update, so this would be nice to get in soon. I'll probably merge tomorrow if there are no other complaints.

leios:
WOO! I'll let you know if the build fails for whatever reason.

c252:
Looks great!

fanninpm:
For those of you looking at this PR and wondering how I got the CI to work, that was not me. Props to @berquist for solving the issues with our current CI setup in #728 and #729.

PR #722, opened 2020-07-04
IFS Common Lisp implementation

Trashtalk217:
The outputting to "out.dat" is slightly dirty and can maybe be improved. Otherwise I'm pretty proud of this one.


contents/IFS/code/clisp/ifs.lisp
```diff
@@ -0,0 +1,26 @@
+;;;; Iterated Function System implementation
+
+(defstruct (point (:constructor make-point (x y))) x y)
+
+(defun chaos-game (iterations shape-points)
+  (loop
+    repeat iterations
+    for rand-point = (svref shape-points (random (length shape-points)))
+    for point = (make-point (random 1.0) (random 1.0)) ; starting point
```
berquist, line 9: I think this creates a new `point` on every iteration of the loop, which none of the other implementations do. Can you change it?

Trashtalk217: Do you mean the `(make-point (random 1.0) (random 1.0))`, because that's only run once. I don't exactly understand what you mean.

ntindle:
[lang: lisp]

contents/IFS/code/clisp/ifs.lisp
berquist: Well, there must be something here that I don't understand, because the code clearly works. I'm fine with it then.

contents/IFS/IFS.md
```diff
@@ -134,6 +134,8 @@ Here, instead of tracking children of children, we track a single individual tha
 [import:5-12, lang:"python"](code/python/IFS.py)
 {% sample lang="c" %}
 [import:18-29, lang:"c"](code/c/IFS.c)
+{% sample lang="lisp" %}
+[import:5-13, lang:"lisp"](code/clisp/ifs.lisp)
```
berquist, line 138:
```suggestion
[import:5-14, lang:"lisp"](code/clisp/ifs.lisp)
```

berquist (changes requested):
Looks good, just a problem with the displayed lines in the text.

leios (approved):
Merged in the suggestion from berquist, so we should be good to merge!

PR #724, opened 2020-07-04
Computus Common Lisp Implementation

Trashtalk217:
I'm very proud of this one. I took some time to clean it up right away and I'm happy with the result.

### DIFF
--- a/contents/computus/code/clisp/gauss-easter.lisp
+++ b/contents/computus/code/clisp/gauss-easter.lisp
@@ -1,0 +1,34 @@
+;;;; Gauss's Easter algorithm implementation
+
+(defun computus (year &optional (servois nil))
+  "Calculates the day of Easter for a given year and optionally its Servois number"
+  (let*
+    ((a (mod year 19))                  ; year's position on the 19 year metonic cycle
+      (k (floor year 100))              ; century index
+      (p (floor (+ 13 (* 8 k)) 25))     ; shift of metonic cycle, add a day offset every 300 years
+      (q (floor k 4))                   ; correction for non-observed leap days
+      (m (mod (+ 15 (- p) k (- q)) 30)) ; correction to starting point of calculation each century
+      (d (mod (+ (* 19 a) m) 30))       ; number of days from March 21st until the full moon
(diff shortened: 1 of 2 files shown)


contents/computus/computus.md
```diff
@@ -285,6 +285,18 @@ For now, we have the code outputting a tuple of $$d$$ and $$e$$, so users can us
 [import, lang:"python"](code/python/gauss_easter.py)
 {% sample lang="crystal" %}
 [import, lang:"crystal"](code/crystal/gauss_easter.cr)
+(defun computus (year &optional (servois nil))
```
leios, line 288: I think this was a copy and paste error.

contents/computus/code/clisp/gauss-easter.lisp
```diff
@@ -0,0 +1,34 @@
...
+      (m (mod (+ 15 (- p) k (- q)) 30)) ; correction to starting point of calculation each century
+      (d (mod (+ (* 19 a) m) 30))       ; number of days from March 21st until the full moon
+      (n (mod (+ 4 k (- q)) 7))         ; century-based offset in weekly calculation
+      (b (mod year 4))                  ; correction for leap days
+      (c (mod year 7))                  ; also a correction for leap days
+      ;; days from d to next Sunday
+      (e (mod (+ (* 2 b) (* 4 c) (* 6 d) n) 7)))
+    ;; historical corrections for April 26 and 25
+    (when (or (and (eql d 29) (eql e 6)) (and (eql d 28) (eql e 6) (> a 10)))
+      (setf e -1))
```
berquist, line 19: Is it possible to do the same thing as in the Rust implementation where you can use a conditional as an expression?

Trashtalk217: It's a bit tricky to do that, because the value of e is necessary for the conditional. I could definitely do it, but I feel it wouldn't help much. The Rust implementation also has to calculate a temp_e to make it work out.

berquist: Yep, not worth it then.

PR #726, opened 2020-07-05
WIP: Further improved stable marriage in python

jonay2000:
In Discord we discussed the original implementation of gale shapley, and the improvement @Amaras made (#687 ). This is just a draft which improves it further, making it more similar to the rust implementation. It's very different to the original one so you'll just have to say what you think about it.

### DIFF
--- a/contents/stable_marriage_problem/code/python/stable_marriage.py
+++ b/contents/stable_marriage_problem/code/python/stable_marriage.py
@@ -1,25 +1,19 @@
 # Submitted by Marius Becker
+# Updated by Amaras
+# Updated again by Jonathan Dönszelmann
 
-import sys
 from random import shuffle
 from copy import copy
-from string import ascii_uppercase
+from string import ascii_uppercase, ascii_lowercase
+
 
 def main():
-    # Set this to however many men and women you want
-    if len(sys.argv) > 1:
-        num_pairs = int(sys.argv[1])
-    else:
-        num_pairs = 5
-
-    # There are only 26 possible names
-    if num_pairs > 13:
-        print('You can\' have more than 13 pairs.')
-        return
+    # Set this to however many men and women you want, up to 26
+    num_pairs = 5
 
     # Create all Person objects
-    men = [ Person(name) for name in ascii_uppercase[0:num_pairs] ]
-    women = [ Person(name) for name in ascii_uppercase[num_pairs:num_pairs*2] ]
+    men = [Person(name) for name in ascii_uppercase[:num_pairs]]
+    women = [Person(name) for name in ascii_lowercase[:num_pairs]]
 
     # Set everyone's preferences
     for man in men:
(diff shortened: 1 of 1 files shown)


contents/stable_marriage_problem/code/python/stable_marriage.py
```diff
@@ -31,101 +25,58 @@ def main():
...
+    for person in people:
+        if person.partner is None:
+            # find someone in this person's preferences who
+            # doesn't already have a partner.
+            for possible_partner in person.preference:
+                if possible_partner.partner is None:
+                    partner = possible_partner
+                    break
+            else:
+                raise Exception(f"Couldn't find a partner for {person}")
```
jonay2000, line 62: Not quite sure if this should be an error

jonay2000: nor if else is the right construct here, it's quite obscure

jonay2000:
We determined that this algorithm doesn't show as clearly as previous implementations, how marriages are formed. Add to that the fact that it might not work, I don't think this PR should ever be merged, and I might close it soon.

Amaras:
The first time I ran the algorithm, I got this output:
`Preferences of the men:
A: b, c, e, d, a
B: a, e, b, c, d
C: b, a, d, e, c
D: c, b, e, d, a
E: b, c, e, d, a

Preferences of the women:
a: E, C, A, D, B
b: D, B, A, E, C
c: C, D, E, A, B
d: E, D, C, A, B
e: E, D, A, B, C


The algorithm gave this solution:
A + b
B + a
C + d
D + c
E + e`
Problem is: C prefers a over d, and a prefers C over B.
This pairing is thus unstable, so the algorithm is wrong.
As it's inspired by the Rust PR (#715), I recommend a careful analysis of the output code before considering merging it.

n1trob4ct3r:
> Let Alice and Bob both be engaged, but not to each other. Upon completion of the algorithm, it is not possible for both Alice and Bob to prefer each other over their current partners. If Bob prefers Alice to his current partner, he must have proposed to Alice before he proposed to his current partner. If Alice accepted his proposal, yet is not married to him at the end, she must have dumped him for someone she likes more, and therefore doesn't like Bob more than her current partner. If Alice rejected his proposal, she was already with someone she liked more than Bob.

This is from the [Gale-Shapely WIkipedia page](https://en.wikipedia.org/wiki/Gale%E2%80%93Shapley_algorithm#Solution)

Stability here means that there are no free men left. This is because engaged men do not propose and a change of partners only takes place if there is a free man who proposes to a woman and the woman likes him over her current partner.\
The men propose in the order of their preference. And if they did get dumped, that must have been for a reason. So if the algorithm ends, the produced matching must be stable.

Amaras:
> Stability here means that there are no free men left.
No, that's not what stability means. It means that you can't find a pair like C - a where they prefer each other over their partner.

It is right that in the GS algorithm, the fact that there is no free man left means that the algorithm must stop, and thus that (as a consequence of the construction of the pairing) the pairing is stable. It doesn't mean that "all men have a partner" implies that "the pairing is stable".

PR #730, opened 2020-07-11
Huffman encoding in Coconut

Amaras:
This is a first draft of what the Huffman encoding would look like in Coconut.

Sadly, I haven't been able to build the tree correctly yet, so the codebook isn't correct, but it seems to be correct from the tree given.

So the output for now (barring the debug `print` statements) is:
```
message: bibbity_bobbity
huffman tree: Node(Node(Node(Leaf(i, 3), Leaf(b, 6)), Node(Leaf(y, 2), Leaf(t, 2))), Node(Leaf(o, 1), Leaf(_, 1)))
codebook: [('i', '000'), ('b', '001'), ('y', '010'), ('t', '011'), ('o', '10'), ('_', '11')]
```
However, the code says that the output should be:
```
message: bibbity_bobbity
huffman_tree: ['b', [[['_', 'o'], 'y'], ['t', 'i']]]
codebook: [('b', '0'), ('_', '1000'), ('o', '1001'), ('y', '101'), ('t', '110'), ('i', '111')]
```
I don't see my mistake yet, so I'll leave this PR as a draft for now

### DIFF
--- a/contents/huffman_encoding/code/coconut/huffman.coco
+++ b/contents/huffman_encoding/code/coconut/huffman.coco
@@ -49,33 +49,6 @@ def build_huffman_tree(message):
 
     return trees[0]
 
-# constructs the mapping with recursion
-"""def build_codebook(tree, code=''):
-
-    frequencies = Counter(message)
-    trees = frequencies.most_common()
-
-    # while there is more than one tree
-    while len(trees) > 1:
-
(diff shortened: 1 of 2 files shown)


Amaras:
So I just changed a few things (namely popping from the other side), it should be good to go.

PR #735, opened 2020-07-15
Added Bogosort in Coconut

Amaras:
Well, it's not really interesting, but I used more idiomatic Python for the `is_sorted` function, iterating over the array and a shifted copy of the array, rather than over the indices.
That's basically all I did from the Python implementation, except removing the `main` function, that serves little purpose in Python.

### DIFF
--- a/contents/bogo_sort/code/coconut/bogo.coco
+++ b/contents/bogo_sort/code/coconut/bogo.coco
@@ -1,12 +1,7 @@
 import random
 
 
-def is_sorted(a):
-    for x, y in zip(a, a[1:]):
-        if x > y:
-            return False
-    return True
-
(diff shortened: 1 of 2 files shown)


contents/bogo_sort/code/coconut/bogo.coco
```diff
@@ -0,0 +1,20 @@
+import random
+
+
+def is_sorted(a):
+    for x, y in zip(a, a[1:]):
+        if x > y:
+            return False
+    return True
```
berquist, line 8: I would like to see a full FP solution rather than the Python solution. Folding the map comes to mind.

Amaras: why `reduce` when you already have `all` built-in?
Thanks for the more functional style requests!

berquist (approved):

PR #736, opened 2020-07-17
Added Bubblesort in Coconut

Amaras:
Same as for Bogosort, I just slightly improved the Python code and removed the `main` function.
I opened those two PRs mainly to have them in Coconut, after thinking a bit about it. 

(Also, I forgot to click the button, so that's why this one arrived late compared to the Bogosort one; I implemented them the same day)

### DIFF
--- a/contents/bubble_sort/code/coconut/bubblesort.coco
+++ b/contents/bubble_sort/code/coconut/bubblesort.coco
@@ -1,0 +1,15 @@
+import random
+
+def bubble_sort(array):
+	length = len(array)
+	for i in range(length):
+		for j in range(length - i - 1):
+			if array[j] > array[j + 1]:
+				array[j], array[j + 1] = array[j + 1], array[j]
+
+
+if __name__ == '__main__':
+	array = range(10) |> map$(-> random.randint(0, 1000)) |> list
+	print('Before sorting:', array)
+	bubble_sort(array)
+	print('After sorting:', array)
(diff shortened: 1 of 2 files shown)


contents/bubble_sort/code/coconut/bubblesort.coco
```diff
@@ -0,0 +1,15 @@
+import random
+
+def bubble_sort(array):
+	length = len(array)
+	for i in range(length):
+		for j in range(length - i - 1):
+			if array[j] > array[j + 1]:
+				array[j], array[j + 1] = array[j + 1], array[j]
```
berquist, line 8: I would rather see a full FP solution rather than a Python translation, even if it is one of the solutions in https://codereview.stackexchange.com/questions/197868/bubble-sort-in-haskell.

contents/bubble_sort/code/coconut/bubblesort.coco
```diff
@@ -0,0 +1,15 @@
...
+
+
+if __name__ == '__main__':
+	array = range(10) |> map$(-> random.randint(0, 1000)) |> list
```
berquist, line 12: Make sure all the tabs are replaced with spaces and that your editor is set up to use the EditorConfig file.

Amaras:
I was going to quit the functional style, but it seems it works quite well. Maybe not for huge lists, but close enough.
I could do an iterator sorting version if requested.

berquist (approved):
This is pretty wild. Thanks!

PR #737, opened 2020-07-22
Forward Euler method in Coconut

Amaras:
Okay, so the commits show two possible versions of the Forward Euler method in Coconut.
Before editing the md file, I want to have some feedback on that.
I'm not sure I can have something more functional style than that, but it could be.

If there's no way, which version do you prefer? V1 is more concise and efficient for checking, but might be more confusing to beginners. V2 could maybe be made a little bit more functional style, and doesn't need creating a list or an array.

### DIFF
--- a/contents/forward_euler_method/code/coconut/euler.coco
+++ b/contents/forward_euler_method/code/coconut/euler.coco
@@ -22,6 +22,6 @@ if __name__ == '__main__':
     n = 100
     threshold = 0.01
 
-    result = list(forward_euler(time_step, n))
+    result = forward_euler(time_step, n)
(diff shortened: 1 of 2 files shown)


berquist:
Could you write a more functional version of 2?

Amaras:
I think I might be able to. I'll try when I have more time to do it.
Thanks for all the reviews!

Amaras:
@berquist if you have some time, could you do a review of this PR? I'm trying to merge and close non-active PRs (i.e. where reviews have not been addressed in a while)

berquist (approved):

PR #745, opened 2020-08-07
Add flood fill in Rust 

fanninpm:
This implementation makes use of the [`ndarray`](https://crates.io/crates/ndarray) crate, which is accessible from the [Rust Playground](https://play.rust-lang.org/).

### DIFF
--- a/contents/flood_fill/code/rust/flood_fill.rs
+++ b/contents/flood_fill/code/rust/flood_fill.rs
@@ -1,5 +1,5 @@
-use std::collections::VecDeque; // 0.13.1
-use ndarray::prelude::*;
+use std::collections::VecDeque;
+use ndarray::prelude::*; // 0.13.1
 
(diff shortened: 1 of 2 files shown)


ntindle:
[lang: rust]

ShadowMitia (changes requested):
I've taken the liberty of adding a `Cargo.toml` file to your project, as we are moving to an automated build system. You could still see if a more recent version of `ndarray` works, though!

Overall it looks good, I don't really know `ndarray` and I think it's overkill to use it here. 
Maybe you have a good reason to recommend it?

Otherwise just a few changes, `clippy` had a few warnings that need to be fixed.

ShadowMitia:
And you also need to solve conflicts in the `.md`, but I think all the line numbers are good.

ujjwal3067:
It's very old PR. Are we still going to continue reviewing it ? author doesn't seems to be active on this PR for 2 years.

fanninpm:
I don't particularly have time to go back through this PR and make changes. Feel free to supersede/cherry-pick.

PR #748, opened 2020-09-02
Flood fill in Javascript using p5.js

0xJonas:
Implementation of flood fill using the p5.js framework. 

I thought it would be nice to have a more visual showcase of the algorithms and p5.js makes this very easy.
If using p5.js is not acceptable, I'm fine with redoing it in node.js in a similar way to the Julia/C implementations. The actual algorithms can stay the same either way.

Link to the p5 sketch: https://editor.p5js.org/delphi1024/sketches/flzb29W1o

### DIFF
--- a/contents/flood_fill/code/javascript/flood_fill.js
+++ b/contents/flood_fill/code/javascript/flood_fill.js
@@ -14,13 +14,11 @@ function isColor(x, y, color) {
 }
 
 function recolorPixel(x, y, oldColor, newColor) {
-  if (isInBounds(x, y) && isColor(x, y, oldColor)) {
-    const index = (x + y * width) * 4
-    pixels[index] = red(newColor)
-    pixels[index + 1] = green(newColor)
-    pixels[index + 2] = blue(newColor)
-    pixels[index + 3] = alpha(newColor)
-  }
+  const index = (x + y * width) * 4
+  pixels[index] = red(newColor)
+  pixels[index + 1] = green(newColor)
+  pixels[index + 2] = blue(newColor)
+  pixels[index + 3] = alpha(newColor)
 }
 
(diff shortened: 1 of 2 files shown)


leios:
To be honest, we haven't really decided what to do with p5.js yet. There is another PR open that didn't really get anywhere on this topic where we were split on whether p5 was a *new* language or not.

I will say that to merge this PR, I would like to see the code closer to the other implementations. We used Float64 values and chose the example for pedagogical reasons.

I am definitely up for discussion here!

0xJonas:
I think that an algorithm like flood fill benefits greatly from a graphical output, especially if the language supports this without too much boilerplate code. p5.js is one way to do it and, since it also has a web-based editor, one can try out and experiment with a sketch very quickly.

On the other hand I understand that it is a good idea to keep the implementations consistent across languages, including possible testing code. Using external libraries in general might also be an issue (if they are not related to the algorithm) since these certainly won't be the same across languages.

I am interested to see where this is going as well. In the meantime, I submitted the code again using node.js in a separate PR (#752), so that there is a JS implementation which is more similar to the already existing implementations.

ntindle:
[lang: javascript]

Amaras:
Hi @0xJonas, thank you for your PR.

> On the other hand I understand that it is a good idea to keep the implementations consistent across languages, including possible testing code. Using external libraries in general might also be an issue (if they are not related to the algorithm) since these certainly won't be the same across languages.

Since 2020, we have decided we need to work on an output validation system (if you have not followed closely), so I am not sure if `p5.js` is a good idea for this indeed.
I'll let you the judge of if we need to close this PR :wink:

0xJonas:
Yeah, p5.js is definitely not suited well for automatic output validation. I'll close this PR.

PR #751, opened 2020-09-05
Update monte_carlo.py

DougBR:
([0,radius], [0, radius]) is in the upper RIGHT quadrant, as mentioned in the explanations for Monte Carlo.


leios (approved):
This is correct. Thanks for the update!

Issue #691, opened 2020-05-06
Compilation instructions for languages that require it


leios:
# Feature Request

## Description

It would be a good idea to have compilation instructions available for languages that require them. For example, C / C++ could have a makefile and rust could have a cargo.toml file available. I had some trouble getting the #686 working, and this might help.

There are a large number of existing implementations that could benefit from this.

## For Algorithm Archive Developers

- [ ] This feature can be added to the [Master Overview](https://github.com/algorithm-archivists/algorithm-archive/projects/11) (if it cannot be, explain why in a comment below -- lack of technical expertise, not relevant to the scope of this project, too ambitious)
- [ ] There is a timeline for when this feature can be implemented
- [ ] The feature has been added to the [Master Overview](https://github.com/algorithm-archivists/algorithm-archive/projects/11)
- [ ] The feature has been implemented (Please link the PR)

leios:
myusernamefrank from twitch chat said:

> sometimes other books or tutorials have a general build environment for all examples, maybe might be a good idea if it is one big project for each language which can be just compiled and generates all binaries?

jiegillet:
I'm with myusernamefrank, one example should do it, something trivial in the code itself (hello world) but non-trivial in the structure ( import a few libraries...).

leios:
So to be clear: You guys would prefer a single chapter on compilation where we explain the process for each language used in the AAA?

I think this is a good idea and would save some headache in the future. I don't know if we also need local build configuration as well?

jonay2000:
Instructions globally are fine, but we need some part of it locally as well. For example, some of the code in the archive uses libraries (such as for Rust: generating random numbers). It's useful to have some instructions on which libraries to load locally. In the case of rust, this is a Cargo.toml file (and a small project structure). In the case of c, maybe a makefile? For Javascript maybe a package.json etc.

leios:
Including the Cargo.toml file, Julia files should probably all have a Project / Manifest available so users can load everything with just `julia --project`

ShadowMitia:
Maybe for C/C++ we could have some script files? A combination of "build.sh" and "build.bat", to make it easily cross platform, and not have to depend on too many third parties? Makefiles don't work out of the box on Windows, and things like CMake are a pain to setup and manage.

This could apply to other languages as well, but in C/C++ it would be a good practice to encourage enabling more warnings, to detect more bugs.

For example in C++ I use all of these:

[...]

PR #755, opened 2020-09-30
Rust version of IFS

Liikt:
I created the IFS in rust. It has to pull in the rand crate though. (Why doesn't rust have a rand lib in the standard lib anyway?)


leios:
firstly, code looks good and runs fine. I am happy to merge it; however, I still have some lingering questions about running code like this from the AAA (#691). Should we maybe include the Cargo.toml with this? I know it's different than what we did with the other implementations for Rust, but I feel it would help!

ntindle:
[lang: rust]

contents/IFS/code/rust/IFS.rs
```diff
@@ -0,0 +1,36 @@
...
+        Point{x: 1., y: 0.},
+    ];
+
+    let mut out = String::new();
+
+    for point in chaos_game(10_000, shapes) {
+        out += format!("{}\t{}\n", point.x, point.y).as_str();
+    }
+
+    std::fs::write("./out.dat", out).unwrap();
```
stormofice, line 35:
```suggestion
    std::fs::write("./sierpinski.dat", out).unwrap();
```

stormofice (commented):
Thanks for the PR!

Overall this code looks good to me and seems to be very similar in concept to #896, which makes me want to approve this. 
However I don't feel completely confident in reviewing Rust code, so a second opinion would be good.

I did include a small suggestion to standardize the output file's name, as seen in #961.

contents/IFS/code/rust/IFS.rs
```diff
@@ -0,0 +1,36 @@
...
+        p.x = 0.5 * (p.x + tmp.x);
+        p.y = 0.5 * (p.y + tmp.y);
+        old_point
+    }).collect()
+}
+
+fn main() {
+    let shapes = vec![
+        Point{x: 0., y: 0.},
+        Point{x: 0.5, y: (0.75 as f64).sqrt()},
```
ShadowMitia, line 25: To avoid clippy warning on this line, I suggest was change to suffix notation.
```suggestion
        Point{x: 0.5, y: 0.75_f64.sqrt()},
```

contents/IFS/code/rust/IFS.rs
```diff
@@ -0,0 +1,36 @@
+use rand::prelude::*;
```
ShadowMitia, line 1: Depending on the version of `rand` you are using, you can remove the `::prelude`
```suggestion
use rand::*;
```

ShadowMitia (changes requested):
Couple of changes are needed to make this compile and remove some tiny warnings.
Otherwise I think it's good!

Amaras:
Hello @Liikt!
Since this has an external dependency (namely the `rand` crate), we now require a `cargo.toml` file for compilation (for automated compilation).
Could you please add the file in if you have some time?

Liikt:
I have commited the suggested changes and a cargo toml

ShadowMitia (approved):
Looks good to me, thanks!

Issue #659, opened 2019-12-20
Edge case fail in all Huffman Encoding implementations.


Trashtalk217:
Whenever any (or most) of the Huffman Encoding implementations receive a string with only one character multiple times (think `'aaaaaaaaaaaaa'`, or `'bbbb'`). The functions fail and give either errors, or weird empty output. Here are some [examples](https://imgur.com/a/4MuqlXr ) of Python, Java and Lua respectively. I assume this has to be fixed because it nowhere states in the chapter that it shouldn't work on one-character-strings.

c252:
These are the ones so far that we know are broken

- [ ] Julia
- [ ] Python
- [ ] Java
- [ ] Javascript
- [ ] Rust
- [ ] C++

berquist:
Here is a first stab at fixing the Python implementation:
```diff
diff --git a/contents/huffman_encoding/code/python/huffman.py b/contents/huffman_encoding/code/python/huffman.py
index 1e1cc54..75700cd 100644
--- a/contents/huffman_encoding/code/python/huffman.py
+++ b/contents/huffman_encoding/code/python/huffman.py
@@ -11,6 +11,9 @@ def build_huffman_tree(message):
     frequencies = Counter(message)
     trees = frequencies.most_common()
 
+    if len(trees) == 1:
+        return [trees[0][0]]
+
     # while there is more than one tree
     while len(trees) > 1:
 
@@ -37,6 +40,9 @@ def build_codebook(tree, code=''):
 
     codebook = []
 
+    if len(tree) == 1:
+        return [(tree[0], '0')]
+
     # split the tree
     left_tree, right_tree = tree
 
```
I figure it's ok to not handle the case where the input is empty. Or should that be handled?

leios:
Right, this should catch it. I'm wondering it. I'll fix the ones I can soon!

Kroppeb:
https://stackoverflow.com/questions/22429854/huffman-code-for-a-single-character

Kroppeb:
I don't like the solution of using 1 bit to encode the solution. Mostly cause it makes the Kotlin implementation more complex than I feel it needs to be.

Colocasian:
[...]

PR #756, opened 2020-10-01
fix both error of issue 739

### LINKED ISSUE #739: The Python implementation of Huffman encoding fails on 1-character-long strings and empty strings
# Bug Report

## Description

There are two bugs here, both in [huffman.py](https://github.com/algorithm-archivists/algorithm-archive/blob/master/contents/huffman_encoding/code/python/huffman.py). First, if you call `build_huffman_tree` with an empty string, it throws an `IndexError` on line 32.

Secondly, if you call `build_huffman_tree` on a string 1 character long (Hypothesis suggested `'0'`), and then call `build_codebook` with the resulting tree, it throws `ValueError: not enough values to unpack (expected 2, got 1)` on line 41. The tree that `build_huffman_tree` builds in this case is actually just a string that's one character long, both in this case and for #659, which I'm pretty sure is unintended behavior.

## Environment

This was tested in Python 3.8, but is likely reproducible on every Python version the script supports.

## Additional context

These bugs were found using the Python property-based testing library Hypothesis. This issue may be a duplicate of, or related to, #659. Apologies if I trimmed too many sections from the template.

## For Algorithm Archive Developers

- [x] The bug can be reproduced
- [x] The bug can be fixed (if not, please explain why not in a comment below)
- [x] There is a timeline to fix the bug
- [ ] The bug has been fixed (Please link the PR)

Liikt:
This should fix #739

### DIFF
--- a/contents/huffman_encoding/code/python/huffman.py
+++ b/contents/huffman_encoding/code/python/huffman.py
@@ -29,11 +29,17 @@ def build_huffman_tree(message):
         # insert the new tree there
         trees.insert(index, (new_tree, new_weight))
 
-    huffman_tree = trees[0][0]
+    # Return an empty list if the message was empty, else the tree
+    huffman_tree = [] if not trees else trees[0][0]
     return huffman_tree
 
 # constructs the mapping with recursion
 def build_codebook(tree, code=''):
+    # Check wether our tree contains more than 1 element or not
+    if not tree:
+        return []
+    elif type(tree) != list:
+        return [(tree, '0')]
 
     codebook = []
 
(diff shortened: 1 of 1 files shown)


thecnoNSMB:
I just ran this past my property test and it passes both cases without finding any further failure modes. It also fixes #659.

contents/huffman_encoding/code/python/huffman.py
```diff
@@ -29,11 +29,17 @@ def build_huffman_tree(message):
...
         trees.insert(index, (new_tree, new_weight))
 
-    huffman_tree = trees[0][0]
+    # Return an empty list if the message was empty, else the tree
+    huffman_tree = [] if not trees else trees[0][0]
     return huffman_tree
 
 # constructs the mapping with recursion
 def build_codebook(tree, code=''):
+    # Check wether our tree contains more than 1 element or not
```
leios, line 38:
```suggestion 
    # Check whether our tree contains more than 1 element or not
```

leios (approved):

Issue #765, opened 2020-10-13
Redundant Validity Checks in Flood Fill Julia and C Implementations


lazyprop:
In the [Julia](https://github.com/algorithm-archivists/algorithm-archive/blob/master/contents/flood_fill/code/julia/flood_fill.jl) and [C](https://github.com/algorithm-archivists/algorithm-archive/blob/master/contents/flood_fill/code/c/flood_fill.c) implementations, the `color` function checks if the point is inside bounds and if `old_val == new_val`.
These conditions are already checked in the `find_neighbours` function. So the points which are passed to `color` will always satisfy these conditions.The conditions are totally unnecessary. 
`color` can simply be:
```py
def color(canvas, p, new_val):
    canvas[p] = new_val
```

lazyprop:
#752, and #743  have the same problem.

PR #743, opened 2020-08-05
Added flood fill in Coconut

Amaras:
Here is the new Coconut code for the flood fill chapter.
I don't see how to make the code more functional, sadly, as it's a really iterative method.

There might be a merge conflict with the C code for now, but in the meantime, this PR is waiting for review.

PR #766, opened 2020-10-13
Remove color function from flood fill because it is redundant

leios:
Direct fix for #765 

Because the color function is essentially a bounds check before coloring and the bounds check is done upon querying neighbors, it's not necessary to check on color as well.

I have removed the function and the text associated with it.

This might also change code in #752 and #743


lazyprop:
This also adds the unnecessary `a.out` binary.

berquist (changes requested):
I've confirmed that this works and the book output is correct, but the compiled `a.out` binary needs to be removed.

leios:
Woops. Removed the binary file. Thanks for the catch!

berquist (approved):

PR #767, opened 2020-10-14
Added Comments

creativeyashi:


c252:
Hi thanks for the contribution. We have not yet come to a consensus on comments in the AAA but we are leaning on the side of not having them unless the code is doing something very language specific. Because the idea is the chapter itself will give a comprehensive understanding of how the algorithm works.

PR #768, opened 2020-10-14
Min max stack example

Iinguistics:
Added a stack class example in js & added imported the code into the md file.


leios:
I do intend to eventually add code to this chapter, so this PR is a good idea in principle; however, I don't know enough JS to say whether this is a good, fundamental implementation of a stack or queue in that language.

right now, my understanding is that this code goes a bit *futher* to allow for outputing the maximum or minimum element in the stack, which is useful, but not what I was thinking of for this chapter.

ntindle:
[lang: javascript]

PR #769, opened 2020-10-15
added code for algorith to check for prime number in c++

AnantGulati:
added code to check if a given number is prime with a run time of O(sqrt(n))


leios:
Hey, we appreciate the submission, but the algorithm archive is first-and-foremost a book. Code without a chapter cannot be accepted at this time.

Thanks!

PR #771, opened 2020-10-16
adding some code for stack and queue in python

nandiya:
code example for queue and stack implementation in python

[x] following rule
[x] Only implement algorithms that are already in the AAA
[x] only one language and one algorithm
[x] already test my code

### DIFF
--- a/contents/stacks_and_queues/code/python/stack_queue.py
+++ b/contents/stacks_and_queues/code/python/stack_queue.py
@@ -3,88 +3,86 @@ class Node:
         self.value = val
         self.next = None
 
-class Stack():
+
+class Stack:
     def __init__(self):
         self.stack_list = None
-    
+
     def push(self, val):
-        if self.stack_list == None:
-            self.stack_list=  Node(val)
-        else :
+        if self.stack_list is None:
+            self.stack_list = Node(val)
+        else:
             temp = Node(val)
             temp.next = self.stack_list
             self.stack_list = temp
-            
+
     def pop(self):
-        if self.stack_list == None:
-            print ("Stack is Empty")
-            return None
+        if self.stack_list is None:
+            raise ValueError("Stack is empty")
         else:
             pop_val = self.stack_list.value
             self.stack_list = self.stack_list.next
-            print("you just pop ", pop_val)
             return pop_val
-    
+
     def get_top(self):
         return self.stack_list.value
-    
+
     def print(self):
-        print ("Current stack list : ")
+        print("Current stack list : ")
         temp = self.stack_list
-        while(temp.next != None):
+        while temp.next:
             print(temp.value)
             temp = temp.next
         print(temp.value)
-        print("========================")
 
-class Queue():
+
+class Queue:
     def __init__(self):
-        self.queue_list=None
-        
-    def enqueue(self,val):
+        self.queue_list = None
+
+    def enqueue(self, val):
         temp = Node(val)
-        if self.queue_list == None:
+        if self.queue_list is None:
             self.queue_list = temp
         else:
             curr = self.queue_list
-            while(curr.next != None):
+            while curr.next != None:
                 curr = curr.next
             curr.next = temp
-            
+
     def dequeue(self):
-        if self.queue_list == None:
-            return None
+        if self.queue_list is None:
+            raise ValueError("Queue is empty")
         else:
             temp = self.queue_list
             dequeue_val = self.queue_list.value
-            self.queue_list = temp.next         
+            self.queue_list = temp.next
             return dequeue_val
-        
+
     def get_front(self):
         curr = self.queue_list
-        while(curr.next != None):
+        while curr.next:
             curr = curr.next
         return curr.value
-    
+
     def print(self):
-        print("Current queue list = ")
         temp = self.queue_list
-        while(temp.next != None):
+        while temp.next:
             print(temp.value)
             temp = temp.next
-        print (temp.value)
-        print("========================")
-    
+        print(temp.value)
+
 
 example_stack = Stack()
 example_stack.push(1)
 example_stack.push(2)
 example_stack.push(3)
 example_stack.print()
-example_stack.pop()
+print("you just pop ", example_stack.pop())
+
 example_stack.push(4)
 example_stack.print()
-print ("current top => ",example_stack.get_top())
+print("current top => ", example_stack.get_top())
 
 example_queue = Queue()
 example_queue.enqueue(1)
@@ -94,4 +92,4 @@ example_queue.print()
 example_queue.dequeue()
 example_queue.enqueue(4)
 example_queue.print()
-print ("current front => ", example_queue.get_front())
+print("current front => ", example_queue.get_front())
--- a/contents/stacks_and_queues/code/python/stack_queue.py
+++ b/contents/stacks_and_queues/code/python/stack_queue.py
@@ -1,5 +1,3 @@
-import numpy as np
-
 class Node:
     def __init__(self, val):
         self.value = val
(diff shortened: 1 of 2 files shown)


nandiya:
@Amaras please review my code and can it be used?

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+        self.value = val
+        self.next = None
+
+class Stack():
+    def __init__(self):
+        self.stack_list = None
+    
+    def push(self, val):
+        if self.stack_list == None:
+            self.stack_list=  Node(val)
```
Amaras, line 12: Slight formatting problem.
```suggestion
            self.stack_list = Node(val)
```

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+        else :
+            temp = Node(val)
+            temp.next = self.stack_list
+            self.stack_list = temp
```
Amaras, line 16: The way you push to the stack seems weird... You push to the beginning, which is not usual. Why don't you simply use a `list` object instead of your linked list?

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+            
+    def pop(self):
+        if self.stack_list == None:
+            print ("Stack is Empty")
```
Amaras, line 20: Why do you need to `print` things, actually? You can do that outside the class, because now it will pollute `stdout` every time you use it...

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+            return None
```
Amaras, line 21: You should probably just raise a `ValueError` exception instead of returning `None`, as the `Node` value can actually be `None`, and it doesn't help you programmatically differentiate between a `None` value, and an empty `Stack`

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            pop_val = self.stack_list.value
+            self.stack_list = self.stack_list.next
+            print("you just pop ", pop_val)
+            return pop_val
```
Amaras, line 26: Okay, so that is consistent with the way you push, I just have problems with the `print` thingy.

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+    
+    def get_top(self):
+        return self.stack_list.value
+    
+    def print(self):
+        print ("Current stack list : ")
+        temp = self.stack_list
+        while(temp.next != None):
```
Amaras, line 34: There are several weird things in this line: the redundant parentheses in the `while` loop, the `temp.next != None` which should be `temp.next is not None`, or even simply `temp.next `.

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+    def __init__(self):
+        self.queue_list=None
+        
+    def enqueue(self,val):
+        temp = Node(val)
+        if self.queue_list == None:
+            self.queue_list = temp
+        else:
+            curr = self.queue_list
+            while(curr.next != None):
```
Amaras, line 50: Okay, so the need to have a `while` loop means your underlying data structure is not useful for your usage.
A queue should have fast insertion on one end and fast popping on the other. While I see that your dequeueing is correct, having slow insertion defeats the purpose of he queue...

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+            return None
+        else:
+            temp = self.queue_list
+            dequeue_val = self.queue_list.value
+            self.queue_list = temp.next         
+            return dequeue_val
+        
+    def get_front(self):
+        curr = self.queue_list
+        while(curr.next != None):
```
Amaras, line 65: `get_front` should probably be a constant time operation, not a linear time one. I'm sorry, but your Queue implementation is not correct... You need to rework it.

Amaras (changes requested):
So, I have only done a code review. I don't even discuss the opportunity of adding code example on this chapter...
Don't worry if I'm too harsh, I am usually on code, be it from myself or others.

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
+class Node:
+    def __init__(self, val):
+        self.value = val
+        self.next = None
```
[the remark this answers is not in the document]
nandiya: ahh, I thought this folder for basic implementation in python and we can contribute in any language, so that's why i tried to make one.
well in python itself already has list() which we can used simply, so yeah back to my first reason..

contents/stacks_and_queues/code/python/stack_queue.py
```diff
@@ -0,0 +1,97 @@
...
+
+class Stack():
+    def __init__(self):
+        self.stack_list = None
+    
+    def push(self, val):
+        if self.stack_list == None:
```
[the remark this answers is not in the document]
nandiya: hmm.. I'm still quite new with python itself.. is there any different if i use "is" and "==" ?? is there any different in compile or maybe something else?

contents/stacks_and_queues/code/python/stack_queue.py
nandiya: ah yeah, forgot to use black XD, my mistake, ty

contents/stacks_and_queues/code/python/stack_queue.py
nandiya: ahh this is back to the first reason.. i'm just implementing the basic way with python, and yeah I thought u're one of the reviewer, bcs I saw your name merged it, sorry abt that

contents/stacks_and_queues/code/python/stack_queue.py
nandiya: done

contents/stacks_and_queues/code/python/stack_queue.py
nandiya: done

contents/stacks_and_queues/code/python/stack_queue.py
nandiya: done

contents/stacks_and_queues/code/python/stack_queue.py
nandiya: done

contents/stacks_and_queues/code/python/stack_queue.py
nandiya: done

contents/stacks_and_queues/code/python/stack_queue.py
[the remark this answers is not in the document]
Amaras: The problem is that the behaviour of `==` depends on how `__eq__` is implemented for the class you consider (you could possibly have all your `Node` addresses be equal to `None`), while `is` directly compare memory addresses of the two objects, making it more efficient, and ultimately more secure to compare identity rather than equality for None

ntindle:
[lang: python]

leios:
What is the state of this PR? I think there are a few lingering changes to be made, but after that, we now have code for stacks and queues, so this can probably be merged in

PR #774, opened 2020-10-22
Create nCr.py

tsaideepak7:
Fermits little theorem to calculate nCr value


leios:
Hello! Thanks for the submission. We currently do not allow code without chapter content associated with it. This project is first-and-foremost a book and should be thought of as such. Code without text is closer to what you might see on Rosetta Code or some other project.

Let me know if you are interested in writing a chapter on this topic and we can talk!

PR #775, opened 2020-10-22
feat(ts): stack and queue implementation

Progyan1997:


leios:
thanks for the implementation; however, we do not yet have a chapter *with code* for stacks and queues. This will be updated soon, so I'll keep this PR in mind as we write the text for that section #771 has a similar issue, so hopefully we can work everything out!

contents/stacks_and_queues/code/typescript/queue.ts
```diff
@@ -0,0 +1,52 @@
...
+   * `size` return size or length of the queue
+   */
+  size(): number;
+  /**
+   * `front` returns first element of the queue
+   */
+  front(): T;
+}
+
+class Queue<T> implements IQueue<T> {
```
ntindle, line 20: @leios do we have standards for TypeHinting in the interface names?

ntindle (commented):
Am typescript dev, this LGTM

contents/stacks_and_queues/code/typescript/queue.ts
leios: I don't think so. Maybe we should add something to the code style guide?

leios (approved):
If it looks good to @ntindle, then I am happy to approve as well!

leios:
I will go ahead and merge this now and then put the infrastructure in another PR to solve issue #838

PR #776, opened 2020-10-22
added factorial in haskell

MagnerSA:


leios:
Hello, thanks for the submission. I am not really sure what this is for, but we currently do not accept code without a corresponding chapter. If you want to write a chapter on this, then we might think about it.

Thanks!

Issue #665, opened 2020-01-15
pdf export / print


dadosch:
# Bug Report
When I try to print a webpage using firefox or chrome, only the first page is printed. Maybe this is a gitbooks issue, so it would be handy to have pdfs of each article which one could download.

### Steps to Reproduce

1. Go to 'https://www.algorithm-archive.org/contents/gaussian_elimination/gaussian_elimination.html'
2. Show print preview of firefox/chrome

### Expected behavior

the full webpage should be shown or there should be a pdf available to download.

## For Algorithm Archive Developers

- [x] The bug can be reproduced
- [x] The bug can be fixed (if not, please explain why not in a comment below)
- [x] There is a timeline to fix the bug
- [x] The bug has been fixed (Please link the PR)

leios:
Right, we do not have pdf exporting available right now and we need to think about how we might do this. For now, we cannot really tackle this issue until we switch to something other than gitbook (or find a better way to make gitbook work for us)

leios:
This has been fixed in #777 (I just forgot to close the issue). Thanks for bringing it up!

PR #777, opened 2020-10-22
Added CSS for printing pages

0xJonas:
Adding some CSS rules to enable printing of pages without the side bar getting in the way and without the content stopping after the first page.

The problems with printing/print preview where mentioned in #665.

### DIFF
--- a/styles/website.css
+++ b/styles/website.css
@@ -28,3 +28,25 @@ img.center {
 #book-search-input input, #book-search-input input:focus, #book-search-input input:hover{
   color: #222;
 }
+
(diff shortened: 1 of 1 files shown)


leios:
firstly, thanks a bunch for this! It seems to work fine on my end. As a note: I am wondering if there is a way to create a custom printing script that will allow the printed pages to be formatted a bit nicer and potentially display video content as a series of images (but only in printing mode)

I guess having a printable mode of each page is actually the solution there.

leios (approved):

PR #780, opened 2020-10-26
Add Computus in Nim

berquist:


leios (approved):
It runs and provides the correct results. Also, this one is just math, so I don't think we need to worry too much about stylistic interpretations of the code.

PR #782, opened 2020-10-28
Make Verlet integration in Nim more idiomatic

berquist:
- `func` signifies that the procedures have no side effects: they are proper functions.
- Prefer implicit return of the last expression over explicit return.
- `when isMainModule` behaves the same way as `if __name__ == "__main__"` in Python.


leios (approved):
following #783 , this can also be merged. Thanks a bunch!

PR #784, opened 2020-10-28
Make Euclid's algorithm in Nim more idiomatic

berquist:
- `func` signifies that the procedures have no side effects: they are proper functions.
- Use the implicit `result` return variable instead of explicit returns.
- `when isMainModule` behaves the same way as `if __name__ == "__main__"` in Python.


leios:
@c252 Any way you can review the nim prs?

On my end, I am happy to merge them!

c252:
Yeah this looks good to me.

leios (approved):
@c252 says it looks good, so I'm on board!

PR #789, opened 2020-12-21
add Go implementation of IFS

radu-munteanu:


leios (commented):
I think the code looks clean enough, but I don't know Go. I will say that this builds and has the correct line numbers

ntindle:
[lang: go]

Amaras:
Sorry for taking so long and not noticing this, but we can't accept PRs coming from forks' `master` branch (you're supposed to track master on your fork)
If you still want to submit to the AAA , please open a PR from a separate branch

PR #790, opened 2020-12-25
Migrating to Github CI

leios:
This should allow us to migrate to Github CI


leios:
This has been tested on a separate branch and does seem to deploy correctly. We'll see if that's actually true ^^

berquist:
It seems to work locally with `act`, but requires a custom Docker image (see https://github.com/nektos/act/issues/107)
```dockerfile
FROM ubuntu:20.04

RUN apt-get update \
    && apt-get upgrade -y \
    && DEBIAN_FRONTEND=noninteractive apt-get install -y \
    build-essential \
    curl \
    git \
    nodejs \
    npm \
    && rm -rf /var/lib/apt/lists/*
```
followed by `sudo act -P ubuntu-latest=ubuntu-builder`

.github/workflows/github_actions.yml
```diff
@@ -0,0 +1,47 @@
...
+          book_check_file="_book/index.html"
+          if [[ ! -f  "${book_check_file}" ]]; then
+            echo "${book_check_file} not found"
+            exit 1
+          fi
+
+      - name: Finishing build and pushing to algorithm-archivists.github.io
+        if: github.event.pull_request.merged == true
+        run: |
+          bash tools/deploy/update_site.bash $GH_TOKEN
```
berquist, line 44: I would like to replace this mess with lifting the CNAME plus git commit bits into the `run` of this task, and just hardcode a bunch of variables, then have the push be another task that uses https://github.com/ad-m/github-push-action.

leios: I did actually try this, but I was not able to get it *quite* working. I would be happy for the help, though!

berquist: Ok, but it will have to wait until this weekend. Can you check the Travis page and see how many free credits are left?

leios: I don't think there's any rush. I'll check Travis later, but I believe we have enough credits because there has not been a huge amount of activity.

leios: I also realized recently that we do not need a separate repo for gh-pages anymore and could just use a gh-pages branch. This would probably make things a bit more streamlined.

ntindle:
Is this still intended to be merged? If so, I can take a look.

leios:
Yeah. I would like to merge it after a review. I don't trust myself to do CI right ^^

ntindle:
Would you mind if I made a PR with a different pipeline setup?

leios:
I am closing this for now because we have merged #821 instead

PR #793, opened 2021-01-07
Bogo sort implementation in Elvish

andrewandreii:
I am not sure if I can use external commands, but I used the `shuf` command (for shuffling) which should be available in most Linux distributions.

### DIFF
--- a/book.json
+++ b/book.json
@@ -216,6 +216,10 @@
         {
           "lang": "coco",
           "name": "Coconut"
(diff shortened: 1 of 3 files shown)


ntindle:
[lang: elvish]

Amaras:
Hi, sorry, but we don't accept sorting algorithm PRs right now, so I'm closing this PR.

PR #794, opened 2021-01-07
Stable marriage problem implementation in Elvish

andrewandreii:
This is an algorithm I came up with so it needs to be checked.

### DIFF
--- a/contents/stable_marriage_problem/code/elvish/stable_marriage.elv
+++ b/contents/stable_marriage_problem/code/elvish/stable_marriage.elv
@@ -25,15 +25,15 @@ fn return-same [thing]{ put $thing }
 
 # pref format for females: [&name=[&name=number ..] ..] for O(1) lookup
 # pref format for males: [&name=[name1 name2 ..] ..] because it's used like a stack
-fn gen-pref [names1 names2 &switch-idx=0]{
+fn gen-pref [names1 names2 &switch-idx=$false]{
   map = [&]
 
-  fn = [thing]{ return-same $thing }
-  if (== $switch-idx 1) {
-    fn = [list]{ switch-index $list }
+  each-fn = [name]{ map[$name] = (shuffle $names2) }
+  if $switch-idx {
+    each-fn = [name]{ map[$name] = (switch-index (shuffle $names2)) }
   }
 
-  each [name]{ map[$name] = ($fn (shuffle $names2)) } $names1
+  each $each-fn $names1
 
   put $map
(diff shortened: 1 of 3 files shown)


contents/stable_marriage_problem/code/elvish/stable_marriage.elv
```diff
@@ -0,0 +1,140 @@
...
+    exit = 1
+    for female $fem-names {
+      # choose men who have $female as their first choice
+      proposals = [(take-if [elem]{ $proposal-cond $elem $female } $men-names)]
+
+      # if we found proposals
+      if (not (empty $proposals)) {
+        exit = 0
+      } else {
+        continue
```
Amaras, line 100: Okay, that one took me a while to understand, but it makes sense. You should say that if we didn't find any proposal, the algorithm has finished.

contents/stable_marriage_problem/code/elvish/stable_marriage.elv
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+      # eliminate those who can't marry because she's already married
+      if (==s (kind-of $fem-pref[$female]) string) {
+        for elem $proposals {
+          men-pref[$elem] = $men-pref[$elem][1..]
+        }
```
Amaras, line 107: I don't understand that loop, could you explain it better? No one is married until the end of the algorithm. So... what is that doing?

Amaras (commented):
I don't really understand the language, but the **Gale-Shapley** (and not Gale-Shapely) algorithm looks correct, though there are a few weird points I'd like you to clarify. I haven't run it though, so testing is not yet done.

Thanks for your contribution!

contents/stable_marriage_problem/code/elvish/stable_marriage.elv
andrewandreii: I have changed the condition to be `is-empty $proposals` (also `empty` renamed to `is-empty`) and modified the comment with more explanation of what the algorithm is doing.

contents/stable_marriage_problem/code/elvish/stable_marriage.elv
andrewandreii: When $female has a pair, $fem-pref[$female] is equal to the male's name, i.e., it's no longer a map but a string. So we need to remove $female from all the males who are in $proposals.

ntindle:
[lang: elvish]

PR #796, opened 2021-01-25
Update to the convolution chapter(s)

leios:
This is a draft for the upcoming convolution chapters. Let me know if there's anything that's unclear to update before merging!


leios:
Some notes on Discord:
1. We should probably use a different test. No random numbers. Maybe sawtooth with square?
2. It is a bit weird that the simple boundaries will grab the first part of the full convolution instead of the center of the full convolution. I did this because I thought it was not particularly intuitive to have the filter start while overlapping the signal. I could rewrite the code or add (something like) the following:
> What is j doing in this case? Well, it is iterating through elements on the input array that will be multiplied by the signal. Here, we chose for it to count up to i, but depending on the use-case, you might want to center the convolutional kernel. To do this, we would count from i-filter_size/2 to i+filter_size/2

I am just making sure these are written down for later

leios:
Ok, I am happy with this chapter. I'll read it over one more time tomorrow before merging

PR #798, opened 2021-01-25
julia bubble sort suggestions

Luapulu:
- Restrict type less. `bubble_sort` should be able to sort any `AbstractVector`.
- Generalised to arbitrary indexing offset so that any offset vector can be sorted.
- Use `_` as a placeholder variable.
- Add `@inbounds` for better performance.
- Have mutating version return its result.
- Add `bubble_sort`, a non-mutating version.
- Use `let` keyword rather than `main` function.
- Replace `[1., 3, 2, 4, 5, 10, 50, 7, 1.5, 0.3]` (`Vector{Any}`) with `rand(1:100, 9)` (`Vector{Int}`).

### DIFF
--- a/contents/bubble_sort/code/julia/bubble.jl
+++ b/contents/bubble_sort/code/julia/bubble.jl
@@ -1,19 +1,24 @@
-function bubble_sort!(a::Vector{Float64})
-    n = length(a)
-    for i = 1:n
-        for j = 1:n-1
-            if(a[j] < a[j+1])
+using Base: copymutable
+
+function bubble_sort!(a::AbstractVector)
+    ax     = axes(a, 1)
+    n      = length(ax)
+    lo, hi = first(ax), last(ax)
+
+    for _ in lo:hi
+        @inbounds for j in lo:hi-1
+            if a[j] < a[j+1]
                 a[j], a[j+1] = a[j+1], a[j]
             end
         end
     end
+
+    return a
 end
 
+bubble_sort(a) =  bubble_sort!(copymutable(a))
 
-function main()
-    a = [1., 3, 2, 4, 5, 10, 50, 7, 1.5, 0.3]
+let a = rand(1:100, 9)
     bubble_sort!(a)
-    println(a)
+    display(a)
 end
-
-main()
--- a/contents/bubble_sort/code/julia/bubble.jl
+++ b/contents/bubble_sort/code/julia/bubble.jl
@@ -1,12 +1,10 @@
 using Base: copymutable
 
 function bubble_sort!(a::AbstractVector)
-    ax     = axes(a, 1)
-    n      = length(ax)
-    lo, hi = first(ax), last(ax)
+    n = length(a)
 
-    for _ in lo:hi
-        @inbounds for j in lo:hi-1
+    for _ in 1:n
+        @inbounds for j in 1:n-1
             if a[j] < a[j+1]
                 a[j], a[j+1] = a[j+1], a[j]
             end
@@ -18,7 +16,7 @@ end
 
 bubble_sort(a) =  bubble_sort!(copymutable(a))
 
-let a = rand(1:100, 9)
+let a = [1, 9, 2, 8, 3, 7, 4, 6, 5]
     bubble_sort!(a)
     display(a)
 end


contents/bubble_sort/code/julia/bubble.jl
```diff
@@ -1,19 +1,24 @@
-function bubble_sort!(a::Vector{Float64})
-    n = length(a)
-    for i = 1:n
-        for j = 1:n-1
-            if(a[j] < a[j+1])
+using Base: copymutable
+
+function bubble_sort!(a::AbstractVector)
```
leios, line 3: Allowing the function to work on arbitrary numbers is a good idea, but I would probably change it to something like:

```
function bubble_sort!(a::Vector{T}) where {T <: Number}
```

contents/bubble_sort/code/julia/bubble.jl
```diff
@@ -1,19 +1,24 @@
...
+    ax     = axes(a, 1)
+    n      = length(ax)
+    lo, hi = first(ax), last(ax)
```
leios, line 6: I feel this is a bit more complicated than it needs to be. In fact, I am not sure what it is actually doing. You are reading in an abstract vector, which means it could be n-dimensional, but sorting only along the first dimension of that, right?

In this implementation, I suppose I would prefer to have the bubble sort function also select the axes to sort over?

contents/bubble_sort/code/julia/bubble.jl
```diff
@@ -1,19 +1,24 @@
...
+            if a[j] < a[j+1]
                 a[j], a[j+1] = a[j+1], a[j]
             end
         end
     end
+
+    return a
 end
 
+bubble_sort(a) =  bubble_sort!(copymutable(a))
```
leios, line 19: I agree that this is better Julia code to have both an in-place and out-of-place operation.

contents/bubble_sort/code/julia/bubble.jl
```diff
@@ -1,19 +1,24 @@
...
 
-function main()
-    a = [1., 3, 2, 4, 5, 10, 50, 7, 1.5, 0.3]
+let a = rand(1:100, 9)
```
leios, line 21: We were trying to keep the same test for all languages, so I was avoiding random numbers because I knew *someone* would want to do this in BrainFuck,

Also the `main()` function was for consistency as well. `let` is not a bad idea, though. It might be a good idea to use in the future.

contents/bubble_sort/code/julia/bubble.jl
```diff
@@ -1,19 +1,24 @@
...
     bubble_sort!(a)
-    println(a)
+    display(a)
```
leios, line 23: Is `display(...)` better here because you are using an abstract vector?

contents/bubble_sort/code/julia/bubble.jl
```diff
@@ -1,19 +1,24 @@
...
+using Base: copymutable
+
+function bubble_sort!(a::AbstractVector)
+    ax     = axes(a, 1)
+    n      = length(ax)
+    lo, hi = first(ax), last(ax)
+
+    for _ in lo:hi
+        @inbounds for j in lo:hi-1
+            if a[j] < a[j+1]
```
leios, line 10: I am not entirely sure this line will work for all abstract vector types? What if you have a vector of self-defined structs? Is the `<` operation always guaranteed to be available?

contents/bubble_sort/code/julia/bubble.jl
```diff
@@ -1,19 +1,24 @@
...
-        for j = 1:n-1
-            if(a[j] < a[j+1])
+using Base: copymutable
+
+function bubble_sort!(a::AbstractVector)
+    ax     = axes(a, 1)
+    n      = length(ax)
+    lo, hi = first(ax), last(ax)
+
+    for _ in lo:hi
```
leios, line 8: Won't the compiler already get rid of `i`? That sounds like something LLVM should do.

leios (commented):
Thanks a bunch for the PR!

We chose a simple type (`Float64`) without any intent to abstract it because it's easier to understand the algorithm without the abstractions, and in general this is the philosophy we use in the book. We are not trying to find the *best* implementations, but the *most understandable*

On the other hand, you bring up a few valid points:
1. Abstracting to numbers is not hard. Why not just do it? I am somewhat against abstracting it out to `AbstractVector`, though.
2. I use `main()` to make sure everything is in the right scope, but that is really C-style, maybe `let` is better? The only issue with that is that the code will run on `include(...)`, when I prefer having things run on function invocation in the REPL.
3. Should we be using `_`? As a brand new programmer, I would find it a bit confusing, but not so much that it would inhibit my understanding of the method.
4. Are we actually being consistent with our test across all languages? If not, `rand()` is probably better than setting numbers manually.
5. `@inbounds` was again one of those things I was 50:50 on. It does lead to better performance, but it's also a "Julia trick" that isn't necessary to understanding the method.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: Why limit to numbers? a sorting function sorts some sort of indexed ordered collection. Why do the contents matter? Limiting it to numbers or `Float64` actually adds complexity to the implementation (in julia at least) and makes it harder to understand than just `AbstractVector`.

Another possibility would be to have no type annotation whatsoever. Type annotations are only needed if you actually have different methods or if the type helps you understand the method. If you think `AbstractVector` is harder to understand than `Vector{<:Number}`, then may be no annotation might be even better.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: I did this to follow the base implementation of sorting functions, which allows for arbitrarily offset indexing. So, you could sort 0-indexed vectors for example. I'm happy to assume one based indexing and throw this out though if you like.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: We can do `[1, 9, 2, 8, 3, 7, 4, 6, 5]` instead for example. The main thing I wanted is to have a concretely typed vector, rather than `Vector{Any}`. I went with rand, because I saw that python had some random vectors but if for reproducibility we may want to avoid this.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: `display` just makes sure you see REPL style printing rather than the one-line compact printing.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: Another option here would be to use something like `@test sort([1, 2, 3]) == [3, 2, 1]`.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: No, `<` is not always defined. But the error message is pretty understandable. Try this:

```julia
julia> struct ABC; a::Int64; end

julia> ABC(5) < ABC(7)
ERROR: MethodError: no method matching isless(::ABC, ::ABC)
Closest candidates are:
  isless(::Missing, ::Any) at missing.jl:87
  isless(::Any, ::Missing) at missing.jl:88
Stacktrace:
 [1] <(::ABC, ::ABC) at ./operators.jl:277
 [2] top-level scope at REPL[75]:1
```

What are you suggesting?

contents/bubble_sort/code/julia/bubble.jl
Luapulu: So, `_` is actually a special write-only variable in julia to signal to the compiler that it will never be used. I'm not sure how smart the compiler is at this point and if it would get rid of `i`. In any case, `_` seems to be the most julian way of doing this kind of thing.

Try:
```julia
julia> _ = 5
5

julia> _
ERROR: all-underscore identifier used as rvalue
Stacktrace:
 [1] top-level scope at REPL[77]:1
```

contents/bubble_sort/code/julia/bubble.jl
leios: I am suggesting we limit this to `Number` because it makes the code way more understandable.

contents/bubble_sort/code/julia/bubble.jl
leios: I think assuming 1-based indexing is fine in this case

contents/bubble_sort/code/julia/bubble.jl
leios: I don't really care too much about _ or not. It's fine.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: Just removed this in the last commit.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: Btw, this also wouldn't work for vector types that are immutable. So, if the argument is that this method needs to be guaranteed to work, then we really would have to limit it to `Vector{<:Number}`. I just don't think this would be representative of actual julia code. People either have no type annotations or annotations that encompass the realm of reasonable inputs, (like a vector of `String` for example).

Luapulu:
So, I removed `rand` and removed the code to extend to arbitrary offsets.

If you don't like `AbstractVector`, I'd still prefer to just have no type annotations. I would avoid type restrictions unless you have different methods or there's no conceivable way to apply the function to other types. In this case it would totally conceivable to sort a vector of strings.

contents/bubble_sort/code/julia/bubble.jl
Luapulu: If you want more compact printing, let me know. We can go with `println`, too. That would also match other languages print statements so it might be easier to understand.

Luapulu:
What's the position on having comments in code? We could add a comment to explain what `@inbounds` does.

Luapulu:
What do you mean by "everything is in the right scope" ? We could also have to test code in global scope, couldn't we?

Anyways, this probably doesn't matter much. Whether you put the test code in local scope in a let-block, a test function or just in global scope probably makes no difference. If we're going for simplicity, we might as well go for global scope. Let me know what you think.

leios:
I would be happy to have a comment for `@inbounds`, again I recognize that that is a common thing people will see in Julia code.

As for the scoping: I am fine with `let`, I just prefer the code to run on function invocation, which is why I had the c-style `main()`. I am against the global scope for Julia because... Well, it's Julia. Scoping is an important thing to consider for performance.

The problem is that none of the languages are consistent in their tests, and this chapter is probably the first one people look at when going to the AAA. I might take it down for a bit and rework it

ntindle:
[lang: julia]

Amaras:
I guess I'm closing this PR, since it was depreciated when we removed all sorting code...

PR #799, opened 2021-01-26
removing sorting (and searching) chapters

leios:
Ok, hear me out:

These chapters are stubs. We all know it. I have been planning on cleaning them up; however, because everyone knows bubble and bogo sort, there have been an insane number of code contributions to these chapters. Most of these are inconsistent and need to be re-worked.

We need to do the following:
1. Revamp the complexity chapter
2. Revamp the sorting and searching chapter to explain why people care about these methods (mainly because they all do the same thing and help explain why complexity matters in practice)
3. Introduce an `is_sorted(...)` function that will work for all sorting methods
4. Add sorting and searching methods one-by-one

The reason I want to remove them (for now) is that I feel they are providing the wrong perspective of the algorithm archive. There are other stubby chapters (I am looking at you, Gift Wrapping), but they can hide behind anonymity a bit. Sorting and searching are some of the first methods people go to when looking at a book on algorithms, so if we do them... We need to do them right, in my opinion.

Now, people are opening chapter requests for other sorting methods using the bubble and bogo sort chapters as a base... Which is ultimately putting more work on me down the road when I am able to revamp the sorting and searching methods, as now I need to heavily revise those chapters as well.

So, thoughts?

PS: I have all the code already written backed up on another branch. It will all come back later this year.


leios:
As there has been no discussion on this, I will merge it tomorrow and close any lingering sorting PRs.

Again, the code is still saved, so it will come back!

PR #803, opened 2021-03-03
Fix github profile link in the affine transformations chapter

cloudgiant:
Fixed the github link at the end of the affine transformations chapter


leios (approved):
Nice catch, thanks!

PR #808, opened 2021-03-22
Comptus implementation in Scala

kenpower:


Amaras (changes requested):
Well, I don't really have anything to say against the implementation: it is a simple algorithm, so its implementation is pretty straightforward.
Though be careful to have the correct link for your code file in the chapter file.

kenpower:
> Well, I don't really have anything to say against the implementation: it is a simple algorithm, so its implementation is pretty straightforward.
> Though be careful to have the correct link for your code file in the chapter file.

Good spot. Thanks!

leios (approved):
Another great PR! Sorry I've taken so long to actually get to these!

PR #816, opened 2021-05-13
Fix typos in IFS/julia and IFS/c

stormofice:
There were some inconsistencies in the naming scheme of the output files.
Now all implementations output into `sierpinski.dat` instead of `out.dat`.


leios (approved):
Thanks for the catch!

PR #817, opened 2021-05-14
Add java implementation for Iterated Function Systems

stormofice:
This implementation uses a self created Point-Wrapper class.
It would be possible to use [Point2D.Double](https://docs.oracle.com/javase/8/docs/api/java/awt/geom/Point2D.Double.html), however this would hinder the readability of the code.

It is needed to clone the current point (line: 29), because if this is not done every entry of the array will reference the last point.

`main` throws an Exception due to the `FileWriter` creation, but a try-catch could also be used (at the cost of readability).

### DIFF
--- a/contents/IFS/IFS.md
+++ b/contents/IFS/IFS.md
@@ -140,6 +140,8 @@ Here, instead of tracking children of children, we track a single individual tha
 [import:18-29, lang:"c"](code/c/IFS.c)
 {% sample lang="coco" %}
 [import:4-16, lang:"coconut"](code/coconut/IFS.coco)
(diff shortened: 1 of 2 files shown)


Amaras (approved):
Apart from the Exception-throwing `main`, I have basically nothing to say against this code.

It's clean, though maybe implementing a clone method instead of directly copying arguments of the previous iteration might be more intuitive.

All in all, this looks like a great contribution, thanks for that :smiley:

leios (approved):
@Amaras likes it and I do to! Let's go ahead and merge it!

PR #818, opened 2021-06-15
Add Rust implementation for The Barnsley Fern

ShadowMitia:
I've based my code on #755 to make things easier and more coherent.

If this gets approved, I would suggest waiting until that PR gets resolved before applying mine. As it would have the same constraints about Cargo.toml.


Liikt:
Without testing if the code is correct (which i'll assume for now and test later) the code itself looks good to me 👍

leios:
2 important things to note here:
1. As per #691, we are thinking about adding build instructions for certain languages. This just means that rust should probably have a `Cargo.toml` to go with it so people can just run `cargo run`.
2. The code produces the following result:
rscheck

It *should* produce:
jlcheck

I noticed that the operators were a little off, but maybe there is a bigger issue here because julia is column-major while rust is row-major? I did not dig too much deeper into the code, but I can in a bit.

contents/barnsley/code/rust/barnsley.rs
```diff
@@ -0,0 +1,101 @@
...
+        vec![
+            Point3::new(0.20, -0.26, 0.0),
+            Point3::new(0.23, 0.22, 0.0),
+            Point3::new(0.0, 0.0, 1.0),
+        ],
+        vec![
+            Point3::new(-0.15, 0.25, 0.0),
+            Point3::new(0.26, 0.24, 0.44),
+            Point3::new(0.0, 0.0, 1.0),
+        ],
```
leios, line 84:
```suggestion
        vec![
            Point3::new(0.0, 0.0, 0.0),
            Point3::new(0.0, 0.16, 0.0),
            Point3::new(0.0, 0.0, 1.0),
        ],
        vec![
            Point3::new(0.85, 0.04, 0.0),
            Point3::new(-0.04, 0.85, 1.60),
            Point3::new(0.0, 0.0, 1.0),
        ],
        vec![
            Point3::new(0.20, -0.26, 0.0),
            Point3::new(0.23, 0.22, 1.6),
            Point3::new(0.0, 0.0, 1.0),
        ],
        vec![
            Point3::new(-0.15, 0.28, 0.0),
            Point3::new(0.26, 0.24, 0.44),
            Point3::new(0.0, 0.0, 1.0),
        ],
```
again, this doesn't solve the issue exactly, but at least these numbers look more correct to me.

ShadowMitia: I'll fix that straightaway, sorry about that!

ShadowMitia:
So I've added a `Cargo.toml` by creating a project with `cargo new rust` inside the code directory. Let me know what you think.

I've changed the constants, and I'll have a look at what's wrong with my generation.

ShadowMitia:
I've found the bug. It was a silly one :grimacing: :sweat_smile: 

You were right about the constants being off, but the issue was with the matrix multiplication.

Basically this was the issue:

```rust
p.x = rhs[0].x * self.x + rhs[0].y * self.y + rhs[0].z * self.z;
p.y = rhs[1].x * self.x + rhs[1].y * self.y + rhs[1].z * self.z;
p.z = rhs[2].x * self.x + rhs[2].y * self.y + rhs[2].z * self.z;
```

because `p.x` gets updated fine, but then `p.y` gets updated with a new `p.x` and old `p.y` and `p.z`. I guess this is why some things still sort of worked by coincidence.

I've moved the code to a method in `Point3` called `matrix_mul` but it could also have been solved by doing

```rust
let x = rhs[0].x * self.x + rhs[0].y * self.y + rhs[0].z * self.z;
let y = rhs[1].x * self.x + rhs[1].y * self.y + rhs[1].z * self.z;
let z = rhs[2].x * self.x + rhs[2].y * self.y + rhs[2].z * self.z;
p.x = x;
p.y = y;
p.z = z;
```

ntindle:
[lang: rust]

leios (approved):
The resulting image looks good now. Thanks!

PR #819, opened 2021-07-19
Convolutions of Images 2d Python implementation  

Ridham177:


Amaras (changes requested):
Alright, thanks for the PR!
I left a few comments, you may or may not take into account, and that's fine.
Overall, it seems like a faithful port of the Julia code, so no major problem there. I still hate the 4 live indices in the convolution, but it is not your fault, so the comment about the 4 loops is not that relevant.

However, there is a blocking point, which is that the `y /= np.linalg.norm` (line 84) won't run because of a type error.
You need to correct that before we can merge it in the AAA.

I think you have not run your code, which is why this bug is here and has slipped under your radar.

But once again, thanks for your interest in the AAA! :smiley:

contents/convolutions/code/python/2d_convolution.py
```diff
@@ -0,0 +1,129 @@
...
+    x = np.random.rand(100, 100)
+
+    # Gaussian signals
+    y = [[math.exp(-(((i-50)/100) ** 2 +
+                     ((j-50)/100) ** 2) / .01) for j in range(100)]
+         for i in range(100)]
+
+    # Normalization is not strictly necessary, but good practice
+    x = x / np.linalg.norm(x)
+    y = y / np.linalg.norm(y)
```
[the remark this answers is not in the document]
Ridham177: You are right that `y` is a list. When I run the code, it runs (and results in the correct output) and y is casted too ndarray.

Ridham177:
@Amaras I updated the if conditions to use suppress. Thanks for that suggestion. Code looks much cleaner. I also updated the if statements to not use backslashes, and use the from function for the cases where a gaussian kernel needs to be created. I ran the code and tested it by using a common input and checking the output of the Julia code and my code. The line `y = y / np.linalg.norm(y)` works (doesn't raise an error). I am unsure why that is the case. If it raises an error on your end, please let me know what I can do to replicate the issue.

leios:
At this stage, I don't fully understand why, but here is the output of your "full_linear.dat" file:
[](url)
pycheck

Here is the one from Julia:
juliacheck

leios:
I can dig into it in a bit, but this will need to be fixed before merging.

Thanks again for the submission!

Ridham177:
@leios Thank you for catching this. I would prefer to dig into it and fix it. Do the other outputs look correct? Also, what program are you using to visualize the output?

leios:
All image files are output so you can visualize them quickly with gnuplot (hence why we have an entire chapter on plotting for the data in the algorithm archive). Here, it should just be:

```
set view map
splot "out.dat" matrix with image
```

The simple linear option also looks incorrect, so far as I could see.

contents/convolutions/code/python/2d_convolution.py
```diff
@@ -0,0 +1,125 @@
...
+    def create_gaussian_signals(i, j):
+        return np.exp(-(((i-50)/100) ** 2 +
+                        ((j-50)/100) ** 2) / .01)
+    y = np.fromfunction(create_gaussian_signals, (100, 100))
+
+    # Normalization is not strictly necessary, but good practice
+    x /= np.linalg.norm(x)
+    y /= np.linalg.norm(y)
+
+    x = np.array([[5.0, 3.0], [4.0, 2.0]])
```
Amaras, line 85: You forgot to do something with the previous `x` and `y`

Amaras:
> The line `y = y / np.linalg.norm(y)` works (doesn't raise an error). I am unsure why that is the case.

It shouldn't raise any issue anymore, as you created a numpy array with `y = np.from_function(...)`
I think you ran it with the new version, which is why it doesn't raise an error, and not the previous one. If you checkout your previous version and try to run it, it should crash with a `TypeError`, with something like "Unimplemented operation '/' for types 'list', 'float'."

> At this stage, I don't fully understand why, but here is the output of your "full_linear.dat" file:
> 
> pycheck

I think that goes with my previous comment.

contents/convolutions/code/python/2d_convolution.py
Ridham177: Thanks.

Ridham177:
@leios The output from the julia code now matches the output of the python code. Thank you for reviewing this.

Ridham177:
@Amaras [This](https://drive.google.com/file/d/1WNlWTyCkhSdxVsmlWqy6gx3m2H7PSTDz/view) video where I try the buggy code using the old version shows that it doesn't raise error, probably because of python magic. With that being said, the old version shouldn't be used because its not clean and the types are not the same. Thanks for reviewing and for improving the code.

Amaras:
> @Amaras [This](https://drive.google.com/file/d/1WNlWTyCkhSdxVsmlWqy6gx3m2H7PSTDz/view) video where I try the buggy code using the old version shows that it doesn't raise error, probably because of python magic. With that being said, the old version shouldn't be used because its not clean and the types are not the same. Thanks for reviewing and for improving the code.

That probably should not happen, but hey! That's numpy magic, rather than Python magic.

contents/convolutions/code/python/2d_convolution.py
```diff
@@ -0,0 +1,122 @@
...
+    out = np.zeros((image_resolution, image_resolution))
+
+    for i in range(image_resolution):
+        x_position = ((i * grid_extents / image_resolution)
+                      - 0.5 * grid_extents)
+        for j in range(image_resolution):
+            y_position = ((j * grid_extents / image_resolution)
+                          - 0.5 * grid_extents)
+            if x_position ** 2 + y_position ** 2 <= radius ** 2:
+                out[i, j] = 1.0
```
Amaras, line 65: There should probably be another way to do that one, but that not really that necessary to implement a clearer way.

Amaras (approved):
Now that the code is much cleaner, clearer, and runs like the Python code, I don't have anything against it.
The `create_circle` could be a bit clearer, but I don't care that much for it.
That's good enough for me, @leios you can probably merge after the action builds.

contents/convolutions/2d/2d.md
```diff
@@ -135,6 +139,8 @@ In code, the Sobel operator involves first finding the operators in $$x$$ and $$
 {% method %}
 {% sample lang="jl" %}
 [import:49-63, lang:"julia"](../code/julia/2d_convolution.jl)
+{% sample lang="py" %}
+[import:39-54, lang:"python"](../code/python/2d_convolution.py)
```
leios, line 143:
```suggestion
[import:36-52, lang:"python"](../code/python/2d_convolution.py)
```

contents/convolutions/2d/2d.md
```diff
@@ -48,6 +50,8 @@ At this stage, it is important to write some code, so we will generate a simple
 {% method %}
 {% sample lang="jl" %}
 [import:30-47, lang:"julia"](../code/julia/2d_convolution.jl)
+{% sample lang="py" %}
+[import:22-36, lang:"python"](../code/python/2d_convolution.py)
```
leios, line 54:
```suggestion
[import:21-33, lang:"python"](../code/python/2d_convolution.py)
```

leios (commented):
Just fixing the line numbers, but while running this, I noticed something odd

```
[leios@noema julia]$ time julia 2d_convolution.jl 

real    0m1.311s
user    0m1.256s
sys     0m0.053s
```

```
[leios@noema python]$ time python 2d_convolution.py 

real    3m18.118s
user    3m17.988s
sys     0m0.017s
```

Python is 151x slower?

Should we lower the resolution of the output image or maybe consider using numpy? Neither are necessary. I am happy enough with the code producing the correct output, but looking forward, we are thinking about having an overlying testing framework for a lot of the AAA and a 3 minute test is a bit long.

Ridham177:
@leios If we use numpy, it would defeat the purpose of the explicit programming. Most things would be done by numpy and wouldn't be visible in the program. We can reduce the signal dimensions to 10,10 instead of 100, 100. However, modifications to other aspects and to julia code would also be required.

leios (approved):
This looks good, thanks for the review and code!

leios:
Happy enough with this, but because you are a first-time contributor, feel free to put your name in the CONTRIBUTORS.md file as well! You don't have to if you do not want to, but I wanted to make sure you knew that was a thing!

Amaras:
Please resolve the conflict on CONTRIBUTORS.md before we can merge your PR, it's the only thing left to do (your branch seems to be late with respect to `master`)

Ridham177:
@Amaras I was lost when fixing this. Please let me know if the fix is correct.

leios (approved):
This looks good to me, Thanks a bunch!

PR #828, opened 2021-08-18
fixing huffman encoding for Julia and adding Test

### LINKED ISSUE #659: Edge case fail in all Huffman Encoding implementations.

leios:
This fixes #659 for Julia. I honestly thought I had done this already, but apparently I missed it!

I also took the liberty to add Test and use the `@testset` macro


contents/huffman_encoding/code/julia/huffman.jl
```diff
@@ -107,6 +116,11 @@ function two_pass_huffman(phrase::String)
     final_string = decode(huffman_tree, bitstring)
     println(bitstring)
     println(final_string)
+    return final_string
```
Amaras, line 119: Maybe you shouldn't `println` if you return from the function and  instead do all the printing in the main namespace or in a main function.

leios: That's a good point. I should probably roll this whole function into a `@testset` and test each component (generating the codebook, encoding, and decoding) separately for each of the different strings. It might still print out during the tests like we do for approximate counting, but I think that's fine for tests.

Amaras: Probably. That full test would be a kind of integration test.
I don't know how it's possible to test the generation of the codebook or Huffman tree correctly without re-implementing the functionality, though :/

leios: I was just going to test that the codebook had all the right keys and that the bitstring was correct by hard-coding correct answers. So yeah, integration test for now.

ntindle:
[lang: julia]

PR #830, opened 2021-08-20
Add Barnsley Farn java implementation

stormofice:


contents/barnsley/code/java/Barnsley.java
```diff
@@ -0,0 +1,92 @@
...
+
+        // This checks to see if a random number is in a bin, if not, that
+        // probability is subtracted from the random number and we check the
+        // next bin in the list
+        for (int i = 0; i < probabilities.length; i++) {
+            if (rand < probabilities[i])
+                return hutchinsonOp[i];
+            rand -= probabilities[i];
+        }
+        return null;
```
stormofice, line 38: This return will never be reached, but Java doesn't know that.
I don't think an explanatory comment is necessary there?

contents/barnsley/code/java/Barnsley.java
```diff
@@ -0,0 +1,92 @@
...
+
+        for (int i = 0; i < n; i++) {
+            outputPoints[i] = point;
+            point = matrixMultiplication(selectArray(hutchinsonOp, probabilities), point.toDoubleArray());
+        }
+
+        return outputPoints;
+    }
+
+    public static Point matrixMultiplication(double[][] operation, double[] point) {
```
stormofice, line 60: Java does not support matrix multiplication in its standard library, so this function was necessary.
Converting between Points and double Arrays is kind of unintuitive, but I think that Points being objects in general does make the code more readable.

Amaras (commented):
Hey there, great PR overall, though I do have a few nitpicks I'd like you to address.

First off, the matrix multiplication code suspiciously lacks the loop braces. Although it's possible for single-line for loops, I don't recommend it for *double* for loops.

Second, I think the conversions between `Point` and `double[]` is inconsistent. Why do you want to return a `Point` from a `double[]` in the `matrixMultiplication` and not just do `Point` to `Point`?

Related to this, maybe adding a `public Point(double[] array)` could be useful for more direct conversions.

Finally, and not blocking as I'm not sure if we have a style guide decision for this, but I don't think we should teach the bad practice of letting `main` throw exceptions that can crash the program at any time with a potentially cryptic message.
I think we already merged Java implementations doing exactly this, so you don't have to modify that part here.

In any case, thanks for the PR :)

contents/barnsley/code/java/Barnsley.java
```diff
@@ -0,0 +1,91 @@
...
+        double[] result = new double[3];
+        for (int j = 0; j < operation[0].length; j++)
+            for (int i = 0; i < operation.length; i++)
+                result[j] += operation[j][i] * point[i];
+        return new Point(result[0], result[1], result[2]);
+    }
+
+    public static void main(String[] args) throws IOException {
```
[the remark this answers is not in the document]
stormofice: I agree, and changed it to use a try-with statement to auto-close the file and also catch the IOException there.
I think it would improve consistency in general, if there were a list of code snippets to write data output to file for the most common languages.

contents/barnsley/code/java/Barnsley.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        return outputPoints;
+    }
+
+    public static Point matrixMultiplication(double[][] operation, double[] point) {
+        double[] result = new double[3];
+        for (int j = 0; j < operation[0].length; j++)
+            for (int i = 0; i < operation.length; i++)
+                result[j] += operation[j][i] * point[i];
```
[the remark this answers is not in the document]
stormofice: Done

contents/barnsley/code/java/Barnsley.java
stormofice: I've decided on keeping it consistent by using double arrays, because it is easier to work with them here.

However is checking for the array length in the constructor really necessary?
I do think that error handling would clutter up the algorithm a bit.

contents/barnsley/code/java/Barnsley.java
stormofice: Done

contents/barnsley/code/java/Barnsley.java
```diff
@@ -51,21 +59,24 @@ public Point(double x, double y, double z) {
...
         return outputPoints;
     }
 
-    public static Point matrixMultiplication(double[][] operation, double[] point) {
+    public static double[] matrixMultiplication(double[][] operation, double[] point) {
         double[] result = new double[3];
-        for (int j = 0; j < operation[0].length; j++)
-            for (int i = 0; i < operation.length; i++)
+        for (int j = 0; j < operation[0].length; j++) {
+            for (int i = 0; i < operation.length; i++){
```
Amaras, line 71: Inconsistent spacing

Amaras (approved):
Alright, you made your choices, and although I don't agree with all of them, it's still good code, so it has my approval.

contents/barnsley/code/java/Barnsley.java
```diff
@@ -5,7 +5,7 @@
 public class Barnsley {
 
     private static class Point {
-        double x, y, z;
+        public double x, y, z;
```
Amaras, line 8: I think the previous version would not have run without the `public` keyword, as the default protection in Java is `private`, if I remember correctly. The names would have been out of scope for the `main` function.

Amaras (commented):
No problem with that code.
We discussed the implementation of the matrix multiplication privately on Discord and @stormofice used the version I gave them, so I'm not sure whether I can review this version.

contents/barnsley/code/java/Barnsley.java
stormofice: The default is package private, which means that it is visible to classes in the same packet.
While it would have worked, I thought it makes sense to specify it here concretely.

Amaras: Fair enough

PR #834, opened 2021-08-22
Add approximate counting C++ sample code

mika314:


Amaras:
Thanks for the PR.
As for my previous review on #835, I'm not comfortable reviewing templated C++ code, so no approving review for me, even if I like your code.

contents/approximate_counting/code/c++/approximate_counting.cpp
```diff
@@ -0,0 +1,71 @@
...
+auto n(double v, double a) { return a * (pow((1 + 1 / a), v) - 1); }
+
+// This function takes
+//    - v: value in register
+//    - a: a scaling value for the logarithm based on Morris's paper
+// It returns a new value for v
+auto increment(double v, double a) {
+  // delta is the probability of incrementing our counter
+  const auto delta = 1 / (n(v + 1, a) - n(v, a));
+  return (drand() <= delta) ? v + 1 : v;
```
mika314, line 31: BTW Julia's version is slightly odd on this line: it does `v++` or `++v` for some reason.

leios: You are talking about these lines?

```julia
    if rand() <= delta
        return v += 1
    else
        return v
    end
```

I just like splitting them up because it's easier to read for me. I don't think its functionally different, is it?

mika314: yes, it feels like it should be:
```
    if rand() <= delta
        return v + 1
    else
        return v
    end
```

leios: oh. Yeah. You are right. Oops. That's what I get for not waiting on an appropriate review.

contents/approximate_counting/code/c++/approximate_counting.cpp
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// This simulates counting and takes
+//     - n_items: number of items to count and loop over
+//     - a: a scaling value for the logarithm based on Morris's paper
+// It returns n(v,a), the approximate count
+auto approximate_count(int n_items, int a) {
+  auto v = 0;
+  for (auto i = 0; i < n_items; ++i)
+    v = increment(v, a);
```
ShadowMitia, line 41: I have a warning for `double` to `int` conversion, because `v` is of type int and `increment` returns a `double`, which would round down the values.

I'm guessing it should be `auto v = 0.0` instead?

contents/approximate_counting/code/c++/approximate_counting.cpp
```diff
@@ -0,0 +1,71 @@
...
+  // delta is the probability of incrementing our counter
+  const auto delta = 1 / (n(v + 1, a) - n(v, a));
+  return (drand() <= delta) ? v + 1 : v;
+}
+
+// This simulates counting and takes
+//     - n_items: number of items to count and loop over
+//     - a: a scaling value for the logarithm based on Morris's paper
+// It returns n(v,a), the approximate count
+auto approximate_count(int n_items, int a) {
```
ShadowMitia, line 38: I have a warning on ligne 56 about `double` to `int` conversion for the value `a`. A `double` is given in, and then it's passed to functions that require it to be `double`. Maybe it should be `double a`?

mika314: I want to reproduce the issue, what compiler are you using, and compilation flags?

contents/approximate_counting/code/c++/approximate_counting.cpp
mika314: I want to reproduce the issue, what compiler are you using, and compilation flags?

contents/approximate_counting/code/c++/approximate_counting.cpp
ShadowMitia: Sure, I was using GCC with "g++ -std=c++17  -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wold-style-cast -Wcast-align -Wunused -Woverloaded-virtual -Wpedantic -Wconversion -Wsign-conversion -Wnull-dereference -Wdouble-promotion -Wformat=2"

mika314: Yes, it should be `double`, I am extremely confused about how it passed the last test case with a=0.5. Let me spend some time with the test cases.

mika314: Mitia, I fix warnings also found the bug in the test.

mika314: I filed the simular issue on Julia code:
https://github.com/algorithm-archivists/algorithm-archive/issues/843

contents/approximate_counting/code/c++/approximate_counting.cpp
mika314: James, I approved Julia's PR

ShadowMitia (approved):

leios (approved):
Looks good! Thanks for the submission and thanks @ShadowMitia for the review!

PR #835, opened 2021-08-22
C++ sample code for The Barnsley Fern

mika314:
The Barnsley Fern with 100000 points
image

### DIFF
--- a/contents/barnsley/code/c++/barnsley.cpp
+++ b/contents/barnsley/code/c++/barnsley.cpp
@@ -1,3 +1,6 @@
+// The code bellow uses C++-17 features, compile it with C++-17 flags, e.g.:
+// clang++ -Wall -Wextra -Wshadow -Wnon-virtual-dtor -Wold-style-cast -Wcast-align -Wunused -Woverloaded-virtual -Wpedantic -Wconversion -Wsign-conversion -Wnull-dereference -Wdouble-promotion -Wformat=2 -gdwarf-3 -D_GLIBCXX_DEBUG -std=c++17 -O3 -c ./barnsley.cpp barnsley
+
 #include <array>
 #include <cassert>
 #include <fstream>
--- a/contents/barnsley/code/c++/barnsley.cpp
+++ b/contents/barnsley/code/c++/barnsley.cpp
@@ -11,7 +11,7 @@ using Vec3 = std::array<double, 3>;
 using Row = std::array<double, 3>;
 using Op = std::array<Row, 3>;
 
-constexpr auto OpN = 4;
+constexpr auto OpN = 4U;
 
 template <size_t N>
 auto operator+(std::array<double, N> x, std::array<double, N> y) {
--- a/contents/barnsley/code/c++/barnsley.cpp
+++ b/contents/barnsley/code/c++/barnsley.cpp
@@ -1,0 +1,118 @@
+#include <array>
+#include <cassert>
+#include <fstream>
+#include <random>
+
+using Vec2 = std::array<double, 2>;
+using Vec3 = std::array<double, 3>;
+using Row = std::array<double, 3>;
+using Op = std::array<Row, 3>;
+
+constexpr auto OpN = 4;
+
+template <int N>
+auto operator+(std::array<double, N> x, std::array<double, N> y) {
+  for (auto i = 0U; i < N; ++i)
+    x[i] += y[i];
+  return x;
+}
+
+template <int N>
+auto operator*(double k, std::array<double, N> v) {
+  for (auto i = 0U; i < N; ++i)
+    v[i] *= k;
+  return v;
+}
+
+template <int N>
+auto operator*(std::array<double, N> v, double k) {
+  return k * v;
+}
+
+auto operator*(const Op& x, const Vec3& y) {
+  auto ret = Vec3{};
+  for (auto i = 0; i < 3; ++i) {
+    ret[i] = 0;
+    for (auto j = 0; j < 3; ++j)
+      ret[i] += y[j] * x[i][j];
+  }
+  return ret;
+}
+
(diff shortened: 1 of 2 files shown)


Amaras:
Thanks for the PR
I am not completely comfortable reviewing templated C++ code, but I quite like what you've done here.
From what I can see, it's all correct, and the fern looks good, so I am sure it's a good fit for the AAA, but I'd like to get a second opinion to be sure.

contents/barnsley/code/c++/barnsley.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <cassert>
+#include <fstream>
+#include <random>
+
+using Vec2 = std::array<double, 2>;
+using Vec3 = std::array<double, 3>;
+using Row = std::array<double, 3>;
+using Op = std::array<Row, 3>;
+
+constexpr auto OpN = 4;
```
ShadowMitia, line 11: This should probably be 4U, because of signedness.

contents/barnsley/code/c++/barnsley.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+template <int N>
```
ShadowMitia, line 13: this should probably be template <unsigned int N>

contents/barnsley/code/c++/barnsley.cpp
```diff
@@ -0,0 +1,118 @@
...
+auto operator+(std::array<double, N> x, std::array<double, N> y) {
+  for (auto i = 0U; i < N; ++i)
+    x[i] += y[i];
+  return x;
+}
+
+template <int N>
```
ShadowMitia, line 20: same, this should probably be template <unsigned int N>

contents/barnsley/code/c++/barnsley.cpp
```diff
@@ -0,0 +1,118 @@
...
+auto operator*(double k, std::array<double, N> v) {
+  for (auto i = 0U; i < N; ++i)
+    v[i] *= k;
+  return v;
+}
+
+template <int N>
```
ShadowMitia, line 27: same, this should probably be template <unsigned int N>

contents/barnsley/code/c++/barnsley.cpp
```diff
@@ -0,0 +1,118 @@
...
+auto operator*(std::array<double, N> v, double k) {
+  return k * v;
+}
+
+auto operator*(const Op& x, const Vec3& y) {
+  auto ret = Vec3{};
+  for (auto i = 0; i < 3; ++i) {
```
ShadowMitia, line 34: should probably be `auto i = 0U`

contents/barnsley/code/c++/barnsley.cpp
```diff
@@ -0,0 +1,118 @@
...
+    ret[i] = 0;
+    for (auto j = 0; j < 3; ++j)
```
ShadowMitia, line 36: should probably be auto j = 0U

contents/barnsley/code/c++/barnsley.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  // This checks to see if a random number is in a bin, if not, that
+  //     probability is subtracted from the random number and we check the
+  //     next bin in the list
+  for (auto i = 0U; i < probabilities.size(); ++i) {
+    if (rnd < probabilities[i])
+      return hutchinson_op[i];
+    rnd -= probabilities[i];
+  }
+  assert(!"check if probabilities adding up to 1");
```
ShadowMitia, line 74: I'm not sure what this does?

ShadowMitia (commented):
I've reviewed a few things, this should disable warnings about "unsigned/signed" conversions.

Otherwise looks good to me.

It should be probably written somewhere that this is C++17 code?

contents/barnsley/code/c++/barnsley.cpp
mika314: it is a template. What do you mean?

contents/barnsley/code/c++/barnsley.cpp
mika314: what is your motivation? I do not understand why it should be unsigned?

ShadowMitia: Because I get warnings of signed/unsigned comparisons. Sorry I should've been clearer on this...

contents/barnsley/code/c++/barnsley.cpp
ShadowMitia: Wooops, not sure what happened here.
I meant, this should probably be `template <unsigned int N>`, again because of signed/unsigned comparisons.

contents/barnsley/code/c++/barnsley.cpp
mika314: If probabilities are not adding up to 1, there is a chance the code will fall through the loop. Or you do not know what assert does?
https://en.cppreference.com/w/cpp/error/assert

ShadowMitia: I'm not super famlliary with them, but it seems you're not testing anything with it?

Amaras: > I'm not super famlliary with them, but it seems you're not testing anything with it?

Basically, it's an `assert(false);`, but with a message.

ShadowMitia: Right ok, that makes sense now! That should be fine then.

mika314:
Mitia, I pushed fixes.

ShadowMitia (approved):

leios (approved):
Thanks again for the code and thanks to @ShadowMitia for the review!

PR #836, opened 2021-08-22
Flood fill in Coconut

Amaras:
This replaces #743, because I merged the master branch, and thought it could have been a good idea.
Anyway, here's the good one, normally. It should have the code in the same state as the last commit from the previous PR.

I'm probably going to delete the old branch in a few days (if I remember it then)

### DIFF
--- a/contents/flood_fill/code/coconut/flood_fill.coco
+++ b/contents/flood_fill/code/coconut/flood_fill.coco
@@ -12,21 +12,12 @@ def inbounds(canvas_shape, location is Point) =
     min(location) >= 0 and location.x < canvas_shape[0] and location.y < canvas_shape[1]
 
 
-def colour(canvas, location is Point, old_value, new_value):
-    if not inbounds(canvas.shape, location):
-        return
-
-    if canvas[location] != old_value:
-        return
-    else:
-        canvas[location] = new_value
-
-
 def find_neighbours(canvas, location is Point, old_value, new_value):
     possible_neighbours = ((Point(0, 1), Point(1, 0), Point(0, -1), Point(-1, 0))
                           |> map$(location.__add__))
 
-    yield from possible_neighbours |> filter$(x -> (inbounds(canvas.shape, x) and canvas[x] == old_value))
+    yield from possible_neighbours |> filter$(x -> (inbounds(canvas.shape, x)
+                                                    and canvas[x] == old_value))
 
 
 def stack_fill(canvas, location is Point, old_value, new_value):
@@ -39,9 +30,8 @@ def stack_fill(canvas, location is Point, old_value, new_value):
         current_location = stack.pop()
         if canvas[current_location] == old_value:
             canvas[current_location] = new_value
-            for neighbour in find_neighbours(canvas, current_location, old_value,
-                                             new_value):
-                stack.append(neighbour)
+            stack.extend(find_neighbours(canvas, current_location, old_value,
+                                         new_value))
 
 
 def queue_fill(canvas, location is Point, old_value, new_value):
@@ -57,8 +47,8 @@ def queue_fill(canvas, location is Point, old_value, new_value):
         current_location = queue.popleft()
         for neighbour in find_neighbours(canvas, current_location, old_value,
                                          new_value):
-            queue.append(neighbour)
             canvas[neighbour] = new_value
+            queue.append(neighbour)
 
 
 def recursive_fill(canvas, location is Point, old_value, new_value):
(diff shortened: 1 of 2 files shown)


lazyprop:
@Amaras The `inbounds` check is already being done once in `find_neighbours` so it doesn't need to be done again in the `color` function. I made an issue (#765) regarding this.  Just a friendly reminder :) 
Also, I have already fixed it in the Python implementation so you might want to take a look at it

contents/flood_fill/code/coconut/flood_fill.coco
```diff
@@ -0,0 +1,123 @@
...
+    def __add__(self, other is Point) = Point(self.x + other.x, self.y + other.y)
+
+
+# This function is necessary, because negative indices wrap around the
+# array in Coconut.
+def inbounds(canvas_shape, location is Point) =
+    min(location) >= 0 and location.x < canvas_shape[0] and location.y < canvas_shape[1]
+
+
+def find_neighbours(canvas, location is Point, old_value, new_value):
```
berquist, line 15: I don't like how this carries over the unused `new_value` from the other implementations, which similarly don't use it.

contents/flood_fill/code/coconut/flood_fill.coco
```diff
@@ -0,0 +1,123 @@
...
+        assert (grid == solution_grid).all()
+    except AssertionError:
+        print('F', end='')
+        failure_count += 1
+    else:
+        print('.', end='')
+        pass_count += 1
+
+    print('')
+    print(TestResults(pass_count, failure_count))
```
berquist, line 116: This is a nice touch, but will probably have to be reworked when output standardization is done.

contents/flood_fill/code/coconut/flood_fill.coco
```diff
@@ -0,0 +1,123 @@
...
+    grid[2,:] = 1
+    solution_grid = np.zeros((5, 5))
+    solution_grid[:3,] = 1
+
+    starting_location = Point(0, 0)
+
+
+    # The following is manual unit testing of the function
+    recursive_fill(grid, starting_location, 0, 1)
+    try:
```
berquist, line 81: Most of this can be lifted into a single testing function that takes in the fill function as an argument, the starting grid, and the expected final grid, encapsulating the try/except/else logic.

berquist (changes requested):
I'm happy with the implementation and clarity, and confirmed it runs under `--strict`, just some minor cleanup stuff.

contents/flood_fill/code/coconut/flood_fill.coco
Amaras: Definitely, I think we'll just have to remove the `test` function for standardization.

contents/flood_fill/code/coconut/flood_fill.coco
Amaras: Both `old_value` and `new_value` are used in all currently merged implementations an in mine.

contents/flood_fill/code/coconut/flood_fill.coco
Amaras: Done

lazyprop:
Also, the `inbounds` check that is being done in `stack_fill`, `queue_fill` and `recursive_fill` can be removed. This is because it is being done once in `find_neighbors`. It need not be checked again. The Julia and Python implementations have been changed accordingly.

Amaras:
> Also, the `inbounds` check that is being done in `stack_fill`, `queue_fill` and `recursive_fill` can be removed. This is because it is being done once in `find_neighbors`. It need not be checked again. The Julia and Python implementations have been changed accordingly.

You can't ensure the functions are called with a point that is not already in bounds. The initial point is *never* checked otherwise.
I could agree for the `recursive_fill`, but the initial point is always an edge case

contents/flood_fill/code/coconut/flood_fill.coco
berquist: Not in this function.

Amaras: Fair enough, it's now corrected

berquist (approved):

lazyprop:
@Amaras you're right. Should this be changed in the other two implementations also?

Amaras:
> @Amaras you're right. Should this be changed in the other two implementations also?

I don't know, maybe we should rather open an issue for that, especially if an issue was opened to remove the checks.

PR #807, opened 2021-03-22
Add Computus in Javascript and Typescript

hugo-s29:
A translation of the computus Julia code in javascript and in typescript.

PR #845, opened 2021-08-24
Remove case sensitive paths

hugo-s29:
The files `Sobel_filters.png` and `sobel_filters.png` (`in contents/convolutions/res`) can't coexist on Windows.
I've checked, only `sobel_filters.png` is used and both images seem to be identical.
So, I've removed the other one.

This issue was discussed in PR #807.


Amaras (approved):
Alright, if you've checked and it's the only thing, then I'm all in on it, and merge as soon as the GH action build finishes.

leios (approved):
I have to be honest, I have no idea where the upper-cased one came from. I never use that convention, so it must have slipped through the cracks. Thanks for the catch and sorry for the troubles!

PR #851, opened 2021-08-30
[Haskell] Barnsley Fern

jiegillet:
Seeing as #704 was reviewed recently, I'm pushing forward with Barnsley Fern.

I also fixed a typo in the chapter: `022` -> `0.22`.

Output for 100k points:
Screen Shot 2021-08-30 at 12 31 21


contents/barnsley/code/haskell/Barnsley.hs
```diff
@@ -0,0 +1,37 @@
+import Data.Array (Array, bounds, listArray, (!))
+import Data.List (intercalate)
+import System.Random
+
+data Point = Point Double Double
+
```
NikulinVs, line 6: {- I think it is worth to combine probabilities and hutchinson arguments in one.
It is not clear from this signature that length of probabilities list should be the same as length of hutchinson array.
Make it into something like:
​
chaosGame :: RandomGen g => g -> Int -> Array Int (Double, (Point -> Point)) -> [Point]
​
-}

contents/barnsley/code/haskell/Barnsley.hs
```diff
@@ -0,0 +1,37 @@
...
+chaosGame :: RandomGen g => g -> Int -> [Double] -> Array Int (Point -> Point) -> [Point]
+chaosGame g n probabilities hutchinson = take n points
+  where
+    (x, g') = random g
+    (y, g'') = random g'
+
+    picks = randomRs (0, 1) g''
+    cumulProbabilities = scanl1 (+) probabilities
+    to_choice x = (+ 1) $ length $ takeWhile (x >) cumulProbabilities
+
```
NikulinVs, line 16: ​
    {-
       Very long line. Better to cut it into more readable pieces.
       For example,
​
       choices = map to_choice picks
       points = Point x y : zipWith (hutchinson !) choices points
​
    -}

contents/barnsley/code/haskell/Barnsley.hs
```diff
@@ -0,0 +1,37 @@
...
+    points = Point x y : zipWith (hutchinson !) (map to_choice picks) points
+
```
NikulinVs, line 18: {-
​
Here is alternative code with IO monad instead of explicit Random Generator.
You do random generation in sequence anyway (through g, g' and g''),
so why not let IO monad care about the sequential stuff for you?
Then you won't even need newStdGen in main function, you work in IO already.
​
(This is the code without taking into account my other comments)
​

​
chaosGame :: Int -> [Double] -> Array Int (Point -> Point) -> IO [Point]
chaosGame n probabilities hutchinson = do
    x <- randomIO
    y <- randomIO
​
    picks <- replicateM n $ randomRIO (0, 1)
    let cumulProbabilities = scanl1 (+) probabilities
        to_choice x = (+ 1) $ length $ takeWhile (x >) cumulProbabilities
​
        points = Point x y : zipWith (hutchinson !) (map to_choice picks) points
    return points
-}

contents/barnsley/code/haskell/Barnsley.hs
```diff
@@ -0,0 +1,37 @@
...
+main :: IO ()
+main = do
+  g <- newStdGen
+
```
NikulinVs, line 22: -- Maybe worth putting this definition of affine outside of main and provide type annotation?
-- Also, why lists as inputs when tuples will suffice?

contents/barnsley/code/haskell/Barnsley.hs
```diff
@@ -0,0 +1,37 @@
...
+  let affine [xx, xy, yx, yy] [a, b] (Point x y) =
+        Point (a + xx * x + xy * y) (b + yx * x + yy * y)
+      barnsley =
+        listArray
+          (1, 4)
```
NikulinVs, line 27: {-
            Why (1, 4)?
            (0, 3) more natural. And you won't need (+ 1) in "to_choice" function inside chaosGame
        -}

contents/barnsley/code/haskell/Barnsley.hs
jiegillet: Good idea, done!

contents/barnsley/code/haskell/Barnsley.hs
jiegillet: I usually try to stay away from `IO` until the very end. I feel code is more portable when it's pure. Yes, it's true that I'm doing `IO` in the end, but I want to focus more on the algorithm, and less on the task...

leios:
What do you think? Is this close to mergable?

jiegillet:
@leios  I probed Seva a couple of times, but I guess he was too busy to check. He told before me that the initial code was already fine, and I've integrated all of his suggestions but one (with a valid reason) so I think this can be approved and merged, honestly. By you probably :)
Sorry, looks like all the Haskell PRs are irregular somehow.

leios:
Ok, I thought that was what happened. Happy to merge it then!

leios (approved):

PR #854, opened 2021-08-31
[#852] [#852] Fix for Julia 1D convolution

### LINKED ISSUE #852: 1D Convolution - always a 0 element
This is possibly talked about in #202 but I'm pretty sure the chapter on convolution came after that issue...

# Bug Report

## Description

There is always a 0 element in the linear convolution.

### Steps to Reproduce

The code says
```julia
# full convolution, output will be the size of x + y
full_linear_output = convolve_linear(x, y, length(x) + length(y))
```

for example,
```Julia
julia> convolve_linear([1,2], [7, 8], 4)
4-element Array{Float64,1}:
  7.0
 22.0
 16.0
  0.0
```

The last element is always 0 because it corresponds to a situation when the signal and filter functions don't overlap at all. 

```
  1 2  1 2  1 2    1 2
  |    | |    |
8 7    8 7    8 7     8 7
  7    22     16    0
```

### Expected behavior

```julia
# full convolution, output will be the size of x + y - 1
full_linear_output = convolve_linear(x, y, length(x) + length(y) - 1)
```

```Julia
julia> convolve_linear([1,2], [7, 8], 3)
4-element Array{Float64,1}:
  7.0
 22.0
 16.0
```

## For Algorithm Archive Developers

- [ ] The bug can be reproduced
- [ ] The bug can be fixed (if not, please explain why not in a comment below)
- [ ] There is a timeline to fix the bug
- [ ] The bug has been fixed (Please link the PR)

jiegillet:
Fixes #852. Fixes #852. 
Hopefully, since I'm not a Julia master :)

### DIFF
--- a/contents/convolutions/code/julia/1d_convolution.jl
+++ b/contents/convolutions/code/julia/1d_convolution.jl
@@ -13,9 +13,7 @@ function convolve_cyclic(signal::Array{T, 1},
 
     for i = 1:output_size
         for j = 1:output_size
-            if (mod1(i-j, output_size) <= length(filter))
-                sum += signal[mod1(j, output_size)] * filter[mod1(i-j, output_size)]
-            end
+            sum += get(signal, mod1(j, output_size), 0) * get(filter, mod1(i-j, output_size), 0)
         end
 
         out[i] = sum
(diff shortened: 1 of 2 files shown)


leios (approved):
Looks good to me! Thanks for the catch!

PR #857, opened 2021-09-03
Tree Traversal: Output Standardization

lazyprop:


lazyprop:
I have made changes to many of the implementations, but there are a few that are
problematic.

- The PHP implementations has a weird way of printing node data.
- The Haskell implementation numbers the nodes in an entire different way from
  the others.
- The Coconut implementation works slightly differently. Shouldn't be hard to
  fix it though.

I'll update this list tomorrow as I'm way too sleepy right now.

Amaras:
Oh well, apparently I have write access to your repo.
The Coconut output should be okay (it was simply a matter of adding `[#]` everywhere)

lazyprop:
@Amaras Coconut does not have any nodes containing `0` which every other language does. This can be fixed just by changing the condition in `create_tree` like this:
```python
    if num_rows == 0:
        return Node(0, ())
```

leios:
Great work with this, maybe we hide Ocaml for now, as it is missing some code snippets?

I think @jiegillet might be willing to fix the haskell? 

I don't know about PhP, though. Does it need a rework entirely?

lazyprop:
Yeah hiding Ocaml would be good for now.

I don't know Haskell but I think it can be fixed by just creating a `createTree` procedure which numbers nodes appropriately.

As for PHP, I think I have a fix. The only thing that needed changing was the `generate_tree` function (that seems to be a common problem in the implementations). I also removed an `if` condition from `DFSRecursive` so maybe the line numbers need changing. I have zero experience with PHP though so if someone can review the PHP changes that would be nice.

jiegillet:
> I think @jiegillet might be willing to fix the haskell?

Of course, please give me a bit of time.

Yeah sorry, back in the day I used to do things my way, since Haskell programs are structured quite differently anyway... I'll fall into the ranks now.

Amaras:
> @Amaras Coconut does not have any nodes containing `0` which every other language does. This can be fixed just by changing the condition in `create_tree` to this:
> 
> ```python
>     if num_rows == 0:
>         return Node(0, ())
> ```
Alright, thanks for the catch.
I also removed the outdated comments, because they would be too misleading.

jiegillet:
Added Haskell.

```
haskell git:(tree-traversal-standardize) ✗ diff --ignore-space-change <(julia ../julia/Tree.jl)  <(ghc -v0 --make TreeTraversal.hs && ./TreeTraversal)
haskell git:(tree-traversal-standardize) ✗ 
```

I used `--ignore-space-change` because Julia has an extra space at the end of the line. This is OK right?

lazyprop:
@jiegillet thanks for the fix! The extra spaces are there in most of the languages. That is because of the way they're printing during traversal. So for automating testing, we'll need to trim extra whitespaces.

lazyprop:
@jiegillet thanks for the fix! The extra spaces are there in most of the languages. That is because of the way they're printing during traversal. So for automating testing, we'll need to trim extra whitespaces.

lazyprop:
I have fixed all implementations except
- Assembly X64 - I don't know assembly
- Emojicode - esolang
- Matlab - I don't have a license so no way to test
- Ocaml - Missing some snippets. It is hidden for now.
- Scratch - No text file.

Except these five, I have manually diffed the outputs of each implementation and can confirm they print exactly the same results. I have checked line numbers for all of the implementations. A few (python, php, coconut) needed correction. This PR is now ready for review.

leios:
Matlab and ocaml do not have a full implementations, so maybe we remove them from the md file, but leave the code in the repo for those that might want to finish the implementation later?

I don't think there is much we can do about scratch, but I think assembly and emojicode *can* be updated.

lazyprop:
In accordance with the new output format, I have updated all `[#]` with `[#]\n`.

PR #858, opened 2021-09-06
Fix asm exit codes

stormofice:
This PR fixes the asm implementations which previously did not return 0 as an exit code.

The affected implementations are the following:
- Verlet Integration
- Monte Carlo Method

This was already done in some others, but this aims to standardize it across different implementations.


berquist (approved):
Confirmed that both of these work with `gcc -no-pie <filename>.s; ./a.out` and checking `echo $?` and didn't return 0 before. Thanks!

PR #861, opened 2021-09-07
Change folder structure of convolutions

stormofice:
This puts every chapters' code into its own directory, which is useful for automatic validation.


Amaras (approved):
No problem with me, the structure seems more logical that way.

berquist (approved):
I've confirmed that this works. Thanks!

PR #862, opened 2021-09-08
Add initial devcontainers

ntindle:
By no means should this be considered complete.

Some of the todos..... 
- Publish this dockerfile as a container to the GH registry so people don't have to build it themselves. 
- Add more languages
- Make sure all the tar stuff works for real

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,106 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.187.0/containers/ubuntu/.devcontainer/base.Dockerfile
+
+# [Choice] Ubuntu version: bionic, focal
+ARG VARIANT="focal"
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+
+# [Optional] Uncomment this section to install additional OS packages.
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,106 @@
...
+
+# Setup D Lang
+RUN sudo sh -c 'mkdir -p ~/dlang && wget https://dlang.org/install.sh -O ~/dlang/install.sh'
+RUN sudo sh -c 'bash ~/dlang/install.sh'
+## From Docs not needed though
+# RUN sudo sh -c 'source ~/dlang/dmd-2.097.2/activate'
+ENV PATH=$PATH:/root/dlang/dmd-2.097.2/linux/bin64
+
+# Setup Go
+RUN sudo sh -c 'wget -c https://dl.google.com/go/go1.14.2.linux-amd64.tar.gz -O - | sudo tar -xz -C /usr/local'
```
leios, line 41: to extract, isn't it usually `tar xvzf`? You can drop the `v`, but shouldn't it be `xvf`, then?

leios:
I am wondering if we should tag people who know the missing languages to add the set-up to this file?

I am also not quite sure what you mean by the first point?

Outside of that, I am a big fan of this.

ntindle:
IMO this is better to be in sooner and open issues for the languages that need fixed as hacktoberfest good first issues

PR #863, opened 2021-09-08
Scons compilation examples

Amaras:
This PR aims to provide examples on how to create compilation (and run) instructions with SCons, as per #691.
As this is currently the only effort to provide concrete examples of compilation instructions, I highly encourage others to open their own draft PRs with their own compilation instructions, until we decide what system to use.
I'd like to give the systems a fair comparison, so try to provide them for the same chapter(s) as I did (in other words, please don't get too ahead of me.)

Below is a basic instruction manual for the Barnsley Fern chapter
## How to use (e.g. create the `out.dat` or `barnsley.dat` file)
1. go to the directory you want to compile code for;
2. run `scons out.dat` or `scons barnsley.dat` (depending on the implementation)
3. (optional) clean the directory with `scons -c .` to remove all the intermediate and final files.

Note: if you want to only clean the `out.dat` file and keep the executables/classes (and so on) files, you can run `scons -c out.dat`


Amaras:
Hum... I think I branched off a PR branch... I'll try to get it correct later or reopen the PR branching off `master`.

The point still stands, as there should not be a problem with that (especially since the SCons PR should be merged after the C++ flood fill one)

ntindle:
Are we doing it this way as opposed to @stormofice 's implementation

Amaras:
> Are we doing it this way as opposed to @stormofice 's implementation

This is one way to do it. I'm not sure if it's the best way.

stormofice:
Does SCons work with every language?

Regardless of that, I don't see why this would interfere with the verifier, as better compilation/running instructions would just simplify the language configuration files.

Amaras:
> Does SCons work with every language?

Out of the box, SCons works with C/C++, Fortran, D, TeX/LaTeX, Java, some Assembler languages, and some other languages. We'd probably need to provide Builders to ensure consistent compilation for Rust, Scala, etc. 
You can see #691 for more detail about my questioning

leios:
I think this is fine and if it makes things easier for @stormofice, then let's do it!

I will say that we probably don't need the flood fill example in this PR, though

Amaras:
> I will say that we probably don't need the flood fill example in this PR, though
Yeah, I'm not exactly sure how to remove the flood fill thingy, properly :/

Amaras:
I'm not sure if I removed the previous branch commits properly, but it should be okay.

PR #846, opened 2021-08-25
Add C# implemenation for 1D Convolutions

stormofice:
Requires .NET 5.0 [[SDK Install instructions](https://dotnet.microsoft.com/download/dotnet/5.0)]

If you want to test this implementation on your machine, create a file in the `code/csharp` directory called `code.csproj` and paste the following into it:
```
<Project Sdk="Microsoft.NET.Sdk">

    <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>net5.0</TargetFramework>
    </PropertyGroup>

</Project>
```
Now you can execute `dotnet run` to run the implementation.

PR #865, opened 2021-09-14
Add C# implementation for the approximate counting algorithm

stormofice:
How to run this is described in #846.


stormofice:
Hi, thanks for the review.

I didn't use docstrings originally, as there was some discussion on whether we should use language specific documentation style comments at all, but I agree that it looks way better and it makes sense here imo.

I also updated the error thresholds according to #901 and the output as per #913.

PR #781, opened 2020-10-26
Clean up Monte Carlo integration in Racket

berquist:

PR #884, opened 2021-10-17
Change Racket lang include from lisp to racket for Euclidean algorithm

berquist:
Following the request in https://github.com/algorithm-archivists/algorithm-archive/pull/781#discussion_r709372440, this changes the only other Racket language include from `lisp` to `racket`, though it looks like this isn't defined for Prism.


leios (approved):
Sorry it took me a while to see this one, but it's a quick, positive change

PR #895, opened 2021-10-23
Added Iterated function system in PowerShell

PaddyKe:
To try the code on systems other than Windows, you can run the code here: https://code.labstack.com/powershell
It will generate a file that you can download afterwards.


stormofice (approved):
Hi, thanks for the contribution!

The code looks clean and the resulting image is similar to the ones generated by the other implementations.

However it should be noted that a PowerShell version greater or equals than `7.0` is required to run this script. 

I don't know if this is the same for the existing PowerShell implementations, but it is something to keep in mind for the automatic compilation/interpretation with SConscript.

[The devcontainer seems to always install the latest PowerShell (currently `7.2.0.1`), so it should not be a problem there]

PR #909, opened 2021-11-01
implemented stack and queue in Go

henrikac:
This PR implements stack and queue in Go.

*Note:* `code/go/stack-queue.go`also includes a `func main` where all the tests are written but I'm not sure where to include these (or if they are even required?)

### DIFF
--- a/contents/stacks_and_queues/code/go/stack-queue.go
+++ b/contents/stacks_and_queues/code/go/stack-queue.go
@@ -1,0 +1,138 @@
+package main
+
+import (
+	"fmt"
+	"os"
+)
+
+type stackNode struct {
+	value interface{}
+	prev  *stackNode
+}
+
+type Stack struct {
+	top    *stackNode
+	length int
+}
+
+// NewStack initializes a new stack.
+func NewStack() *Stack {
+	return &Stack{top: nil, length: 0}
+}
+
+// Push adds an element to the stack and returns the length of the stack.
+func (s *Stack) Push(v interface{}) int {
+	n := &stackNode{value: v, prev: s.top}
+	s.top = n
+	s.length++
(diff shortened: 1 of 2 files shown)


depate (approved):
Looks good to me!

I like the use of the empty interface to make it more general.

One could think about commenting that behind the declaration to make it clear to beginners what `interface{}` means in this context, but maybe Leios wants to decide here. 

> Note: code/go/stack-queue.goalso includes a func main where all the tests are written but I'm not sure where to include these (or if they are even required?)

The code is able to run with `go run stack-queue.go` on its own that way, that's enough. It doesn't need to be embedded somewhere. One could move the "test code" into a separate function and call it in `main()` that's how it's sometimes done, e.g. Monte Carlo and others. But that's not a must either, I think.

henrikac:
> One could think about commenting that behind the declaration to make it clear to beginners what interface{} means in this context

I'm all for making comments in the code if there is some part of an algorithm that can be difficult to understand but I am not a fan of making comments describing language syntax, e.g. what `T` is in `Array<T>` in Java.

Amaras:
> The code is able to run with go run stack-queue.go on its own that way, that's enough. It doesn't need to be embedded somewhere.

We want to produce executables (as much as possible for all compiled languages) with the build system in construction, so I am not sure this point is valid right now.

leios:
@Amaras Does this PR conflict with the build system?

Amaras:
We haven't implemented the Go build system yet, but we need to think about the future.
It doesn't conflict with anything yet, though

henrikac:
@Amaras do I need to make any changes before this can get merged?

henrikac:
`go run stack-queue.go` should work

PR #911, opened 2021-11-02
Add devcontainer setup for Elm and fix Euler Elm example

ShadowMitia:
Add setup for Elm, part of #871


ntindle:
It seems to me the elm script I picked to test this is incompatible with the current version of elm. Not sure why they would change the whole process but did you get any elm to run with the binaries that are installed?

Also the binaries work fine but it feels weird to merge if they can't run our elm.

jiegillet:
The Elm contribution is mine, it's written in Elm 0.18, since then 0.19.1 came out, and that's the version of the binary.
I'm willing to update the Elm code to 0.19.1. Would that solve the issue?

ShadowMitia:
I don't know anything about Elm, I think the Elm script didn't work on my end either, but I figured it installed properly so it was good enough.

Has Elm changed that much between versions?

ntindle:
> The Elm contribution is mine, it's written in Elm 0.18, since then 0.19.1 came out, and that's the version of the binary.
> 
> I'm willing to update the Elm code to 0.19.1. Would that solve the issue?

I believe that would solve the issue. Thanks for being willing to update

jiegillet:
> Has Elm changed that much between versions?

Yeah, there were a bunch of breaking changes from 0.18 to 0.19, but 0.19 is supposed to be stable. 

Elm is a frontend development language, so it's not something you can run as is like a script. You can use the `elm` executable to create an `.html` or `.js` file, or use it to run a local server for development.

jiegillet:
Is it cool to push to this branch? I'm afraid very few people know Elm, so waiting for a review would take ages. However, this is just an update, I had to change to different packages, etc, but the algorithm itself wasn't touched at all.

If you run `elm make src/Euler.elm` form the `elm` directory, you should get a `index.html` that you can open in the browser. If that works, it's all good.

ntindle:
Yes it's okay. I'll review later today

ntindle:
Okay great news! This certainly runs. I have no clue how to review if it "works" from an algo perspective. Here is the output. @leios  or someone else with good math skills please run this (after unzipping) wiht something like 

`python3 -m http.server 9929` 
to serve it and verify its functional. 
[index.zip](https://github.com/algorithm-archivists/algorithm-archive/files/7488557/index.zip)


The dev container works though.

ntindle (approved):

ShadowMitia:
@jiegillet I found a bug where when you press `start` and then during the animation press `start` again, the animation runs, but it doesn't correspond to the original animation. Is this a bug in the migration? I couldn't make the code run on an older version of elm...

jiegillet:
Haha whoops, that was just me being sloppy, 0.18 or 0.19. I fixed it :)

PR #898, opened 2021-10-23
Add Approximate counting in Java

PaddyKe:

PR #913, opened 2021-11-03
attempt at output standardization for approximate counting

### LINKED ISSUE #907: inconsistent approximate counting implementation
The ***Approximate Counting*** algorithm currently has four implementations and four different ways of how to act on the result of `abs((avg - n_items) / n_items) < threshold`.

| Julia | What it does |
| --- | --- |
| Julia | Returns `true` or `false` |
| Python | prints `"passed"` if the condition is `true` |
| C++ | returns `"pass"` if the condition is `true`, otherwise `"fail"` |
| C | aborts program execution |

leios:
This is an attempt to fix #907 and was also mentioned in #898


stormofice (commented):
Hi, this looks good to me, however I have two small nitpicks regarding this PR.

Shouldn't the first line ("Counting Tests, 100 trials") also start with `[#]\n`, as its output should not be verified?

Also the python implementation includes a newline after the first line, which isn't present in the other implementations. This does not matter for automatic validation, as empty lines are skipped, but it seems like an unnecessary inconsistency.

leios:
Is this ready to be merged @stormofice

leios:
Great, thanks!

PR #914, opened 2021-11-05
Add devcontainer setup for whitespace

ShadowMitia:
Add setup for whitespace language, part of #871 

I found a slightly patched code of the original whitespace compiler (so that it actually compiles). The source seems trustworthy, and it's THE reference compiler out there :grin: 

If you try the code in the AAA it wants inputs (it will just seem to hang, it's actually wanting input of two integers).


ShadowMitia:
For reference the original downloads are here https://web.archive.org/web/20130908095715/http://compsoc.dur.ac.uk/whitespace/download.php

ntindle:
Run whitespace code with `wspace`

ntindle (approved):

ntindle:
Whoops.. this commented out some langs in the container

PR #918, opened 2021-11-06
Add setup for devcontainer for Piet

ShadowMitia:
Adds setup for devcontainer for Piet. Part of #871.


ntindle:
This works, albeit it is a bit convoluted. I used the tool to transpile to python3 and executed it there. That is okay with me though.

ntindle (approved):

PR #920, opened 2021-11-06
Standardise code directory names

ShadowMitia:
This PR standardises all directories that might have different names for the same language. Eg. c++ -> cpp. This should help with #906.

For `go/golang` like names I decided to use only the name (dropping the lang part).


henrikac (commented):
looks like you forgot to update the reference to these code examples in `<algorithm>.md` e.g. `tree_traversal/tree_traversal.md`

ShadowMitia:
Oooups :sweat_smile:

ShadowMitia:
For reference I renamed folders with this command `find . -name golang -type d | rename 's/golang/go/'` from the root of AAA.

ShadowMitia:
For reference, this command helps rename paths in files grep -RiIl 'c++' | xargs sed -i 's/c++/cpp/g'. You have to double check because it might rename things that are not paths (eg. example command in C++ which has a `-std=c++17` flag).

henrikac:
The changes in the "Files changed" tap looks fine to me.

> You have to double check because it might rename things that are not paths (eg. example command in C++ which has a -std=c++17 flag).

This is, imo, something that you should have checked before pushing ;)

ShadowMitia:
Nah, I would never do that. I just forgot to change the paths in all the files that needed it. That's perfectly fine :grin: 

What I REALLY shouldn't do is work on contributions late at night.

ntindle (approved):
LGTM

ntindle:
Lots of files changed here. Would appreciate two reviews for this one

ShadowMitia:
Also let me know if there are other folders that need renaming.

ntindle:
I would prefer to do them in a separate PR if so due to size and risk mitigation

ShadowMitia:
Fair enough

leios (approved):
I also double-checked and I think its good. Thanks again for all the effort here!

PR #923, opened 2021-11-07
removing Euler and adjacent chapters

leios:
After revisiting some older chapters, I have decided to remove the Euler chapter. This is because of the following reasons:
1. The chapter was not technically wrong, but it was very misleading as to how people truly implement Euer methods i practice. I think starting from the Taylor series makes sense from a physical point of view, but not mathematically
2. The code basically had the derivative hard-coded in, which is not how people often implement this.

I could re-work the chapter to keep the code, but I would need another chapter to link it to RK, Bakcwards Euler, or other methods. I think a full re-work is needed and is on the table in a few weeks when I am working on my next project (which will involve RK).


ntindle:
I would prefer not to delete this. I would rather unlink it on the site so that we may be able to use it later 

Thoughts?

leios:
Sure, that's also fine

ntindle (approved):

PR #924, opened 2021-11-08
Implement Stacks and Queues in Rust.

PeanutbutterWarrior:
I've done the same as the Java and Typescript examples code-wise, however I've added minimal code comments. One is Rust specific, pointing out where a reference is returned instead of the value due to the borrow checker. I've also added comments after the print statements in the test to show the output of them.


henrikac (commented):
looks fine to me

contents/stacks_and_queues/code/rust/Queue.rs
```diff
@@ -0,0 +1,47 @@
...
+       }
+    }
+
+    // Note that this returns a reference to the value
+    // This is in contrast to dequeue which gives ownership of the value
+    fn front(&self) -> Option<&T> {
+        if self.list.len() > 0 {
+            Some(&self.list[0])
+        } else {
+            None
```
ishwar00, line 20: is it safe return None as front element when Queue is empty?

ShadowMitia: What do you mean safe?
You're returning an `Option<>` so returning a `None` is one of the expected values.

ishwar00: Okay _safe_ may not be correct word here.
I thought like, say i was doing something like `Q.front() * 5`, then returning `None` would not be right thing to do, right? instead executing `panic!("Attempt to access element from a empty Queue")` may be helpful.

ShadowMitia: You can't do `Q.front() * 5` anyway with an `Option`. `panic!` is not an exception, you are crashing the application to say something bad happened (and implied unrecoverable, which is not the case here).

Plus `panic!` only works at runtime, which is kinda late to warn on possible problems.

When you write a public facing function, you have to define what the function takes and returns. You could only return `T`, and crash/exception/error handle.  
But `Option` says, "ok we can have either a legit value or `None`". 
That `None` can mean many things, but it describes that no valid data can be returned. And you don't want your user to fall into cases where you get weird/undefined behaviours, so being explicit helps a lot. And the programmer than has to deal with the different cases (or can take shortcuts if he knows that special cases can't happen). 

You don't really care what the programmer does, what you do care about is correctly defining and documenting what your function does, deosn't do, and special cases. And there is always several ways to do that, but I think in this case having `Option` does the job well enough.

You can see in the standard library for example [here](https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html#method.front) that `Optional` is also used in similar cases.

ishwar00: Thank you for being so kind. 
You completely make sense.

contents/stacks_and_queues/code/rust/Stack.rs
```diff
@@ -0,0 +1,45 @@
...
+        Stack {
+            list: Vec::new(),
+        }
+    }
+
+    // Note that this returns a reference to the value
+    // This is in contrast to pop which gives ownership of the value
+    fn top(&self) -> Option<&T> {
+        if self.list.len() > 0 {
+            Some(&self.list[self.list.len() - 1])
```
ShadowMitia, line 16: It might best to use the methods in `Vec` for this
```suggestion
            self.list.last()
```

contents/stacks_and_queues/code/rust/Queue.rs
```diff
@@ -0,0 +1,47 @@
...
+       Queue{
+           list: VecDeque::new(),
+       }
+    }
+
+    // Note that this returns a reference to the value
+    // This is in contrast to dequeue which gives ownership of the value
+    fn front(&self) -> Option<&T> {
+        if self.list.len() > 0 {
+            Some(&self.list[0])
```
ShadowMitia, line 18: It might best to use the methods in VecDeque for this
```suggestion
            self.list.front()
```

ShadowMitia (changes requested):

ShadowMitia (changes requested):
This looks good overall. Could you just add a `Cargo.toml` and I think we should be good.

I think this should be enough, but feel free to edit it.

```
[package]
name = "stack_and_queue"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]


[[bin]]
path = "./Queue.rs"
name = "queue"

[[bin]]
path = "./Stack.rs"
name = "stack"
```

PeanutbutterWarrior:
Like I've said before, if the name of the built binary isn't _main_ then it won't be copied into the build directory by SConscript. Cargo can't build both files because into _main_, so I think that we should leave it to be compiled with rustc.

SCons can't do multiple files from the same chapter yet, which will need fixing for #950 as you pointed out there. I think it might be best to hold off merging this PR (and #950) until we work that out.
We could also add a special case SConscript file here to deal with this, but as it's probably going to crop up again elsewhere we should make it work by default

ShadowMitia:
Right I forgot about that. #950 is for interpreted languages, it doesn't really make sense to modify Rust stuff there. It's best to find a solution separately (unless it's convenient all at once I guess).
I'll approve this PR then, @Amaras what do you suggest we do regarding the scons stuff?

ShadowMitia (approved):

Amaras:
> I'll approve this PR then, @Amaras what do you suggest we do regarding the scons stuff?

Since both files have a main function, the smartest move in my opinion would be to put a SConscript with two distinct rustc targets (say `build/rust/stack` and `build/rust/queue`).

We'll merge once this is done, since there already is an approval.

PeanutbutterWarrior:
I've had to rebase in order to get the latest SCons updates, as this branch was created before most of it. No conflicts had to be resolved. Git was confused, so I had to force push. I've added the SConscript file and checked that it builds

Amaras (approved):
Alright, the SCons part looks good to me.

We'll just need to merge the master branch correctly and it'll be good.﻿

leios:
Is there anything holding this PR back? I think it's mergable, but am not sure why we waited before

PR #926, opened 2021-11-08
adding a nojekyll file

leios:
This is an attempt to fix some of the build errors on the gh-pages branch. In this case, it seems like github pages is assuming a jekyll back-end, when we are using honkit. [These guys seem to have had a similar problem].(https://stackoverflow.com/a/41224377)


ntindle (approved):
Willing to give it a shot

PR #928, opened 2021-11-09
Force SCons to use GCC when building .c files

PeanutbutterWarrior:
This fixes SCons on Windows when MSVC is installed.  Previously it would attempt to use MSVC, but fail due to trying to pass it gcc libraries to link.

This needs testing on Linux before it is merged.

It may be better to have the sconstruct file detect which compiler is being used and adjust the calling based on that, so users with only MSVC or with clang can also build. Fixing it for now would still seem prudent, and we can change the behaviour later if wanted.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -15,7 +15,7 @@ env = Environment(ENV={'PATH': os.environ['PATH']})
 env['CC'] = 'gcc'
 for tool in ['gcc','gnulink']:
    env.Tool(tool)
-env['CCFLAGS']=''
+env['CCFLAGS'] = ''
 
 # Add other languages here when you want to add language targets
 languages = ['c']
--- a/SConstruct
+++ b/SConstruct
@@ -10,12 +10,7 @@ To run the compilation for all implmeentations in one language, e.g. Rust, run t
 from pathlib import Path
 import os
 
-env = Environment(ENV={'PATH': os.environ['PATH']})
-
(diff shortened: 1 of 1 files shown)


Amaras:
I can confirm it runs as previously on my Ubuntu Linux machine. I will review the changes more fully later

Amaras:
Alright, so the changes are quite simple, but I feel they are a bit too hard coded, especially when we're gonna add many more languages.
I'm still okay with this though.

PeanutbutterWarrior:
The most recent changes don't work, SCons has the same behaviour as before my commit.

SConstruct
```diff
@@ -8,8 +8,14 @@ Currently, the aim is to provide a way to compile or copy the implementation fil
...
 from pathlib import Path
+import os
 
-env = Environment()
+env = Environment(ENV={'PATH': os.environ['PATH']})
+
+env['CC'] = 'gcc'
+for tool in ['gcc','gnulink']:
+   env.Tool(tool)
+env['CCFLAGS']=''
```
Amaras, line 18:
```suggestion
env['CCFLAGS'] = ''
```

Amaras (approved):
Alright, it looks good to me. I think it's okay if we require GCC to compile.
The suggestion is basically just a nitpick, but it would bother me slightly less if you committed the change

PR #930, opened 2021-11-10
Fix devcontainer for Kotlin

ShadowMitia:
Fix devcontainer for Kotlin. Part of #871.


ntindle:
LGTM. Run kotlin with `kotlinc filename.kt -include-runtime -d filename.jar && java -jar filename.jar`

ntindle (approved):

PR #931, opened 2021-11-10
Improve SCons to use default files

PeanutbutterWarrior:
I've reworked SCons to have a default behaviour for languages defined by one file, and allow individual languages in chapters to provide a different behaviour if necessary.
This removes the necessity for an SConscript file for every language for every chapter.

New languages can be added by adding their name and file extension to the `languages` dictionary in _SConstruct_, adding a builder as before, and creating a _LANGUAGE_SConscript_ file in the _sconscripts_ directory. This file will be the default behaviour for files of that language. A list of the files that need to be compiled can be `Import`ed as `files_to_compile`. An example of this is the `c_SConscript` file.
If an SConscript file is found in the same directory as a code file that would otherwise be compiled it will instead run the SConscript file, and not pass it to the default file for compiling. This lets individual files have dependecies, as seen in Approximate Counting

I've confirmed it works on Windows when the changes in #928 are applied.


SConstruct
```diff
@@ -18,9 +18,34 @@ for tool in ['gcc','gnulink']:
...
+    if (code_dir := (chapter_dir / 'code')).exists():
+        for path in code_dir.iterdir():
+            if path.stem in languages:
+                # Check for overriding sconscript
+                if (sconscript_path := path / 'SConscript').exists():
+                    sconscripts.append(sconscript_path)
+                    SConscript(sconscript_path, exports='env')
+                else:
+                    files_to_compile[path.stem].extend(
+                                                       path.glob(f'*.{languages[path.stem]}')
```
Amaras, line 41: I'm bothered by this massive overhang. Could you just do a +4 spaces indent?

SConstruct
```diff
@@ -5,7 +5,7 @@ This provides Builder objects for each of the language implementations in the AA
 
 Currently, the aim is to provide a way to compile or copy the implementation files to the build directory, as well as to provide ways to run them and capture their output.
 
-To run the compilation for all implmeentations in one language, e.g. Rust, run the command `scons build/c`, and the resulting executables will be available in the `cuild/c` directory, each in their respective algorithm directory, containing the executable."""
+To run the compilation for all implmentations in one language, e.g. C, run the command `scons build/c`, and the resulting executables will be available in the `build/c` directory, each in their respective algorithm directory, containing the executable."""
```
Amaras, line 8: Thanks for correcting the typos

SConstruct
```diff
@@ -18,9 +18,34 @@ for tool in ['gcc','gnulink']:
...
+# Put 'name_of_language_directory' : 'file_extension'
+languages = {'c': 'c'}
 
 env.C = env.Program
 
-SConscript('SConscript', exports='env languages')
+Export('env')
+
+sconscripts = []
+files_to_compile = {language: [] for language in languages.keys()}
```
Amaras, line 29:
```suggestion
files_to_compile = {language: [] for language in languages}
```
This should also do the trick

Amaras (commented):
Alright, I'm nearly ready to merge this in.
You'll just need to merge the current main's SConstruct file, and I think I'll approve and merge this PR in.

SConstruct
PeanutbutterWarrior: 4 spaces indent was confusing about whether it was a statement or part of the previous one. Removing the extra line was the best alternative.

Amaras (approved):
Alright, it's pretty much good for me, just the type I'll commit right away.
I'll merge it directly once the checks succeed after that commit.

PR #932, opened 2021-11-11
Fix warnings and remove VLA on C++ Thomas algorithm

ShadowMitia:
This PR fixes warnings generated by the code, and removes a variable length array which is not legal in C++ (it's a compiler extension, which most of them do have, but is not strictly speaking C++).

I've also `const`ified and added references where I thought it would be best applicable.

The only part I'm not sure of is :

    for (int i = static_cast<int>(size) - 2; i >= 0; --i) {

and my use of the `static_cast`. size is a `std::size_t` and we can't just have `i` be one too (or any `unsigned`) because of overflow issues which would break the condition. I'm not sure what's the best thing to do in this scenario.
My guess is it's generally fine, and you should only be worried on very large arrays. Which is not the point here.


contents/thomas_algorithm/code/cpp/thomas.cpp
```diff
@@ -1,43 +1,45 @@
...
+  const std::vector<double> b = {1.0, 3.0, 6.0};
+  const std::vector<double> c = {4.0, 5.0, 0.0};
+  std::vector<double> x = {7.0, 5.0, 3.0};
 
-    std::cout << "The system" << std::endl; 
-    std::cout << "[1.0  4.0  0.0][x] = [7.0]" << std::endl;
-    std::cout << "[2.0  3.0  5.0][y] = [5.0]" << std::endl;
-    std::cout << "[0.0  3.0  6.0][z] = [3.0]" << std::endl;
-    std::cout << "has the solution" << std::endl;
+  std::cout << "The system\n";
```
Amaras, line 32: I am not sure changing `std::endl` to `\n` is necessarily a good practice, since `std::endl` also flushes the buffer.
Some platforms flush stdout when encountering a `\n`, but it's not the case in standard C++

ShadowMitia: I believe it's best practice, and if not, there is no reason to flush so much.
Flushing is a costly operation, and shouldn't be used all the time. 

Granted, in this scenario the cost is minimal, but encouraging `std::endl` can lead to performance penalties for no reason.

If you haven't seen it, this video covers the subject quite well I think : https://youtu.be/GMqQOEZYVJQ

Amaras: Okay, fair enough.

leios (approved):
This looks good to me. Happy to merge it in!

PR #934, opened 2021-11-11
Fixing C warnings

Amaras:
Currently, no warnings are activated during compilation for C.
Because of this, the code might be inconsistent in certain places.
I aim to fix or remove some errors in the C code, depending on if we consistently violate warnings or not.

Currently, only the following algorithms build without warnings:
 - approximate_counting
 - computus
 - forward_euler_method
 - graham_scan
 - jarvis_march
 - monte_carlo_integration
 - thomas_algorithm
 - tree_traversal
 - verlet_integration

The main errors I see in the rest of them are `sign-compare`, `unused-parameter` and `unused-variable`, along with a `implicit-function-declaration` and two `return-type`.
I did not implement any changes to the code yet, but I'll do that in the following days (hence why it's a draft PR).


Amaras:
It seems that, actually, I have done immediately exactly what I have said I would not do.

Anyway, I did 1 commit per algorithm fix, so that reverting is easier.

I can tell there is probably something wrong on the Gaussian elimination algorithm because I could completely eliminate the `rows` parameter in one function... This kind of fix was to brutally remove unused variables and parameters and adjust callers.

The rest of the fixes are simply fiddling around between `size_t` and `int` until the compiler stops throwing `signed-compare` errors.

ntindle:
Does this mean that Gaussian is wrong as a whole?

Amaras:
> Does this mean that Gaussian is wrong as a whole?

I have not checked the correctness of the code yet, so I can't answer properly. It's just that, intuitively, there's something wrong if you're able to not pass one of the dimensions of the matrix.

Amaras:
Okay, I think I know the basic assumption of the Gauss-Jordan algorithm's implementation. We assume that we have an n*n matrix, and we augment it with the vector, which means that we have an n*(n+1) matrix. As such, we technically don't need the `rows` parameter, since it's just `cols - 1`

contents/gaussian_elimination/code/c/gaussian_elimination.c
```diff
@@ -50,7 +50,7 @@ void gaussian_elimination(double *a, const size_t rows, const size_t cols) {
 void back_substitution(const double *a, double *x, const size_t rows,
                        const size_t cols) {
 
-    for (int i = rows - 1; i >= 0; --i) {
```
stormofice, line 53: This needs to be `i >= 0`, otherwise the loop body will never be executed

stormofice (changes requested):
Hi, thanks for the fixes!

Overall the changes look good to me, however there is a small typo in the gaussian elimination code.

Apart from that, it is ready to be merged.

contents/gaussian_elimination/code/c/gaussian_elimination.c
```diff
@@ -50,7 +50,7 @@ void gaussian_elimination(double *a, const size_t rows, const size_t cols) {
...
+    for (size_t i = rows - 1; i == 0; --i) {
```
stormofice, line 53: This needs to be i >= 0, otherwise the loop body will never be executed

stormofice (approved):

PR #936, opened 2021-11-14
Stack queue in cpp

ishwar00:


contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,100 @@
+#include<iostream>
+using namespace std;
```
ShadowMitia, line 2: It is not recommend to use `using namespace std;`, thought I think it's less of an issue in `.cpp` files.
See here : https://isocpp.org/wiki/faq/coding-standards#using-namespace-std

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,100 @@
...
+     * [value][next] -> [value][next] -> ... -> [value][next] -> [value][next]
+     *  (front Node)   (intermediat Nodes)    (rear Node)      (dummy Node)
+     */
+    template<typename T>
+    struct Node {
+        /**
+        * next: will store right Node address
+        */
+        T value;
+        Node<T>* next;
```
ShadowMitia, line 16: I think we should be using smart pointers for this.

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    template<typename T>
+    class queue {
+    private:
+        /**
+         * _front :  points to left most node
+         * count: keeps track of current number of elements present in queue excluding dummy Node
+         * rear:  points to most recent Node added into the queue, which is just left size of dummy Node
+         */
+        Node<T>* _front;
```
ShadowMitia, line 28: I would try and avoid the use of `_`, it's not wrong, but there are cases where it's easy for it to become undefined behaviour that it would best to encourage other ways of renaming when there are conflicts with member names.

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,100 @@
...
+        Node<T>* rear;
+        size_t count;
+    public:
+        queue() : _front(nullptr), rear(nullptr), count(0ULL) {
+            _front = rear = new Node<T>(0); // creating a dummy Node
```
ShadowMitia, line 33: You don't need a dummy Node if you have a `rear` pointer. 
This also causes bugs in your code, you can have an empty `queue` but using `front()` on it will give you a value, which is wrong. And that value can be modified and then lost when other values are pushed.

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,100 @@
...
+            if (count > 0) {
+                Node<T>* buffer = _front;
+                _front = _front->next;
+                count = count - 1;
+                delete buffer;
+            }
+            rear = (count != 0) ? rear : _front;
+        }
+
+        T& front() const { return _front->value; }
```
ShadowMitia, line 59: I would also add the version with `T const& front() const`
This allows to use `front()` in cases where a `const`-reference is needed.

```suggestion
        T& front() const { return _front->value; }
        T const& front() const { return _front->value; }
```

contents/stacks_and_queues/code/c++/stack.cpp
```diff
@@ -0,0 +1,80 @@
+#include<iostream>
```
ShadowMitia, line 1: I won't repeat it here, but a lot of the comments in `queue` apply here too.

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,100 @@
...
+        // returning reference can very usefull if someone wants to modify _front element
```
ShadowMitia, line 60:
```suggestion
```
I think this comment isn't necessary, a non-`const` reference can obviously be used to modify.

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+int main() {
+    my::queue<int> Q;
+
+    Q.push(0);
+    Q.push(1);
+    Q.push(2);
+    Q.push(3);
+    cout << "count: " << Q.size() << endl;
```
ShadowMitia, line 83: Replace `endl` with `\n`. `endl` does a newline and flush, which is not performant (and not required for everyline).

ShadowMitia (changes requested):
A good start, a few bugs to fix and a few things to iron out and you should be good.

contents/stacks_and_queues/code/c++/queue.cpp
ishwar00: thank you, i will use fully qualified names.

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,94 @@
...
+         * count: keeps track of current number of elements present in queue
+         * rear_pointer:  points to most recent Node added into the queue, which is right most Node
+         */
+        std::shared_ptr<Node<T>> front_pointer;
+        std::shared_ptr<Node<T>> rear_pointer;
+        size_t count;
+    public:
+        queue() : count(0ULL) { }
+
+        void push(const T& element) {
```
henrikac, line 36: I believe the chapter and the other examples uses `enqueue` instead of `push`.

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,94 @@
...
+    Q.push(1);
+    Q.push(2);
+    Q.push(3);
+    std::cout << "count: " << Q.size() << my::endl;
+    Q.front() = 0;
+
+    while (Q.empty() != true) {
+        std::cout << "element: " << Q.front() << my::endl;
+        Q.dequeue();
+    }
```
henrikac, line 92: Not sure if this is important but try to make this similar to the already implemented examples

contents/stacks_and_queues/code/c++/queue.cpp
```diff
@@ -0,0 +1,94 @@
+#include<iostream>
+#include<memory>
+#include<cassert>
+
+namespace my {
+    const char endl = '\n';
```
henrikac, line 6: Why this? why not just use `std::endl`?

ShadowMitia: I agree `'\n` should be inlined.
`std::endl` adds a newline and does a flush, unless you really need realtime output, it's very inefficient to add it to every line.

contents/stacks_and_queues/code/c++/queue.cpp
ishwar00: [https://www.algorithm-archive.org/contents/stacks_and_queues/stacks_and_queues.html](url)
they use `push` here, so i just followed along.

ShadowMitia: It does for stack but not for queue.

ishwar00: for example, will often use `dequeue()` instead of `pop()` and `front()` instead of `top()`.

I read above lines in explanation, so i thought they kept `enqueue` as `push`, my bad.

contents/stacks_and_queues/code/c++/queue.cpp
ishwar00: Yeah, i should have made them similar.

ShadowMitia (approved):
Looks good to me. Thank you!

PR #938, opened 2021-11-17
feat: add publish dockerfile

ntindle:
publish the dockerfile to aaa-langs
move the Dockerfile to the root of the repo


ntindle:
use the below command to test after logging in with [these instructions](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry)

```bash
docker run -v $(pwd):/tmp -it ghcr.io/algorithm-archivists/aaa-langs:feature-publish-container bash
```

ntindle:
@ShadowMitia can you take a look?

ntindle:
Can @algorithm-archivists/development give this an approval?

.github/workflows/publish_container.yml
```diff
@@ -0,0 +1,14 @@
+name: Publish Docker
+on: [push]
```
ShadowMitia, line 2: I've seen other examples of github actions where they restrict branches this is run on. I'm thinking maybe restrict to code pushed to master?
```suggestion
on:
  push:
    branches:
      - master
```

ShadowMitia (approved):
Looks good to me.

I just have a suggestion for restricting building the image on master branch.

While we're at it though, maybe now would be a good time to change to `main` before we do any changes that depend on the branch name.

.github/workflows/publish_container.yml
ntindle: we can do that, I had to leave it out to test it

ShadowMitia (approved):
Looks good to me. I'm still suggesting we do the full change to `main` before, but otherwise I think we're good.

leios (approved):
I also think it's fine and am happy to approve it

ntindle:
`docker run -v $(pwd):/tmp -it ghcr.io/algorithm-archivists/aaa-langs:latest bash`

PR #939, opened 2021-11-22
Added C++ compilation

Amaras:
I added the compilation infrastructure for C++.

I had to add the `-std=c++17` option because the Barnsley fern code would not compile without it.

Apart from that, everything compiles successfully in the right folder on my machine :smiley:


ShadowMitia (commented):
I think this looks fine. I'm wondering, maybe we should already add a few more flags? Like -pedantic -Wextra -Wall -Werror?
I'm not saying being exhaustive, but already have a minimal setup for future contributions.

Amaras:
> I think this looks fine. I'm wondering, maybe we should already add a few more flags? Like -pedantic -Wextra -Wall -Werror?

That's something I want to have, but we need to fix warnings before we can do that.

leios (approved):
This looks fine to me, happy to approve!

ShadowMitia:
> > I'm wondering, maybe we should already add a few more flags? Like -pedantic -Wextra -Wall -Werror?
> 
> That's something I want to have, but we need to fix warnings before we can do that.

That makes sense

Amaras:
Okay, after careful consideration, I think we should have this merged as soon as possible, so we can move on.
I really want to come back to fixing the warnings, but we need to keep going while there is a bit of momentum left.

In short, I'm merging this in.

PR #940, opened 2021-11-22
Added x64 assembly compilation

Amaras:
And here we go again with X64 Assembly.

There was a little annoyance this time: the way we programmed the assembly required the compilation to use the link flag `-no-pie`. As such, to prevent environment duplication, I elected to add the link flag to all the executables.

Apart from this, compilation was not affected, and all executables compile successfully on my Linux machine.

We really should set up a Windows environment to check whether it works as intended there, though.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -11,9 +11,10 @@ from pathlib import Path
 import os
 
 env = Environment(ENV={'PATH': os.environ['PATH']},
-		  tools=['gcc', 'gnulink', 'g++'])
+		  tools=['gcc', 'gnulink', 'g++', 'gas'])
 
-env['AS'] = 'as'
+env['ASFLAGS'] = '--64'
+
 env['CCFLAGS'] = ''
 env['CXXFLAGS'] = '-std=c++17'
 
--- a/SConstruct
+++ b/SConstruct
@@ -13,18 +13,20 @@ import os
 env = Environment(ENV={'PATH': os.environ['PATH']})
 
 env['CC'] = 'gcc'
+env['AS'] = 'as'
 for tool in ['gcc','gnulink']:
    env.Tool(tool)
 env['CCFLAGS'] = ''
 env['CXXFLAGS'] = '-std=c++17'
+env['LINKFLAGS'] = '-no-pie'
 
 # Add other languages here when you want to add language targets
 # Put 'name_of_language_directory' : 'file_extension'
-languages = {'c': 'c', 'cpp': 'cpp'}
+languages = {'c': 'c', 'cpp': 'cpp', 'asm-x64': 's'}
 
 env.C = env.Program
 env.CPlusPlus = env.Program
-
+env.X64 = env.Program
 
 Export('env')
 
@@ -46,7 +48,8 @@ sconscript_dir_path = Path('sconscripts')
 for language, files in files_to_compile.items():
     if files:
         if (sconscript_path := sconscript_dir_path / f"{language}_SConscript").exists():
-            SConscript(sconscript_path, exports = {'files_to_compile': files})
+            SConscript(sconscript_path, exports = {'files_to_compile': files,
+                                                   'language': language})
         else:
             print(f'{language} file found at {files[0]}, but no sconscript file is present ')
 
--- a/SConstruct
+++ b/SConstruct
@@ -10,12 +10,10 @@ To run the compilation for all implementations in one language, e.g. C, run the 
 from pathlib import Path
 import os
 
-env = Environment(ENV={'PATH': os.environ['PATH']})
+env = Environment(ENV={'PATH': os.environ['PATH']},
+		  tools=['gcc', 'gnulink', 'g++'])
 
-env['CC'] = 'gcc'
 env['AS'] = 'as'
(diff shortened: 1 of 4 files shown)


SConstruct
```diff
@@ -13,18 +13,20 @@ import os
 env = Environment(ENV={'PATH': os.environ['PATH']})
 
 env['CC'] = 'gcc'
+env['AS'] = 'as'
 for tool in ['gcc','gnulink']:
    env.Tool(tool)
 env['CCFLAGS'] = ''
 env['CXXFLAGS'] = '-std=c++17'
+env['LINKFLAGS'] = '-no-pie'
```
ShadowMitia, line 21: I don't know that flag, but it gets applied to all languages, which might cause issues?
For C/C++ that's maybe fine (though maybe not what we want).

ShadowMitia:
I have executables that segfault on my machine, the ones that do run:
- cooley_tukey
- monte_carlo_integration
- verlet_integration

SConstruct
Amaras: This could be a problem indeed for security reasons in C and C++.
This was done on the main environment to prevent environment duplication. Otherwise I would have just cloned it.

Amaras:
The only segfault I could reproduce on my machine is for `forward_euler_method`, which does so in `__printf`, probably because of a bad formatting in `results_loop`.

I will need more detail for the others, and is probably not my responsibility. I had already noticed the segfault in some of the x64 implementations, but did not take the time to report them, I think.

contents/cooley_tukey/code/asm-x64/SConscript
```diff
@@ -0,0 +1,6 @@
+Import('*')
+from pathlib import Path
+
+dirname = Path.cwd().parents[1].stem
+
+env.X64(f'#/build/asm-x64/{dirname}', Glob('*.s'), LIBS=['m'])
```
ShadowMitia, line 6:
```suggestion
env.X64(f'#/build/asm-x64/{dirname}', Glob('*.s'), LIBS=['m'], parse_flags="-Wl,-no-pie")
```

contents/forward_euler_method/code/asm-x64/SConscript
```diff
@@ -0,0 +1,6 @@
+Import('*')
+from pathlib import Path
+
+dirname = Path.cwd().parents[1].stem
+
+env.X64(f'#/build/asm-x64/{dirname}', Glob('*.s'), LIBS='m')
```
ShadowMitia, line 6:
```suggestion
env.X64(f'#/build/asm-x64/{dirname}', Glob('*.s'), LIBS='m', parse_flags="-Wl,-no-pie")
```

sconscripts/asm-x64_SConscript
```diff
@@ -0,0 +1,6 @@
+Import('files_to_compile language env')
+from pathlib import Path
+
+for file in files_to_compile:
+    chapter_name = file.parent.parent.parent.stem
+    env.X64(f'#/build/{language}/{chapter_name}', str(file))
```
ShadowMitia, line 6:
```suggestion
    env.X64(f'#/build/{language}/{chapter_name}', str(file), parse_flags="-Wl,-no-pie")
```

ShadowMitia (commented):
I might've found a way to pass it only to specific languages. 
I found the information here: https://scons.org/doc/production/HTML/scons-man.html#f-ParseFlags, basically you can use `-Wl,` to ass stuff to the linker.
It might not be very cross-platform though... Should support gcc/clang at least.

ShadowMitia:
> 

I've run some quick tests, it does seem to also be coming from `printf` formatting as well on my end, but it happens on more examples. I'll open an issue for it.

ShadowMitia:
> The only segfault I could reproduce on my machine is for `forward_euler_method`, which does so in `__printf`, probably because of a bad formatting in `results_loop`.
> 
> I will need more detail for the others, and it is probably not my responsibility. I had already noticed the segfault in some of the x64 implementations, but did not take the time to report them, I think.

I've run some quick tests, it does seem to also be coming from printf formatting as well on my end, but it happens on more examples. I'll open an issue for it.

Amaras:
> It might not be very cross-platform though... Should support gcc/clang at least.

We basically don't matter if it's cross platform at this point, do we? We already forced the use of gcc, gnulink, next will be g++ and the GNU X86_64 assembler.

ShadowMitia:
For now I think you're right

Amaras:
Alright, @PeanutbutterWarrior and I decided on Discord to give up on trying to build x86-64 assembly on Windows.
The linker seemed to outright ignore the `-no-pie` option, as well as `--no-pie`, `-static` and `--static`.

Commit 12350c5 fixed a bug in @ShadowMitia's proposal that made it impossible to link correctly. However, the expected behaviour was achieved by using the `LINKFLAGS` argument in the `env.X64` calls. (The commit also simplifies the SConstruct so that's always a plus).

As such, I think this PR is ready to merge, just possibly waiting for the approval of another maintainer.

leios (approved):
This looks great! Thanks for the work here!

PR #943, opened 2021-11-27
Fix warning on MSVC for barnsley C++ code

ShadowMitia:


stormofice (approved):
LGTM, tested on MSVC v19.29.

Thanks for the fix!

Issue #944, opened 2021-11-28
Split operator method and Cooley-Tukey in Rust don't compile.


PeanutbutterWarrior:
# Bug Report

## Description

Both split operator method and Cooley-Tukey don't compile in Rust. Only slight changes are needed as far as I can tell.

### Steps to Reproduce

Neither file has a Cargo.toml file which is required for dependencies. After creating a Cargo.toml for each with `num = "0.4.0"`, `rustfft = "5.1.1"`, and `rand = "0.8.4"` and running `cargo build` they fail with the errors in the screenshots

### Expected behavior

They should compile without errors.

## Screenshots

Cooley-tukey errors:
![image](https://user-images.githubusercontent.com/50717143/143773117-c3ae244a-27ef-4ed4-aa6a-1c27bd1c174a.png)

Split-Operator method errors:
![image](https://user-images.githubusercontent.com/50717143/143773131-38fea7ce-c195-4edd-87ad-97f11b8a168c.png)

## Environment

 - OS: Windows
 - OS Version: Windows 10 Home versiom 20H2 build 19042.1348
 - Cargo version: cargo 1.55.0 (32da73ab1 2021-08-23)

## For Algorithm Archive Developers

- [ ] The bug can be reproduced
- [ ] The bug can be fixed (if not, please explain why not in a comment below)
- [ ] There is a timeline to fix the bug
- [ ] The bug has been fixed (Please link the PR)

Amaras:
What happens if you replace `rustfft::FFTPlanner` with `rustfft:FftPlanner`? It won't fix the Cooley-Tukey method, but that could help for the Split-op method

PeanutbutterWarrior:
Solved by #945 . A previous version of the library was needed

PR #945, opened 2021-11-28
Add Rust to SCons

PeanutbutterWarrior:
Add Rust to the SCons build system. If a _Cargo.toml_ file is present then it builds with Cargo, otherwise it will build with rustc. As such a _Cargo.toml_ file is necessary if any external crates are necessary (usally an `extern crate ...;` in the source file). This could be replaced with a custom `Scanner`, but that's probably overkill and error prone.

In order to have all the code files get picked up by SCons, I have the _*.rs_ and the _Cargo.toml_ file in the same directory, and define the path to the source file in _Cargo.toml_. An example of this is the Cooley-Tukey code.

I have the _*.pdb_ files and the _target/_ directories marked by SCons for cleaning when using the `-c` argument. This deletes the _Cargo.lock_ files, causing cargo to recompile the external crates on the next build.

Due to Cargo putting the executable into a directory in the target directory, it is built in the same place as the source file and the executable then copied to the _build/_ directory. There is a command line argument to copy the build artifacts to a directory (--out-dir) but this is only available on nightly rust. When (and if) this gets stabilized then it would simplify the SConstruct file to take advantage of this.


PeanutbutterWarrior:
Cooley-Tukey and Split Operator Method don't compile, but this is a problem with those files rather than SCons. I've already 
opened an issue for it.

contents/monte_carlo_integration/code/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rand = "0.8.4"
+
+[[bin]]
+path = "./monte_carlo.rs"
+name = "main"
```
ShadowMitia, line 13: I would use the file name to make things clearer.

```suggestion
name = "monte_carlo"
```

contents/split-operator_method/code/rust/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+num = "0.4.0"
+rustfft = "5.1.1"
+
+[[bin]]
+path = "./split_op.rs"
+name = "main"
```
ShadowMitia, line 14:
```suggestion
name = "split_op"
```

contents/huffman_encoding/code/rust/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+itertools = "0.10.1"
+
+[[bin]]
+path = "./huffman.rs"
+name = "main"
```
ShadowMitia, line 13:
```suggestion
name = "huffman"
```

contents/cooley_tukey/code/rust/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rand = "0.8.4"
+rustfft = "6.0.1"
+
+[[bin]]
+path = "./fft.rs"
+name = "main"
```
ShadowMitia, line 14:
```suggestion
name = "fft"
```

SConstruct
```diff
@@ -10,13 +10,22 @@ To run the compilation for all implementations in one language, e.g. C, run the
...
+
+rust_rustc_builder = Builder(action='rustc $SOURCE -o $TARGET$PROGSUFFIX')
+
+env = Environment(ENV={'PATH': os.environ['PATH']},
+                  BUILDERS={'rustc': rust_rustc_builder, 'cargo': rust_cargo_builder},
+                  tools=['gcc', 'gnulink', 'g++', 'gas'])
 
 env['CCFLAGS'] = ''
 env['CXXFLAGS'] = '-std=c++17'
+env['ASFLAGS'] = '--64'
```
ShadowMitia, line 28: I don't know `scons` why is this needed for Rust?

contents/barnsley/code/rust/Cargo.toml
```diff
@@ -1,9 +1,13 @@
...
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-rand = "0.8.4"
\ No newline at end of file
+rand = "0.8.4"
+
+[[bin]]
+path = "./barnsley.rs"
+name = "main"
```
ShadowMitia, line 13: I would use the file name to make things clearer.

```suggestion
name = "barnsley"
```

ShadowMitia (commented):
I wonder if we shouldn't keep the `src/` to not diverge too much from default `Cargo.toml` settings. We still want to show good practices with Rust.

I see there are builders for both `rustc` and `cargo`, which is fine. But maybe it'll be simplest if we just keep the `cargo` one?

SConstruct
PeanutbutterWarrior: That's part of the assembly build. It might have got mixed up in a rebase

contents/monte_carlo_integration/code/rust/Cargo.toml
PeanutbutterWarrior: The name determines what the name of the executable is. If it's not constant then it will need to be passed as another arguement to the build, and as it gets renamed automatically when it's moved I thought this would be simpler.

ShadowMitia:
Dockerfile install looks fine to me.

Amaras (approved):
Alright, I tested the Rust compilation with your docker file, and it works (except the two you pointed out in #944, of course).

I'm happy to merge this in when you are ready.﻿

PeanutbutterWarrior:
I've fixed the issues with some programs not compiling due to the wrong library versions being used. Solves #944 . It could do with updating, but that's another issue.

Amaras (approved):
Can confirm it works in the docker container.

Happy to merge this :)﻿

PR #947, opened 2021-11-28
Add Go to SCons

PeanutbutterWarrior:
Adds Go to SCons.

Probably should be checked on docker, but it's probably too simple to fail.


Amaras (approved):
Looks good to me.

I think in the future, we should think about separating the builders from the SConstruct.

However, it is not necessary for now. ﻿

PR #948, opened 2021-11-28
Added Fortran compilation

Amaras:
And here we go with the next language: Fortran.

Adding it was straightforward since there was no library to add.

The only problem was the double definition of the `build/fortran/euclidean_algorithm` target, which I had to create a SConscript for.﻿


ShadowMitia (approved):
@PeanutbutterWarrior approval

ntindle (approved):
Approving so a maintainer approves

PR #949, opened 2021-11-28
Added Coconut compilation

Amaras:


ShadowMitia (approved):
Looks good to me, and tested on the latest Docker and works fine.
The only thing I would discuss is maybe using a more recent version of python? 
3.10 is already out. I tested with 3.9 it seems to work fine. Is there a reason to use 3.8?
I'm not sure what's the best practice is with coconut.

Amaras:
> The only thing I would discuss is maybe using a more recent version of python? 3.10 is already out. I tested with 3.9 it seems to work fine. Is there a reason to use 3.8? I'm not sure what's the best practice is with coconut.

Since we target for 3.8+, it will work for 3.9 and 3.10 no problem.
The best practice would be to have universal code. However, since I sometimes use Python 3.8 syntax, compilation for (at least) 3.8 is needed.
If we get to it, we can bump the target version to 3.9 or 3.10.
We probably also need to have the `--strict` option, but I think some don't compile with this flag, so this should be a separate PR.

PR #950, opened 2021-11-29
Add interpreted languages to SCons

PeanutbutterWarrior:
Adds most of the interpreted languages to SCons. Their source files are simply copied to the build directory. The languages added are:
- Bash
- Julia
- Lolcode
- Lua
- PHP
- Powershell
- Python
- Ruby
- Smalltalk
- Viml

Viml might be a typo and supposed to be vim.
I've left matlab out as it currently isn't able to run in the Docker container and I don't know how it runs.
Scratch doesn't seem to have consistent files, as such I have left it out for now. It's common enough (6 uses) that we should work out how we are managing this.
Piet is made up of .png files and are in a directory so won't be noticed by SCons. Only euclidean formula has been written (drawn?) in Piet, so this isn't major.
Whitespace is also in directories and won't be noticed by SCons, but once again is only for euclidean formula.

I haven't checked that interpreters for these are in the Docker image, although I think that most of them are.


PeanutbutterWarrior:
If this is merged it brings us to 15/45 languages building.

Amaras:
Viml (or rather VimL) is the Vim scripting language.

I'll review this more carefully later

ShadowMitia (commented):
I have an issue with renaming the files to the chapter name, as some projects have multiples files. Copying the files and keeping their original names should be sufficient.
We could also have a SConscript for those specific ones, but I feel it's not necessary since it's just copying files about.

ShadowMitia:
Note: you can just grab the filenames like this:
```python
env.Copier(f'#/build/lang/{file.name}', str(file))
```

PeanutbutterWarrior:
You're probably right, I'll fix that. I also realize that I've missed Javascript, so I'll add that too.

leios:
How does this work with #963? Should we merge one or the other first?

PeanutbutterWarrior:
This one isn't updated yet, and won't work after #963 gets merged. If we can sort out and merge #963 first so I can change this one to fit the new system, and only then merge this one

ShadowMitia:
I might have an issue, it seems we're only copying the files over, right? So that will always succeed even if the code has issues, no?
Or am I missing something? ^^

Amaras:
If you find a way to compile interpreted files, you are welcome to tell us that way.
The main problem is that we can't detect things like syntax or name errors.

However, it's basically the same thing as for the x86-64 implementation of the forward Euler method that segfaults... We can't do much more than copy files over (well, we *could* run them, but that's gonna add a lot of complexity and time)

SConstruct
```diff
@@ -33,13 +37,25 @@ env['ASFLAGS'] = '--64'
...
+    'javascript': 'js',
+    'julia': 'jl',
+    'lolcode': 'lol',
+    'lua': 'lua',
+    'php': 'php',
+    'powershell': 'ps1',
+    'python': 'py',
+    'ruby': 'rb',
+    'rust': 'rs',
+    'smalltalk': 'st',
```
Amaras, line 57: There are Smalltalk compilers. Do we really want to require an interpreter?

ShadowMitia:
Right I think I got mixed up with the code verification stuff. Thanks for the clear-up!

Amaras (approved):
Alright, it's all good for me now!
I'll let you deal with the conflicts in the SConstruct before merging

PR #955, opened 2021-11-29
Change the "Publish Container" Action to only run in the main repository

stormofice:
This prevents this action from failing in forked repositories


ntindle (approved):
LGTM if it runs

PR #961, opened 2021-12-01
IFS: Standardize output files

stormofice:
This PR aims to standardize the output file name and the amount of points generated.

The output file will now always be named `sierpinski.dat` and it will contain 10000 points.


ShadowMitia (changes requested):
For `IFS.c` I would share the size of the array, maybe in a variable, so that you can more easily change the size. And makes the code less hard coded.

ShadowMitia (approved):
Looks good to me!

Issue #954, opened 2021-11-29
Best way to output multiple executables for the same chapter and language


PeanutbutterWarrior:
### Current Behaviour

Currently SCons assumes that each language's implementation of each chapter has only a single file that needs building. Having multiple files that are compiled together into a single executable is a special case that justifies a SConstruct file to handle it specifically, but often there are two or more files that are compiled into separate executables, such as in Stacks and Queues. An minimal example of the current structure generated is
```
arcane-algorithm-archive/
|──build/
|   |──c/
|   |   +  barnsley.exe
|   |   +  euclidean_algorithm.exe
|   |
|   |──python/
|   |   +  barnsley.py
|   |   +  euclidean_algorithm.py
```
If a language has multiple implementations for a chapter, or implements parts of the chapter in separate files then the executable generated from one file will overwrite the other (SCons will actually detect this, throw an error and not build either which is arguably worse).

### Desired behaviour
We want all the files to be compiled into separate executables. In order to do this they need either different names or to be put into different folders. Multiple ways to do this have been suggested, described below along with pros and cons for them.

#### Option 1: Duplicate the structure of `arcane-algorithm-archive/contents`
This would be putting the executables into folders depending on their chapter, in folders based on language. The executables would be named the same as the source file (with the extension changed as necessary). The structure of this is below
```
arcane-algorithm-archive/
|──build/
|   |──barnsley/
|   |   |──python/
|   |   |   +  barnsley.py
|   |   |──c/
|   |   |   +  barnsley_1.exe
|   |   |   +  barnsley_2.exe
|   |
|   |──euclidean_algorithm/
|   |   |──python/
|   |   |   +  euclidean_algorithm.py
|   |   |──c/
|   |   |   +  euclidean_algorithm.exe
```
Pros:
- Easy to navigate
- Will not run have naming issues in future unless the structure of `arcane-algorithm-archive/contents` is changed
- Building specific chapters is easy with SCons

Cons:
- Deeply nested
- Redundant folder dulplication of the language folders
- Building specific languages is difficult with SCons (unless this is implemented specifically)

#### Option 2: Same as Option 1, but with languages being the outer folders
This has most of the same pros and cons as _Option 1_, however building specific languages is easy and specific chapters hard

#### Option 3: Suffix files with `_source_file_name`
This would have the same structure as currently, however the executable names would be `chapter_filename`. The folder structure this would generate is below
```
arcane-algorithm-archive/
|──build/
|   |──c/
[...]

PR #963, opened 2021-12-01
Change SCons output structure

PeanutbutterWarrior:
The PR for #954
Currently:
- Changes the output structure to the new one for all builds
- Adds aliases for languages to just use their name e.g. `scons rust`
- Searches chapters recursively for source files
- Adds aliases for chapters


ntindle:
This seems to work to me. Not one of the people who originally weighed in though

leios (approved):
I think this looks great! I would have loved to see this with "edge cases" like the code for the convolution chapter to make sure it works, but I think it is fine as is.

I'll wait a second to merge it in case there are more informed opinions

PeanutbutterWarrior:
I've tested nested chapters with dummy files and it works as far as I can tell. I've got no idea what will happen if there's a code directory inside another, so we should avoid that (although I don't know how that would happen). This also doesn't replicate the structure inside the individual language directories, so if that matters it will need an SConscript file.

PR #964, opened 2021-12-02
Implement Graham Scan in Rust

PeanutbutterWarrior:
This applies the changes requested from the abandoned PR #479


contents/graham_scan/code/rust/graham_scan.rs
```diff
@@ -0,0 +1,96 @@
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        // Neither field of Point will be NaN, so this is safe
+        self.partial_cmp(other).unwrap()
+    }
+}
+
+fn counter_clockwise(a: &Point, b: &Point, c: &Point) -> f64 {
+    (b.x - a.x) * (c.y - a.y) - (b.y - a.y) * (c.x - a.x)
+}
+
```
ShadowMitia, line 32: A quick comment like the other functions could be nice here

ShadowMitia (changes requested):

contents/graham_scan/code/rust/graham_scan.rs
```diff
@@ -0,0 +1,96 @@
...
+            } else if i == points.len() {
+                break;
+            } else {
+                i += 1;
+            }
+        }
+
+        m += 1;
+        points.swap(i, m);
+    }
```
ShadowMitia, line 68: This would be better rewritten with a `while` loop. It's confusing to have a mutable index inside a `for` loop (and it really feels like something that shouldn't be done, but I need to double check on that).

It could also be rewritten as a double `for` loop, which should be clearer, and I think does exactly the same. And you wouldn't need to have a `m` counter outside.

ShadowMitia (changes requested):

contents/graham_scan/code/rust/graham_scan.rs
ShadowMitia: Follow up: Modifying `i` is fine, I'm just not used to doing this :grin:

ShadowMitia (approved):
Looks good to me! Thanks!

PR #967, opened 2021-12-04
Removing probability folder

leios:
I was looking at the new probability distribution chapter and noticed the licensing image at the bottom was not correct. We can either change the md file so it points to the right image or change the directory structure.

I am not sure what the reasoning was to create a probability sub-directory to begin with. Maybe there was a chance of creating individual discrete and continuous PDF chapters?

No matter the case, people were confused before when we did not have a flat hierarchy, so I think renaming `probability/distributions` to `probability_distributions` makes the most sense.


leios:
Sorry, I forgot to tag @shudipto-amin

ntindle (approved):

PR #969, opened 2021-12-05
Tweaked SCons compilation action step

Amaras:
This is a PR to use parallel compilation ﻿during the build action.

This will hopefully make the compilation go faster (not that we curently need it, but it's still a nice addition for the future).


ShadowMitia (approved):
Looks good to me.

PR #971, opened 2021-12-05
Fixed Coconut compilation problem

Amaras:
The issue this PR fixes is the following:

When building for the first time, the Coconut compiler considers the build target a directory.
However, when running a second time, SCons is confused because it expected a file.

This can be fixed by adding the `.py` suffix to the build target (see the Coconut SConscript).
﻿


ShadowMitia (approved):
Approved!

PR #973, opened 2021-12-05
Added Java compilation 

### LINKED ISSUE #972: Java
This is an issue to ask how to address the compilation of Java code.
Do we want to have classes around, or do we want a .jar file?

Amaras:
This is the PR to close #972.

For now, I have implemented compilation with `.class` files.
We might want to use `.jar` files instead, but they require a manifest file.

It it also possible since Java 11 to simply run a single file, so a copier could be enough (assuming Java 11+).
﻿


leios (approved):
Gotta be honest, I don't know Java super well, but I trust this

Amaras:
Alright, I guess I will merge this PR, it has been open long enough for people to criticize, and no disagreement came of it.
If you have any qualms, please open a PR in response to correct this.
The GH Actions run correctly, so I don't think there will be a problem with it, though (the changed filenames were fixed)

PR #975, opened 2021-12-16
Dockerfile: set default to bash

ShadowMitia:
This sets a default to /bin/bash. The default being `sh` this should avoid issues for people running the docker with setting a shell.

This is easily overridden by passing the shell to the command as usual.


Amaras (commented):
If you're going to do this, you also probably could change the PATH to include the Rust toolchain?

ShadowMitia:
I'll double check but that only happens in the Github Action, no? When you launch the docker the rust toolchain is available (at least with bash). I'm just making sure it's not launched into `sh` by default.

Amaras (approved):
Ok, good for me, but you should probably remove the setup in the GH actions, then﻿

ntindle:
Does the GH Actions still work with this?

leios:
If the GH builds works here after updating it to main, it should be fine, right?

PR #979, opened 2021-12-27
Updated tree traversal chapter

ell-hol:
Made it more pythonic and compact. Usage of temp variable wasn't necessary nor was the loop through children or the appending of the root node to the stack.

### DIFF
--- a/contents/tree_traversal/code/python/tree_traversal.py
+++ b/contents/tree_traversal/code/python/tree_traversal.py
@@ -46,18 +46,12 @@ def dfs_recursive_inorder_btree(node):
 
 
 def dfs_stack(node):
-    stack = []
-    stack.append(node)
-
-    temp = None
-
-    while len(stack) > 0:
-        print(stack[-1].data, end=' ')
-        temp = stack.pop()
-
-        for child in temp.children:
-            stack.append(child)
-
+    stack = [node]
+    while stack:
+        node = stack.pop()
+        if node.children:
+            stack += node.children
+        print(node.data, end=' ')
 
 def bfs_queue(node):
     queue = []
--- a/contents/tree_traversal/code/python/tree_traversal.py
+++ b/contents/tree_traversal/code/python/tree_traversal.py
@@ -54,17 +54,12 @@ def dfs_stack(node):
         print(node.data, end=' ')
 
 def bfs_queue(node):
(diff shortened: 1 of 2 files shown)


contents/tree_traversal/code/python/tree_traversal.py
```diff
@@ -46,31 +46,20 @@ def dfs_recursive_inorder_btree(node):
...
-        temp = stack.pop()
-
-        for child in temp.children:
-            stack.append(child)
-
+    stack = [node]
+    while stack:
+        node = stack.pop()
+        if node.children:
+            stack += node.children
```
Amaras, line 53: Couldn't you do `stack.extend`, rather than using `__iadd__`?

contents/tree_traversal/code/python/tree_traversal.py
```diff
@@ -46,31 +46,20 @@ def dfs_recursive_inorder_btree(node):
...
-        print(queue[0].data, end=' ')
-        temp = queue.pop(0)
-
-        for child in temp.children:
-            queue.append(child)
+    queue = [node]
+    while queue:
+        node = queue.pop(0)
+        if node.children:
+            queue += node.children
```
Amaras, line 61: Same as above, maybe use `.extend`?

Amaras (commented):
Thank you for your first PR in the AAA.
I agree this chapter needed a bit of a better implementation, which you provided.
However, since you changed the number of lines in some functions, you will need to modify the line numbers in `contents/tree_traversal/tree_traversal.md`.

I'll wait for you to do that before approving.

Amaras (approved):
LGTM, I'll merge when the build action runs correctly :)

contents/tree_traversal/code/python/tree_traversal.py
ell-hol: Good idea, done.

contents/tree_traversal/code/python/tree_traversal.py
ell-hol: Good idea, done.

Amaras:
Alright, there seems to be a GH Actions problem, so we'll see when the incident is resolved, sorry for the delay.
There won't be a problem, so an admin might merge it anyway.

PR #980, opened 2021-12-27
Update bfs_queue

ell-hol:
Made it more pythonic and compact


Amaras:
Hi, thank you for your first PRs in the AAA.

Could you maybe have both the stack and queue approaches in #979, since this does not really make sense as its own PR?

Amaras:
Since this was put into #979, I'm going to close this

PR #981, opened 2022-01-01
add Go implementation of IFS

radu-munteanu:


radu-munteanu:
updated https://github.com/algorithm-archivists/algorithm-archive/pull/789

PR #982, opened 2022-01-01
add Go implementation of IFS

radu-munteanu:


radu-munteanu:
Ref https://github.com/algorithm-archivists/algorithm-archive/pull/789

radu-munteanu:
[lang: go]

Amaras:
Thanks for your PR, we will try to find a Go reviewer to review this as quickly as possible :)

Also, we paused the [lang: lang] GH action, so labeling is mostly manual again, sorry about that

contents/IFS/code/go/IFS.go
```diff
@@ -0,0 +1,82 @@
...
+type point struct {
+	x, y float64
+}
+
+func randPoint(points []point) point {
+	return points[rand.Int()%len(points)]
+}
+
+func chaosGame(n int, shapePoints []point) []point {
+	if n < 1 || len(shapePoints) == 0 {
```
henrikac, line 28: Since `shapePoints` is a `[]point` it is possible to pass `nil` to the function and I am not sure what would happen if you check `len(nil)`?

radu-munteanu: > Since `shapePoints` is a `[]point` it is possible to pass `nil` to the function and I am not sure what would happen if you check `len(nil)`?

It will work. https://play.golang.com/p/94ZTjMcLEHo

radu-munteanu:
Good luck with AAA.

Amaras (approved):
If it looks good to @henrikac, then it's good for me

Amaras:
Why was this closed before it was merged @radu-munteanu?

PR #984, opened 2022-01-11
Added Kotlin compilation

Amaras:
This PR is to add Kotlin compilation to SCons.

I have decided to compile to a JAR because compiling to individual classes was more anoying.

I could probably compile to a JAR including the Kotlin runtime, but this did not seem necessary.

The tool is taken nearly directly from the scons-contrib repository (under MIT license).

If you have any other remarks, feel free to put them below.﻿


leios (approved):
I don't know kotlin, but the build process seems fair

PR #987, opened 2022-01-18
Eliminating C/C++ implicit conversions

Amaras:
This PR is to remove all implicit conversions in both the C and C++ codes we have.

Each file change was put in its own commit for easier revertability.
Most changes were mainly following the compiler instructions, athough a few codes needed a change of types in their structure (looking at Huffman encoding in C).

I can confirm each code compiles correctly on my machine, *with* the Conversion error in place.

If you have trouble with compilation on Windows because of the change, please put it below and we'll see what we can do.
There should not be a problem on Linux, though.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -60,7 +60,7 @@ for language, tools in languages_to_import.items():
 
 Export('env')
 
-env['CCFLAGS'] = '-Wall -Wextra -Werror -pedantic'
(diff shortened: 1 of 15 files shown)


ShadowMitia (approved):
Looks good to me. All the code runs and seems to all be giving the expected answers.

PR #990, opened 2022-01-31
Implemented stack and queue in C#

hwat212:


fo1981 (approved):

Liikt:
What I don't like about this PR is that the dequeue operation doesn't return the dequeued object. Now that said neither does the cpp version. I would personally prefer what rust, java and ts do currently, because in most use cases you want to dequeue on object and then still work with it.

PR #998, opened 2022-04-14
adding more info on stormer verle

### LINKED ISSUE #994: Error Term in Verlet Integration Page
Hi team,

This is regarding the Verlet Integration page that's here: https://www.algorithm-archive.org/contents/verlet_integration/verlet_integration.html

Similar to how Equation 4 is sum of Eq. 2 and 3, we can see that Eq. 5 is a difference of 2 and 3. When we do that, the error term turns out to be O(del. t^3) and not O(del. t^2) as stated in the page. Did I get something wrong? Or should that page be corrected? Happy to provide more information if this post is not clear.

Best,
Harsha

leios:
This fixes #994 and explains why the error term is actually dt^2 for Stormer Verlet

### DIFF
--- a/contents/verlet_integration/verlet_integration.md
+++ b/contents/verlet_integration/verlet_integration.md
@@ -64,13 +64,24 @@ Here is what it looks like in code:
 [import:3-13, lang:"lisp"](code/clisp/verlet.lisp)
 {% endmethod %}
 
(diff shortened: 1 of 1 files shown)


Amaras (approved):
LGTM

PR #999, opened 2022-04-18
Remove Whitespace imports for Euclidean Algorithm

### LINKED ISSUE #486: Whitepace code not present in Euclidean chapter
I know this sounds like a joke because of how we implemented whitespace in the Euclidean algorithm chapter; however, if you highlight the code snippet, there is no actual whitespace there. It is available in the `code/whitespace` directory, though.

Amaras:
This PR aims to close #486 and declutter the open issues to be a bit more manageable.

Because #486 was a problem with codeblocks, I simply removed them to eliminate the issue entirely.

If someone knows how to solve the issue, please do open a PR and undo my modifications, the code is still available in its folder.

### DIFF
--- a/contents/euclidean_algorithm/euclidean_algorithm.md
+++ b/contents/euclidean_algorithm/euclidean_algorithm.md
@@ -51,8 +51,6 @@ The algorithm is a simple way to find the *greatest common divisor* (GCD) of two
 [import:4-18, lang="php"](code/php/euclidean.php)
 {% sample lang="factor" %}
 [import:1-13, lang="factor"](code/factor/euclid.factor)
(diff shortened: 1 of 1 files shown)


leios:
I mean, we have the whitespace commands in the code at the bottom. I don't think it's that bad of an idea to just leave it as is (or interweave the whitespace code with the comments?)

Amaras:
The problem is using different files for the inline and end code blocks.
Maybe we should just use the commented file instead, because it's currently completely useless.

leios:
Right, so I think we should fix the whitespace commented file so it's actually compilable and use that

PR #1007, opened 2022-11-19
Convolutions in 1d in rust

eengamer2007:
implemented the convolutions 1d in rust

### DIFF
--- a/contents/convolutions/1d/code/rust/1d_convolution.rs
+++ b/contents/convolutions/1d/code/rust/1d_convolution.rs
@@ -1,0 +1,102 @@
+use std::fs::File;
+use std::io::Write;
+use std::cmp::max;
+
+fn main() {
+    let x = normalize(create_sawtooth(200));
+    let y = normalize(create_sawtooth(200));
+
+    let len_x = x.len();
+    let len_y = y.len();
+
+    let full_linear_output = convolve_linear(&x, &y, len_x + len_y - 1);
+    let simple_linear_output = convolve_linear(&x, &y, len_x);
+    let cyclic_output = convolve_cyclic(&x, &y);
+
+    // Save the convolutions to plot them.
+    // The way I do it is a little weird but it is to store the data the same way as the other programs
+    let mut full_file = File::create("full_linear.dat").unwrap();
+    for i in full_linear_output {
+        write!(full_file, "{}\n", i).unwrap();
+    }
+    full_file.sync_all().unwrap();
+
+    let mut simple_file = File::create("simple_linear.dat").unwrap();
+    for i in simple_linear_output {
+        write!(simple_file, "{}\n", i).unwrap();
+    }
+    simple_file.sync_all().unwrap();
+
+    let mut cyclic_file = File::create("cyclic.dat").unwrap();
+    for i in cyclic_output {
+        write!(cyclic_file, "{}\n", i).unwrap();
+    }
+    cyclic_file.sync_all().unwrap();
+}
+
+// Generates a sawtooth function with a given length.
+fn create_sawtooth(length: usize) -> Vec<f64> {
+    let mut array: Vec<f64> = Vec::with_capacity(length);
+    for i in 0..length {
+        array.push((i+1) as f64 / 200.);
+    }
+    array
+}
+
+// Normalizes the given array.
+fn normalize(array: Vec<f64>) -> Vec<f64> {
+    let norm = norm(&array);
+    let mut output: Vec<f64> = Vec::with_capacity(array.len());
+
+    for i in 0..array.len() {
+        output.push(array[i] / norm);
+    }
+
+    output
+}
+
+// Calculates the norm of an array.
+fn norm(array: &Vec<f64>) -> f64 {
+    let mut sum = 0.;
+    for i in array.iter() {
+        sum += i*i;
+    }
+    sum.sqrt()
+}
+
+
+// Modulus function that handles negative values correctly.
+// Assumes that y >= 0.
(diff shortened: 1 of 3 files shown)


eengamer2007:
[lang: rust]

contents/convolutions/1d/code/rust/1d_convolution.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Normalizes the given array.
+fn normalize(array: Vec<f64>) -> Vec<f64> {
+    let norm = norm(&array);
+    let mut output: Vec<f64> = Vec::with_capacity(array.len());
+
+    for i in 0..array.len() {
+        output.push(array[i] / norm);
+    }
```
Amaras, line 53: If you're only using `i` for indexing, consider using an iterator-based for loop.
```suggestion
    for value in array {
        output.push(value / norm);
    }
```

contents/convolutions/1d/code/rust/1d_convolution.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    output
+}
+
+// Calculates the norm of an array.
+fn norm(array: &Vec<f64>) -> f64 {
+    let mut sum = 0.;
+    for i in array.iter() {
```
Amaras, line 61: I don't think the `.iter()` call is necessary
```suggestion
    for i in array {
```

contents/convolutions/1d/code/rust/1d_convolution.rs
```diff
@@ -0,0 +1,102 @@
...
+        write!(cyclic_file, "{}\n", i).unwrap();
+    }
+    cyclic_file.sync_all().unwrap();
+}
+
+// Generates a sawtooth function with a given length.
+fn create_sawtooth(length: usize) -> Vec<f64> {
+    let mut array: Vec<f64> = Vec::with_capacity(length);
+    for i in 0..length {
+        array.push((i+1) as f64 / 200.);
```
Amaras, line 41: The 200 here is a magic number, why is it here?

contents/convolutions/1d/code/rust/1d_convolution.rs
```diff
@@ -0,0 +1,102 @@
...
+fn modulus(x: isize, y: isize) -> isize {((x%y) + y) % y}
+
+fn convolve_linear(signal: &Vec<f64>, filter: &Vec<f64>, output_size: usize) -> Vec<f64> {
+    let mut output = Vec::with_capacity(output_size);
+
+    for i in 0..(output_size as isize) {
+        let mut sum: f64 = 0.;
+        for j in max(0, i as isize - filter.len() as isize)..=i as isize {
+            if j < signal.len() as isize && (i - j) < filter.len() as isize {
+                sum += signal[j as usize] * filter[(i-j) as usize]
```
Amaras, line 79: You return nothing, so you should not forget the `;`
```suggestion
                sum += signal[j as usize] * filter[(i-j) as usize];
```

Amaras (commented):
Hi, thank you for your interest in the AAA.
I've run clippy (with `-W clippy::pedantic` as well) on your code and put all its suggestions here, as well as some suggestions and questions of mine.

I also don't like the fact that you cast *a lot* between `usize` and `isize` in `convolve_linear` and `convolve_cyclic`, but I'm not sure how you could do otherwise without increasing complexity (and it makes pedantic clippy complain a lot)

contents/convolutions/1d/code/rust/1d_convolution.rs
eengamer2007: This code is mostly a rewrite of the c# code with some changes to work with rust. So i'm not entirely sure why it's there, I think its there for normalization of the sawtooth.

Amaras: Yeah, that's also what I gathered. I think you should probably not use 200 as a magic number, but you can use `length` and add a comment like `// divide by length for normalization`

contents/convolutions/1d/code/rust/1d_convolution.rs
```diff
@@ -0,0 +1,99 @@
...
+    for value in array {
+        output.push(value / norm);
+    }
+
+    output
+}
+
+// Calculates the norm of an array.
+fn norm(array: &[f64]) -> f64 {
+    let sum = array.iter().map(|i| i * i).sum::<f64>()
```
Amaras, line 60: Well, now this line makes the code not compile.

eengamer2007: fixed it

contents/convolutions/1d/code/rust/1d_convolution.rs
eengamer2007: fixed with a different suggestion

contents/convolutions/1d/code/rust/1d_convolution.rs
eengamer2007: i did what you suggested

eengamer2007:
i saw in the code of conduct that I should add my name to the contributers list since this is my first contribution. Should that be in this pull request?

leios:
Yeah, add that to this pr, if you can

Amaras:
Oof, sorry for the delay in the response.

So, the main problem with your PR currently is that the line numbers don't track correctly.
I don't see a reason to refuse your code in the AAA now, so please make sure the lines you include your code at are indeed the correct ones before I merge them.

eengamer2007:
sorry for the late fix, school was keeping me busy.

PR #1010, opened 2023-03-29
Adding Racket to the build system

Amaras:
This is a PR to add Racket to the build system.
This has been tested on my Linux (Ubuntu 22.04) machine, and works correctly on the two Racket examples we have in the AAA.


ntindle:
Is it in the dockerfile?

PR #1011, opened 2023-08-18
Flood fill in C#

DimYfantidis:
C# implementation of Flood fill. Added the code in flood_fill.md

### DIFF
--- a/contents/flood_fill/code/csharp/FloodFill.cs
+++ b/contents/flood_fill/code/csharp/FloodFill.cs
@@ -1,0 +1,187 @@
+﻿using System.Linq;
+using System.Diagnostics;
+using System.Collections.Generic;
+
+
+namespace Graphics
+{
+    abstract class FloodFill
+    {
+        // A simple point on the 2 dimensional integer plane.
+        private class Point2I
+        {
+            // Coordinates
+            public int x;
+            public int y;
+            
+            public Point2I(int x, int y)
+            {
+                this.x = x;
+                this.y = y;
+            }
+
+            public override bool Equals(object? o) 
+            {
+                if(!(o is Point2I other))
+                    return false;
+                return this.x.Equals(other.x) && this.y.Equals(other.y);
+            }
+            
+            public override int GetHashCode() 
+            {
+                return HashCode.Combine(x, y);
+            }
+        }
+        
+        // Utility object for comparing List<List<float>> objects.
+        private class FloatListEqualityComparer : IEqualityComparer<List<float>>
+        {
+            public bool Equals(List<float>? one, List<float>? two) 
+            {
+                if (ReferenceEquals(one, two))
+                    return true;
+                if (one == null || two == null)
+                    return false;
+                return one.SequenceEqual(two);
+            }
+
+            public int GetHashCode(List<float> list) {
+                return list.GetHashCode();
+            }
+        }
+        
+        private static bool InBounds(Point2I size, Point2I loc) 
+        {
+          if (loc.x < 0 || loc.y < 0) {
+            return false;
+          }
+          return loc.x < size.x && loc.y < size.y;
+        }
+
+        private static List<Point2I> FindNeighbors(ref List<List<float>> grid, Point2I loc, float oldValue)
+        {
+            var possibleNeighbors = new List<Point2I>
+            {
+                new(loc.x, loc.y + 1),
+                new(loc.x + 1, loc.y),
+                new(loc.x, loc.y - 1),
+                new(loc.x - 1, loc.y)
+            };
+            var neighbors = new List<Point2I>();
+
+            foreach (Point2I possibleNeighbor in possibleNeighbors)
+            {
+                var size = new Point2I(grid[0].Count, grid.Count);
+                var x = possibleNeighbor.x;
+                var y = possibleNeighbor.y;
+                if (!InBounds(size, possibleNeighbor)) {
+                    continue;
+                }
+                if (grid[x][y].Equals(oldValue)) {
+                    neighbors.Add(possibleNeighbor);
+                }
+            }
+            return neighbors;
+        }
+        
+        private static void RecursiveFill(ref List<List<float>> grid, Point2I loc, float oldValue, float newValue) 
+        {
+            if (oldValue.Equals(newValue)) {
+                return;
+            }
+            grid[loc.x][loc.y] = newValue;
+
+            var possibleNeighbors = FindNeighbors(ref grid, loc, oldValue);
+            foreach (Point2I possibleNeighbor in possibleNeighbors) {
+                RecursiveFill(ref grid, possibleNeighbor, oldValue, newValue);
+            }
+        }
+        
+        private static void  QueueFill(ref List<List<float>> grid, Point2I loc, float oldValue, float newValue) 
+        {
+            if (oldValue.Equals(newValue)) {
+                return;
+            }
+
+            var queue = new Queue<Point2I>();
+            queue.Enqueue(loc);
+            grid[loc.x][loc.y] = newValue;
+
+            while (queue.Count > 0) 
+            {
+                var currentLoc = queue.Dequeue();
+                var possibleNeighbors = FindNeighbors(ref grid, currentLoc, oldValue);
+                foreach (Point2I neighbor in possibleNeighbors) 
+                {
+                    grid[neighbor.x][neighbor.y] = newValue;
+                    queue.Enqueue(neighbor);
+                }
+            }
+        }
+        
+        
+        private static void StackFill(ref List<List<float>> grid, Point2I loc, float oldValue, float newValue) 
+        {
+            if (oldValue.Equals(newValue)) {
+                return;
+            }
+
+            var stack = new Stack<Point2I>();
+            stack.Push(loc);
+
+            while (stack.Count > 0) 
+            {
+                var currentLoc = stack.Pop();
+
+                var x = currentLoc.x;
+                var y = currentLoc.y;
+
+                if (grid[x][y].Equals(oldValue)) 
+                {
+                    grid[x][y] = newValue;
+                    var possibleNeighbors = FindNeighbors(ref grid, currentLoc, oldValue);
+                    foreach (Point2I neighbor in possibleNeighbors) {
+                        stack.Push(neighbor);
+                    }
+                }
+            }
+        }
+
+
+        private static void Main(string[] args)
+        {
+            // All neighbouring zeros, adjacent to (1, 1), must be replaced with 3 in the end result.
+            var grid = new List<List<float>>
+            {
+                new(){0, 0, 1, 0, 0},
+                new(){0, 0, 1, 0, 0},
+                new(){0, 0, 1, 0, 0},
+                new(){8, 0, 1, 1, 1},
+                new(){0, 0, 0, 0, 0}
+            };
+            var solutionGrid = new List<List<float>>
+            {
+                new(){3, 3, 1, 0, 0},
+                new(){3, 3, 1, 0, 0},
+                new(){3, 3, 1, 0, 0},
+                new(){8, 3, 1, 1, 1},
+                new(){3, 3, 3, 3, 3}
+            };
+            var startingPoint = new Point2I(1, 1);
+            var gridComparator = new FloatListEqualityComparer();
+
+            var testGrid = new List<List<float>>(grid);
+            RecursiveFill(ref testGrid, startingPoint, 0, 3);
+            Debug.Assert(testGrid.SequenceEqual(solutionGrid, gridComparator), "Recursive Flood Fill Failed");
+            
+            testGrid = new List<List<float>>(grid);
+            QueueFill(ref testGrid, startingPoint, 0, 3);
+            Debug.Assert(testGrid.SequenceEqual(solutionGrid, gridComparator), "Queue Flood Fill Failed");
+
+            testGrid = new List<List<float>>(grid);
+            StackFill(ref testGrid, startingPoint, 0, 3);
+            Debug.Assert(testGrid.SequenceEqual(solutionGrid, gridComparator), "Stack Flood Fill Failed");
+
+        }
+    }
+}
--- a/contents/flood_fill/flood_fill.md
+++ b/contents/flood_fill/flood_fill.md
@@ -96,6 +96,8 @@ In code, this might look like this:
 [import:10-25, lang="python"](code/python/flood_fill.py)
 {% sample lang="coco" %}
 [import:15-20, lang="coconut"](code/coconut/flood_fill.coco)
+{% sample lang="cs" %}
+[import:61-85, lang="csharp"](code/csharp/FloodFill.cs)
 {% endmethod %}
 
 
@@ -118,6 +120,8 @@ In code, it might look like this:
 [import:55-63, lang="python"](code/python/flood_fill.py)
 {% sample lang="coco" %}
 [import:52-61, lang:"coconut"](code/coconut/flood_fill.coco)
+{% sample lang="cs" %}
+[import:87-98, lang="csharp"](code/csharp/FloodFill.cs)
 {% endmethod %}
 
 The above code continues recursing through available neighbors as long as neighbors exist, and this should work so long as we are adding the correct set of neighbors.
@@ -135,6 +139,8 @@ Additionally, it is possible to do the same type of traversal by managing a stac
 [import:27-36, lang="python"](code/python/flood_fill.py)
 {% sample lang="coco" %}
 [import:23-34, lang:"coconut"](code/coconut/flood_fill.coco)
+{% sample lang="cs" %}
+[import:123-148, lang="csharp"](code/csharp/FloodFill.cs)
 {% endmethod %}
 
 This is ultimately the same method of traversal as before; however, because we are managing our own data structure, there are a few distinct differences:
@@ -180,6 +186,8 @@ The code would look something like this:
 [import:38-53, lang="python"](code/python/flood_fill.py)
 {% sample lang="coco" %}
 [import:36-49, lang:"coconut"](code/coconut/flood_fill.coco)
+{% sample lang="cs" %}
+[import:100-120, lang="csharp"](code/csharp/FloodFill.cs)
 {% endmethod %}
 
 Now, there is a small trick in this code that must be considered to make sure it runs optimally.
@@ -264,6 +272,8 @@ After, we will fill in the left-hand side of the array to be all ones by choosin
 [import:, lang="python"](code/python/flood_fill.py)
 {% sample lang="coco" %}
 [import, lang="coconut"](code/coconut/flood_fill.coco)
+{% sample lang="cs" %}
+[import, lang="csharp"](code/csharp/FloodFill.cs)
 {% endmethod %}
 
 


Amaras:
Hello @DimYfantidis and thank you for your interest in the AAA.

We'll try to review your PR as soon as possible, but be aware that it might take a while, as I'm not sure we currently have available C# reviewers.

From what I see, there is a lot of boilerplate code, which seems like a lot to me, but I'm not a C# expert.

I'll nonetheless get a first review path going soon-ish, so that it doesn't get too stuck in review limbo

Amaras:
I tried compiling on my machine, and it simply didn't.
I'm using an Ubuntu 20.04 machine, with the `mono-complete` package.
Could you provide compilation instructions for Linux?
Here's the output:
```
❯ mcs FloodFill.cs 
FloodFill.cs(65,33): error CS1525: Unexpected symbol `+', expecting `)', `,', or `identifier'
FloodFill.cs(65,36): error CS1525: Unexpected symbol `)', expecting `)', `,', or `identifier'
FloodFill.cs(66,16): error CS1525: Unexpected symbol `new', expecting `)', `,', or `identifier'
FloodFill.cs(66,34): error CS1525: Unexpected symbol `.', expecting `)', `,', or `identifier'
FloodFill.cs(67,16): error CS1525: Unexpected symbol `new', expecting `)', `,', or `identifier'
FloodFill.cs(67,30): error CS1525: Unexpected symbol `.', expecting `)', `,', or `identifier'
FloodFill.cs(68,16): error CS1525: Unexpected symbol `new', expecting `)', `,', or `identifier'
FloodFill.cs(68,34): error CS1525: Unexpected symbol `.', expecting `)', `,', or `identifier'
FloodFill.cs(156,20): error CS1525: Unexpected symbol `)', expecting `(' or `type'
FloodFill.cs(156,25): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(156,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(156,31): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(156,34): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(157,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(157,25): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(157,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(157,31): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(157,34): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(158,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(158,25): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(158,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(158,31): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(158,34): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(159,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(159,25): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(159,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(159,31): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(159,34): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(160,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(160,25): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(160,28): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(160,31): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(160,34): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(164,20): error CS1525: Unexpected symbol `)', expecting `(' or `type'
FloodFill.cs(164,25): error CS1525: Unexpected symbol `3', expecting `(' or `type'
FloodFill.cs(164,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(164,31): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(164,34): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(165,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(165,25): error CS1525: Unexpected symbol `3', expecting `(' or `type'
FloodFill.cs(165,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(165,31): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(165,34): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(166,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(166,25): error CS1525: Unexpected symbol `3', expecting `(' or `type'
FloodFill.cs(166,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(166,31): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(166,34): error CS1525: Unexpected symbol `0', expecting `(' or `type'
FloodFill.cs(167,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(167,25): error CS1525: Unexpected symbol `3', expecting `(' or `type'
FloodFill.cs(167,28): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(167,31): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(167,34): error CS1525: Unexpected symbol `1', expecting `(' or `type'
FloodFill.cs(168,16): error CS1525: Unexpected symbol `new', expecting `(' or `type'
FloodFill.cs(168,25): error CS1525: Unexpected symbol `3', expecting `(' or `type'
FloodFill.cs(168,28): error CS1525: Unexpected symbol `3', expecting `(' or `type'
FloodFill.cs(168,31): error CS1525: Unexpected symbol `3', expecting `(' or `type'
FloodFill.cs(168,34): error CS1525: Unexpected symbol `3', expecting `(' or `type'
Compilation failed: 58 error(s), 0 warnings
```

DimYfantidis:
Thank you for your time, I'm very sorry but it seems that there are no special compilation instructions for Linux. Normally, "mcs FloodFill.cs" should be enough, but it seems that I've written code that could compile and execute due to specific IDE configurations that I had on my machine. I have made changes to the code and it now compiles successfully on Ubuntu 22.04.3 simply by typing the aforementioned command (It would be very unlucky if it was to fail Ubuntu 20.04 though very unlikely; I believe it should compile fine on your machine now).

I'm not sure on how I could change the code I have commited. Is it possible that, in this specific pull request, I could change the contents of FloodFill.cs? Or should I create a new pull request? If the former is possible, notify me so that we could come in contact on Discord. (I'm a new at contributing to open source projects and github projects)

Thanks again and I'm very sorry.

Amaras:
You can simply modify the code and `git add FloodFill.cs` and then `git commit` with a relevant message.
When you then `git push`, we'll get the up to date code in this PR. :)

DimYfantidis:
Done. I'm on standby in case another problem occurs, although now I suppose everything is fine. 👍

Amaras:
Thanks, it compiles now.
I'll try to get a PR going to get C# into our build system to make sure we don't have a problem like that any more.
Hopefully we can get a competent C# developer who can look at your PR soon :)

Amaras:
Alright, the PR to get C# in our build system is up (#1012)

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,190 @@
...
+        private class FloatListEqualityComparer : IEqualityComparer<List<float>>
+        {
+            public bool Equals(List<float> one, List<float> two)
+            {
+                if (ReferenceEquals(one, two))
+                    return true;
+                if (one == null || two == null)
+                    return false;
+                return one.SequenceEqual(two);
+            }
```
Amaras, line 49: Couldn't you shorten this function like so?
```suggestion
            public bool Equals(List<float> one, List<float> two)
            {
                return ReferenceEquals(one, two) && one != null && two != null && one.SequenceEqual(two);
            }
```

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,190 @@
...
+                return list.GetHashCode();
+            }
+        }
+
+        private static bool InBounds(Point2I size, Point2I loc)
+        {
+          if (loc.x < 0 || loc.y < 0) {
+            return false;
+          }
+          return loc.x < size.x && loc.y < size.y;
```
Amaras, line 61: shouldn't this be?
```suggestion
          return loc.x >= 0 && loc.y >= 0 && loc.x < size.x && loc.y < size.y;
```

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,190 @@
...
+                new Point2I(loc.x, loc.y + 1),
+                new Point2I(loc.x + 1, loc.y),
+                new Point2I(loc.x, loc.y - 1),
+                new Point2I(loc.x - 1, loc.y)
+            };
+            var neighbors = new List<Point2I>();
+
+            foreach (Point2I possibleNeighbor in possibleNeighbors)
+            {
+                var size = new Point2I(grid[0].Count, grid.Count);
```
Amaras, line 77: presumably, you don't need to recompute `size` every loop

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,190 @@
...
+
+        private static void Main(string[] args)
+        {
+            // All neighbouring zeros, adjacent to (1, 1), must be replaced with 3 in the end result.
+            var grid = new List<List<float>>
+            {
+                new List<float>(){0, 0, 1, 0, 0},
+                new List<float>(){0, 0, 1, 0, 0},
+                new List<float>(){0, 0, 1, 0, 0},
+                new List<float>(){8, 0, 1, 1, 1},
```
Amaras, line 162: Why the 8 inside all the 0's?

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,190 @@
...
+            }
+
+            public override int GetHashCode()
+            {
+                int hash = 17;
+                hash = 31 * hash + x;
+                hash = 31 * hash + y;
+                return hash;
+            }
+        }
```
Amaras, line 37: Is all this necessary, and if so, where?

Amaras (changes requested):
Alright, I'm not a C# expert, so this can only ever be a first pass.

Feel free to disagree with me on everything I said, since I'm not an expert.
The only change I demand is the `size` computation on each loop that can't be assumed to be optimized away, the rest is up to you (with explanations for disagreements, though, please)

contents/flood_fill/code/csharp/FloodFill.cs
DimYfantidis: Let's say `one` and `two` are not `null`.  So the expression `one != null && two != null` evaluates to `true`. This way, 
`ReferenceEquals(one, two) && one != null && two != null && one.SequenceEqual(two)` can be simplified to 
`ReferenceEquals(one, two) && one.SequenceEqual(two)`. If `one` and `two` point to different `List`s, then the first argument will evaluate to false. That means that, no matter what `one.SequenceEqual(two)` (if they have equal contents or not) evaluates to, the result will always be false, which ultimately defeats the purpose of overriding of the `bool Equals(List<float>, List<float>)` function.

DimYfantidis: So, I'm afraid the initial function and the suggested changes are not equivalent.

contents/flood_fill/code/csharp/FloodFill.cs
DimYfantidis: Different integers represent different attributes (e.g.: different colors in a canvas). Other implementations use 0 and 1 to create an example grid, but i supposed i could throw a third value inside. This way it becomes more clear that it is not a binary problem, but there can be more than two values.

contents/flood_fill/code/csharp/FloodFill.cs
DimYfantidis: You are right, it seems that there is no equality comparison nor hashing of `Point2I` instances in the code. I will remove these functions.

contents/flood_fill/code/csharp/FloodFill.cs
Amaras: It was too late, I used the wrong operator.
I meant to use `||` instead of the first `&&`

contents/flood_fill/code/csharp/FloodFill.cs
Amaras: Indeed, the C++ implementation needs an edit.
The thing is, you don't use the size of each line, you use the size of the first line and assume it's the same for each line.

contents/flood_fill/code/csharp/FloodFill.cs
Amaras: Okay, fair enough

contents/flood_fill/code/csharp/FloodFill.cs
DimYfantidis: Ah, yes that's better. It didn't cross my mind to correct your typo, I'm sorry. 
`&&` has higher priority than `||` so I believe you are right.

contents/flood_fill/code/csharp/FloodFill.cs
DimYfantidis: Hmm, indeed. I blindly copied and pastied that part from the C++ implementation which is also wrong on this case.
I will probably remove it from the loop, but I could probably come up with a way to find in which line of the matrix the given point is situated, in order to get its length. Thanks for your observations.

DimYfantidis:
Alright, I have refactored code and markdown according to the above discussion.

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,168 @@
...
+                return ReferenceEquals(one, two) || one != null && two != null && one.SequenceEqual(two);
+            }
+
+            public int GetHashCode(List<float> list)
+            {
+                return list.GetHashCode();
+            }
+        }
+
+        private static bool InBounds(Point2I size, Point2I loc)
```
andrewhooker2, line 38: One thing to consider here would be the readability to know what the variables of "siz" and "loc" are trying to convey. It looks as though size is supposed to convey bounds and loc is trying to convey a point. 

**Variable name change suggestion:** 
- siz -> bounds
- Ioc -> point ( although loc is still a pretty viable name here )

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            return loc.x >= 0 && loc.y >= 0 && loc.x < size.x && loc.y < size.y;
```
andrewhooker2, line 40: With your functions I notice that you don't handle for errors:
A simple check you could consider throwing in would be something like:

`    if (bounds == null || point == null)
        throw new ArgumentNullException(bounds == null ? nameof(bounds) : nameof(point));
`
Feel free to update this to fit your needs.

andrewhooker2: I won't add in this comment per function but, it is something to consider.

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,168 @@
...
+            stack.Push(loc);
+
+            while (stack.Count > 0)
+            {
+                var currentLoc = stack.Pop();
+
+                var x = currentLoc.x;
+                var y = currentLoc.y;
+
+                if (grid[x][y].Equals(oldValue))
```
andrewhooker2, line 120: I'd suggest adding a check to see if you are out of bounds here with your x and y.
This would help prevent IndexOutOfRangeException Errors in the future.

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,168 @@
...
+                {
+                    grid[neighbor.x][neighbor.y] = newValue;
+                    queue.Enqueue(neighbor);
+                }
+            }
+        }
+
+        private static void StackFill(ref List<List<float>> grid, Point2I loc, float oldValue, float newValue)
+        {
+            if (oldValue.Equals(newValue)) {
```
andrewhooker2, line 106: Would a simple == check not work here?

Just curious if we could do something like
if (oldValue == newValue) return

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,168 @@
...
+                return list.GetHashCode();
+            }
+        }
+
+        private static bool InBounds(Point2I size, Point2I loc)
+        {
+            return loc.x >= 0 && loc.y >= 0 && loc.x < size.x && loc.y < size.y;
+        }
+
+        private static List<Point2I> FindNeighbors(ref List<List<float>> grid, Point2I loc, float oldValue)
```
andrewhooker2, line 43: In this instance I don't think it is necessary for you to use the ref keyword with:
`ref List<List<float>> grid`

Since you are only reading from the grid and not modifying it, there is no need to use the **ref** keyword.

andrewhooker2 (commented):
Feel free to comment back I would love to hear from you!

contents/flood_fill/code/csharp/FloodFill.cs
ntindle: I'd prefer the orig if possible

andrewhooker2: It would be preferred to use the original for readability.
If you are wanting to shorten and since there is only two options for return you could use a ternary operator here. 

**Readability > Complexity** for C# Code is what I will always suggest.

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Diagnostics;
+using System.Collections.Generic;
+
+
+namespace Graphics
+{
+    abstract class FloodFill
+    {
+        // A simple point on the 2 dimensional integer plane.
+        private class Point2I
```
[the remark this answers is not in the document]
DimYfantidis: Hello, thanks for the feedback.

I had initially named it "2D" but then i remembered how OpenGL uses "`glVertex2f()`" to refer to a vertex (vector) of 2 **floats**, "`glVertex2i`" to refer to a vertex of 2 **integers** and "`glVertex2d`" to refer to a vertex of 2 **doubles**. Since we're using points with integers, I used `Point2I` since according to OpenGL (other libraries do it as well) it would mean "Point of two Integers". I understand that at first glance it's confusing, but now that I have given some context I hope that the logic behind it is clear :)

It's simply a convention that I stick to personally.

contents/flood_fill/code/csharp/FloodFill.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            // Coordinates
+            public int x;
```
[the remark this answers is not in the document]
DimYfantidis: I understand that according to Object Oriented Programming principles all variables should have setter and getter functions. For that reason, sometimes I don't really like OOP as it takes simple things and makes them complicated. 

The program that I wrote is supposed to be as simple as possible and easy to understand for a beginner programmer and algorithm enthusiast that would like to make a first contact with the flood fill algorithm in C#. I feel that 
```C#
public int getX() { return x; }
public int getY() { return y; }
public void setX(int x) { this.x = x; }
public void setY(int y) { this.y = y; }
```
would add extra complexity to the program. Although, no problem from my part, I could implement encapsulation for `x` and `y`.

contents/flood_fill/code/csharp/FloodFill.cs
DimYfantidis: Yes it would, but values are floats and my IDE goes crazy because: 
"`'==' operator is prone to floating point arithmetic errors. Try using 'abs(oldValue, newValue) < ERROR_CONSTANT' instead`"
So, one quick way out is to use the built-in `float.Equals(float)` method because if C# is smart enough to correct me on it, it should be smart enough to implement it itself.

contents/flood_fill/code/csharp/FloodFill.cs
DimYfantidis: I didn't do it for read/write reasons, I did it for passing quickly the data structure into the function in $\mathcal{O}(1)$ time.

From what I've understood from C#, since the `ref` keyword exists then it's like C++ where everything is passed by value except if you say "pass by reference" explicitly (C++ uses the `&` symbol). This way, since `List<List<float>>` is an object of $\mathcal{O}(n^{2})$ space complexity, then copying it should take $\mathcal{O}(n^{2})$ time to pass it inside the function. Passing only its reference should take $\mathcal{O}(1)$ time. In C++ this would be written as
```C++
const  List<List<float>>& grid
```
in order to avoid copying the entire data structure. 

Keep in mind that I am using my C++ experience and making some assumptions that the inner mechanisms of C# work in the same way. If what I'm saying is wrong and there is no correspondence between the two languages feel free to correct me.

PR #1012, opened 2023-08-20
Adding C# to the build system (using the Mono toolchain)

Amaras:
This is a PR to add C# to our build system, making sure we have a common ground.
I see we have the dotnet sdk, but I could not find the command to compile using it, don't hesitate to close this PR if you give me the proper way.

Because I used Mono, there were a few changes in two files, as Mono doesn't allow functions inside functions, apparently.

This follows from #1011, but might need to be merged before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
 
 # [Optional] Uncomment this section to install additional OS packages.
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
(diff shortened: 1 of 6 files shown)


Amaras:
Right... @ntindle or @ShadowMitia, do I need to separate the Docker stuff from the SCons stuff for the build to pass?

ntindle:
Docker should probably go in first. I can't remember the exact repo ci we used but if we did it right, it won't use the dockerfile you provide here but the one on the base branch

ShadowMitia:
FYI: for dotnet you can use this as reference: https://github.com/SebLague/Chess-Challenge/discussions/3
But it could be good to support both? I don't really know C# ecosystem well enough.
I don't have an opinion on which one is better to use.

Amaras:
There we go, everything builds :)
If someone wants to review it, please do so we can move on

ShadowMitia (approved):
LGTM

PR #1013, opened 2023-08-22
Add C# compilation with Mono

Amaras:
Because pushing both the Dockerfile changes and the SCons changes didn't work, here is the C# Docker change


Amaras:
@ntindle here's your time

ntindle (approved):

PR #1015, opened 2023-08-23
Actions/checkout@v2 to actions/checkout@v3

ShadowMitia:
Warning is generated inside github action logs, because it uses an old version of nodejs. 
This should remove the warning, and won't break the action once github finishes migration. 

See https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/ for more details.


Amaras (approved):
LGTM

PR #1016, opened 2023-09-06
PRNG + LCG chapters

leios:
This is a draft PR for the next set of chapters. It will include the following:

* [ ] An intro to RNG
    * [ ] a discussion of where RNG is used in different fields
    * [ ] quick histogram check to see how random your RNG is
* [ ] Diehard tests
    * [ ] Birthday spacings
    * [ ] Overlapping permutations
    * [ ] Ranks of matrices
    * [ ] Monkey tests
    * [ ] Count the 1s
    * [ ] Parking lot test
    * [ ] Minimum distance test
    * [ ] Random spheres test
    * [ ] The squeeze test
    * [ ] Overlapping sums test
    * [ ] Runs test
    * [ ] The craps test
* [ ] Linear Congruential Generators
    * [ ] Explain why they might be used in CG applications / GPU kernels (due to other RNG methods being cumbersome)

Might be biting off a bit too much, but I've been wanting to do diehard tests forever now and it'll be fun creating the testsuite.


Amaras:
Idea: add Linear-Feedback Shift Registers as a (cheap) hardware PRNG, although it can be quite weak as a PRNG

PR #1020, opened 2023-10-30
Declare "noninteractive" environment variable globally in Dockerfile

ShadowMitia:


berquist:
Is #1021 an accidental incomplete duplicate?

berquist:
My only concern is that this will be annoying for someone using this image as a dev container.  Is there a way to set image build args for dev containers?

ShadowMitia:
ARGs are only available during builds, so it shouldn't interfere afterwards. And you can override them.

berquist (approved):
Right, that slipped my mind.  All seems good.

PR #1021, opened 2023-10-30
Add --no-cache-dir to pip install instructions in Dockefile 

ShadowMitia:


berquist (changes requested):
Aside from the rebase, can you also switch these from `pip install` to `python -m pip install`?

ShadowMitia:
All done

